PR #3587, opened 2022-03-07
chore: add bootstrap 3 copy on npm postinstall

NGPixel:
This copy the bs3 dist into the correct location on npm install.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -41,6 +41,14 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[package.json]
+indent_style = space
+indent_size = 2
+end_of_line = lf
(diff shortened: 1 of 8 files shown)


.editorconfig
```diff
@@ -41,6 +41,14 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[package.json]
```
rjsparks, line 44: Did you intend for this to come in here?

rjsparks (approved):

.editorconfig
NGPixel: Yes, unfortunately the syntax doesn't allow for completely different paths for the same block, so it has to be duplicated.

PR #3601, opened 2022-03-09
chore: fix docker dev boot for parcel build

NGPixel:


docker/scripts/app-init.sh
```diff
@@ -4,6 +4,12 @@ WORKSPACEDIR="/root/src"
 
 service rsyslog start
 
+# Generate static assets
+
+npm install
+echo "Building static assets... (this could take a minute or two)"
+npx parcel build
```
larseggert, line 11: Should maybe explain that `npx parcel build` must be re-run when sources are changed, since this is a new step with bs5?

PR #3620, opened 2022-03-11
fix: Remove attempt to do smart linewraps for doc titles

### LINKED ISSUE #3609: Weird line breaking in (at least) WG doc lists
See wg/dnsop/documents/. The document title for draft-ietf-dnsop-rfc5933-bis has an extra break that should not be there. The same is true for RFC 3258. This does not appear on the current Datatracker page.

larseggert:
Fixes #3609.


larseggert:
I have no idea what the failing check is complaining about, even after reading "details".

PR #3622, opened 2022-03-11
fix: Switch to showing a single "Edit materials" button

### LINKED ISSUE #3594: multiple copies of "edit materials" button in bootstrap5 branch
The bootstrap 5 version of the `meeting/materials` page has five copies of the "Edit materials" button in the Proceedings Materials section, one for each material type. There should be only one button - it takes you to the page to edit all of the proceedings materials.

larseggert:
Fixes #3594.


jennifer-richards (approved):
Looks good - I'm a little worried the placement makes it unclear which rows it applies to, but I don't see an obviously better alternative.

PR #3625, opened 2022-03-11
fix: point to GitHub for bug reports

rjsparks:


NGPixel (approved):

ietf/templates/500.html
```diff
@@ -17,9 +17,9 @@ <h2>Internal Server Error.</h2>
   </p>
   <p class="error-text">
     A failure report with details about what happened has been sent to the
-    server administrators.  It would be helpful if you would file a bug
-    report with additional information at the
-    <a href="https://trac.ietf.org/trac/ietfdb/newticket">IETF database issue tracker</a>, too.
+    server administrators.  It would be helpful if you would create an issue
+    providing additional information at 
+    <a href="https://github.com/ietf-tools/datatracker/issues">GitHub</a>, too.
```
jennifer-richards, line 22: Is it a conscious decision to point to the issue list here instead of to the new ticket page? (I.e., this link is different from the others in this PR, is that intentional?)

rjsparks: I waffled, but it's OBE - all should point to the page that Nick created, and I'm fine with those changes.

PR #3642, opened 2022-03-14
chore: bring docker-compose to root level

NGPixel:
This moves docker-compose to the root of the project, enabling the resulting containers to be named after the parent directory, instead of "docker".

All scripts `cleanall`, `cleandb` and `run` have been updated. No apparent change for the user, they are still in the same location.


rjsparks (commented):
Squashed and merged before entering trivial "yes" review.

PR #3646, opened 2022-03-15
feat: Save HTML source to file when tidy validation fails during test

larseggert:
Instead of dumping the whole source code to the console.


rjsparks (approved):
Consider writing these to somewhere already git ignored instead.

PR #3649, opened 2022-03-15
chore: remove svn/trac infrastructure

rjsparks:


requirements.txt
```diff
@@ -66,7 +66,6 @@ rfc2html>=2.0.1
 scout-apm>=2.23.0
 selenium>=3.141.0,<4.0
 six>=1.10.0
-svn>=1.0.1
 tblib>=1.3.0
 tqdm>=3.7.0
 #Trac>=1.0.10,<1.2
```
jennifer-richards, line 71: Remove Trac entirely while we're at it?

jennifer-richards (approved):
This looks fine to me. Searching through the codebase for `trac_`, I came across a [`trac_links(text)` method](https://github.com/ietf-tools/datatracker/blob/868afd287c6f16b9fb41eb8d95f1637ca105c6cc/ietf/release/views.py#L29) that should probably be removed or updated. Approving because I think it's fine to merge this and handle that separately if more convenient.

requirements.txt
rjsparks: trac_links will go away when we refactor the /release view.
pulling the #Trac line out of requirements is reasonable.

PR #3651, opened 2022-03-15
fix: Allow AD and Secretariat to update the people to notify about conflict-reviews until the are announced

### LINKED ISSUE #3639: Conflict Review documents should allow "Send Notices To" to be edited by appropriate parties
The ability to edit the "send notices to" field should be similar to the ability to change the conflict review text. 
cc. @akvezza, @cindymorgan

russhousley:
Allow AD and Secretariat to update the people to notify about conflict-reviews until the are announced. Fixes #3639.

### DIFF
--- a/ietf/templates/doc/document_conflict_review.html
+++ b/ietf/templates/doc/document_conflict_review.html
@@ -105,7 +105,7 @@
                 <td></td>
                 <th>Send notices to</th>
                 <td class="edit">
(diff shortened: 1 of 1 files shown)


rjsparks:
Why should this gated on the state of the document in any way?
If the secretariat or an AD (or other stream-appropriate person) wanted to change the list of addresses to "also notify" years after the document was approved in case someone added a comment to the document and they wanted to be notified, why should the tool prevent that?

PR #3656, opened 2022-03-16
fix: Fix a regression in #3625

larseggert:
Tag closed incorrectly.


rjsparks (approved):

rjsparks:
Why the newline change at the end of the file?

larseggert:
My editor uses `.editorconfig`, which I think makes that happen. If we don't want final newlines trimmed, we should adjust the settings there.

PR #3658, opened 2022-03-16
ci: Run the test suite for PRs and pushes

larseggert:


.github/workflows/ci-run-tests.yml
```diff
@@ -0,0 +1,27 @@
+name: Run all tests
+
+on:
+  push:
```
NGPixel, line 4: I would remove the push trigger, as this would re-run the tests a second time upon merge, which is not necessary.

NGPixel (changes requested):

NGPixel (approved):

larseggert:
Yes, at least the apt stuff could be canned.

NGPixel:
@larseggert I made some significant changes to the workflow so that it uses the "datatracker-test-base" image (which is automatically built when either package.json or requirements.txt get modified)

The image is essentially the same as the dev image, but uses python:3.6-bullseye as base instead and it removes some dev specific installs / configs that are not relevant for testing.

The workflow now run directly from inside the test container instead of launching a docker-compose. I disabled the MariaDB container as we use sqlite during tests, but we can re-enable it in the future if needed.

Right now it fails at the run tests step as it wants to apply migrations first. However, adding `manage.py migrate` just before fails as it expects some tables to be present already...

larseggert:
> Right now it fails at the run tests step as it wants to apply migrations first. However, adding `manage.py migrate` just before fails as it expects some tables to be present already...

Do you have a resolution to that?

rjsparks:
The test runner applies migrations automatically - if it's running into trouble saying it can't touch a table, someone may have added some code that tries to use tables before django is actually initialized. Can you give me a link to the failure?

Fwiw, ietf/manage.py migrate by itself will migrate the database configured in settings/settings_local. The tests are using the database configured in settings__sqlitetest. If you're not invoking the tests with --settings=settings_sqlitetest, they _will_ fail. And do not try to run migrate with --settings=settings_sqlitetest - It may or may not succeed, but after you read about what the test runner does, you'll see that's non-sensical.

NGPixel:
Finally got the tests to run correctly. The culprit was a missing docker-cli on the base test image, which is somehow needed to talk to the DB.

PR #3662, opened 2022-03-17
fix: More fixes for narrow screens

larseggert:
Need to keep checking for issues, there are more.


ietf/static/js/ietf.js
```diff
@@ -46,6 +46,49 @@ $(document)
...
+                }
+                // now replace only that text inside the element's HTML
+                var newhtml = $(this)
+                    .html()
+                    .replace(text, newtext);
+                $(this)
+                    .html(newhtml);
+            });
+
+        // $("#content table.tablesorter")
```
rjsparks, line 70: Did you mean to bring this in?

ietf/templates/meeting/requests.html
```diff
@@ -32,21 +32,22 @@ <h2 class="mt-5" id="{{ area.grouper.acronym }}">
...
             </thead>
             {% for session in area.list %}
                 {% ifchanged %}
+                    {% if not forloop.first %}</tbody>{% endif %}
                     <thead>
                         <tr>
                             <th class="table-warning" colspan="7">{{ session.current_status_name|capfirst }}</th>
                         </tr>
                     </thead>
+                    <tbody>
```
rjsparks, line 49: Isn't this going to leave an unclosed `<tbody>` at the last iteration of the loop?

ietf/templates/meeting/requests.html
```diff
@@ -79,16 +80,12 @@ <h2 class="mt-5" id="{{ area.grouper.acronym }}">
...
-                                                    {% if constraint.name.slug == "bethere" %}
-                                                        {% person_link constraint_person %}
+                                                    {% if constraint.name.slug == "bethere" %}{% person_link constraint_person %}
                                                     {% else %}
                                                         {% with constraint.name.slug as constraint_name_slug %}
-                                                            <span class="{% if constraint_name_slug == 'conflict' %}text-danger{% elif constraint_name_slug == 'conflic2' %}text-warning{% elif constraint_name_slug == 'conflic3' %}text-success{% else %}{{ constraint_name_slug }}{% endif %}">
+                                                            <span class="{% if constraint_name_slug == 'chair_conflict' %}text-danger{% elif constraint_name_slug == 'tech_overlap' %}text-info{% elif constraint_name_slug == 'key_participant' %}text-success{% else %}{{ constraint_name_slug }}{% endif %}">
```
rjsparks, line 86: This should become aware of the constraints in use by a given meeting: https://github.com/ietf-tools/datatracker/blob/main/ietf/meeting/models.py#L125.

rjsparks (changes requested):
Please verify the new loop construction in requests.html doesn't leave an unclosed `<tbody>`

larseggert:
I don't think it does - the tidy check that the test suite now runs would have caught that.

ietf/static/js/ietf.js
larseggert: Yes. This really helps with wrapping on narrow screens.

rjsparks:
The tidy check must not do what you think it does.

I checked out the PR and looked at the html generated by ietf.meeting.tests_views.SessionTests.test_meeting_requests.
I put it in a file and got this result:
```
rjsparks@undex lars % grep tbody ~/tmp/foo.html
                    <tbody>
                    </tbody>
                    <tbody>
```
Looking into the call to tidy:
```
    r.status_code: '200'
    r["content-type"]: 'text/html; charset=utf-8'
    skip_verify: 'False'
    errors: 'line 430 column 13 - Warning: trimming empty <i>
line 527 column 21 - Warning: trimming empty <i>
line 520 column 17 - Warning: trimming empty <button>
line 598 column 17 - Warning: trimming empty <i>
line 638 column 17 - Warning: trimming empty <i>
line 687 column 17 - Warning: trimming empty <i>
line 730 column 29 - Warning: trimming empty <i>
line 735 column 33 - Warning: trimming empty <i>
'
```

larseggert:
I looked into this, and HTML Tidy misses a bunch of HTML errors. Gah. I am adding HTML5 validation based on vnu in #3682 and will rework this PR once that is done.

jennifer-richards (approved):
Looks ok to me aside from what @rjsparks  already flagged

jennifer-richards:
Oops, didn't realize my approval would authorize github to merge - should have just been a comment but I don't see a way to change that.

ietf/templates/meeting/requests.html
larseggert: I can do a filter, but the line Robert indicated doesn't give me enough information about what the desired behavior is?

ietf/templates/doc/search/status_columns.html
```diff
@@ -64,7 +64,7 @@
             <br>
             Action Holder{{ doc.documentactionholder_set.all|pluralize }}:
             {% for action_holder in doc.documentactionholder_set.all %}
-                {% person_link action_holder.person title=action_holder.role_for_doc %} {{ action_holder|action_holder_badge }}{% if not forloop.last %},{% endif %}
+                {% person_link action_holder.person title=action_holder.role_for_doc %}{% if action_holder|action_holder_badge %} {{ action_holder|action_holder_badge }}{% endif %}{% if not forloop.last %},{% endif %}
```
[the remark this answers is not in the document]
larseggert: The issue is that if there is no `action_holder`, there should also not be a space before the (empty) badge.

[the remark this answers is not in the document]
rjsparks: Look to see if that space is appropriate for everywhere the action_holder templatetag is used. If it is, then add it to the html that the templatetag emits when a badge should be present.

ietf/templates/meeting/requests.html
rjsparks: this is probably bigger than this first pass and should be split into a separate ticket. A meeting knows which constraint names are in use. Right now, we have hardcoded how to present each constraint into the template (chair_conflict is text-danger, etc). We would need to add a filter, as jennifer is suggesting, that encapsulated what class to provide for each constraint type.

PR #3681, opened 2022-03-19
fix: remove overflow-hidden from div#content

### LINKED ISSUE #3666: Meeting schedule editor floating panel no longer floats
### What happened?

The meeting schedule editor's UI includes a footer that holds the tray for unscheduled sessions, session info panel, and a few other UI elements. When the window size is too small to show the entire schedule, this footer is meant to float pinned to the bottom of the viewport so it is accessible as the user scrolls up and down on the schedule.

On the bs5 branch, this footer is rendered at the very bottom of the schedule display. To access it, the user must scroll to the end of the page.

Need to restore the floating behavior.

Firefox

### Code of Conduct

jennifer-richards:
Allows position="sticky" items to attach to the viewport instead of the #content element.

Fixes #3666

### DIFF
--- a/ietf/templates/base.html
+++ b/ietf/templates/base.html
@@ -84,7 +84,7 @@
                         </ul>
                     </div>
                 {% endif %}
-                <div class="col overflow-hidden mx-lg-3 ietf-auto-nav" id="content">
+                <div class="col mx-lg-3 ietf-auto-nav" id="content">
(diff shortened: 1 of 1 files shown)


jennifer-richards:
The failing test is the one fixed in efc87dc35b16b25210c64d703528c26669ab313d

PR #3682, opened 2022-03-19
ci: Also check generated HTML with the "vnu" validator

larseggert:
Because HTML Tidy apparently misses a bunch of errors. Also fix a bunch of those missed HTML errors.


larseggert:
@rjsparks once this one is merged, I'll use it to validate #3662.

rjsparks (approved):

Issue #3680, opened 2022-03-19
In page /group/secdir/reviewers/ the document name wraps over multiple lines


kivinen:
### What happened?

Doing this will make the list items multiple lines long which makes the page too spare. 

### Code of Conduct

- [X] I agree to follow the IETF's Code of Conduct

PR #3683, opened 2022-03-19
Fix for reviewer table

kivinen:
This fixes the reviewer list so that document name does not span multiple lines. Fixes issue #3680.

### DIFF
--- a/ietf/templates/group/reviewer_overview.html
+++ b/ietf/templates/group/reviewer_overview.html
@@ -93,7 +93,7 @@
                                                     {{ assignment_to_closure_days }} day{{ assignment_to_closure_days|pluralize }}
                                                 {% endif %}
                                             </td>
-                                            <td>
+                                            <td class="text-nowrap">
                                                 {{ doc_name }}
                                                 {% if reviewed_rev %}-{{ reviewed_rev }}{% endif %}
                                             </td>


larseggert (approved):

larseggert:
I approved, but I did point out to Tero that changes like this will make the page unusable on phones.

rjsparks (approved):

kivinen:
Lars Eggert writes:
> I approved, but I did point out to Tero that changes like this will make the
> page unusable on phones.


I disagree on that. The current datatracker reviwers page is
completely usable on phones, only problem is that the dates do not
have nowrap-text class in them, so the phone splits them to three
rows, making the table rows too high. The page will properly scroll
x and y direction so you can see everything just by scroll left and
right.

There is nothing wrong with having page that you need to scroll back
and worth to see completely. Actually with the current page you can
even zoom it out so it will fit on the screen, but of course then the
text is really tiny, but you can still see all the information at one
glance in case you need that.

I hope the new bs5 system does not do the stupid thing that some pages
do, i.e., disable page zooming, and try to fit everything on the
screen. Those pages are usually completely unusable on phone, but
luckily you can usually request a "computer" version of the page, and
get the proper page which you can scroll around and zoom as needed.

(I.e. I tested the current datatracker page on my phone and it works
fine there, I have not tested the sandbox version with phone).
-- 
***@***.***

rjsparks:
the page at sandbox-ng.ietf.org does not scroll horizontally. It does not render correctly (even before this PR's change) in portrait mode - it has to be viewed in landscape mode.

@larseggert - is horizontal scrolling disabled here intentionally? As long as these are done with tables rather than grids, have you considered https://getbootstrap.com/docs/5.0/content/tables/#always-responsive?

PR #3684, opened 2022-03-19
fix: Use an input field for draft submit status search

### LINKED ISSUE #3674: Search box on /submit/status makes little sense
Using the draft-chooser widget at /submit/status is not likely the right thing to do.

The widget is fed from the names of Document objects that already exist. Anything that's being submitted for the first time won't be there.

Anyone who has any business messing with submission status knows the full draft name. The convenience of a slow auto-complete isn't worth essentially the only thing the control might buy.

(It's also very confusing to click on something that looks like you should be able to type into and have the form change and want input somewhere else from where you clicked).

kesara:
Fixes #3674 

Use an input field similar to document search on the navigation bar.
Selecting the draft name will automatically submit the form.

### DIFF
--- a/ietf/templates/submit/search_submission.html
+++ b/ietf/templates/submit/search_submission.html
@@ -9,22 +9,16 @@
         Please enter the name of the Internet-Draft you wish to view the
         submission status of.
     </p>
-    <form method="post">
+    <form id="doc-submit-form" method="post">
         {% csrf_token %}
         <div class="mb-3">
             <label class="form-label">I-D name</label>
-            <select class="form-control select2-field"
+            <input class="form-control select2-field"
                     data-placeholder="draft-..."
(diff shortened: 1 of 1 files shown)


ietf/templates/submit/search_submission.html
```diff
@@ -9,22 +9,16 @@
         Please enter the name of the Internet-Draft you wish to view the
         submission status of.
     </p>
-    <form method="post">
+    <form id="doc-submit-form" method="post">
         {% csrf_token %}
         <div class="mb-3">
             <label class="form-label">I-D name</label>
-            <select class="form-control select2-field"
+            <input class="form-control select2-field"
```
larseggert, line 16:
```suggestion
            <input class="form-control"
```

PR #3686, opened 2022-03-19
fix: Use related docs pages count when no page count

evyncke:
Fixing issue #3503 

This is also my first PR for datatracker, so, please review it carefully even if it is only a dozen of lines ;-)

### DIFF
--- a/ietf/doc/utils_search.py
+++ b/ietf/doc/utils_search.py
@@ -144,7 +144,17 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
         l.append(rel_rfc_aliases[rel.source_id])
         l.sort()
 
+def augment_docs_with_related_docs_info(docs):
+    """Augment all documents with related documents information.
+    At first, it handles only conflict review document page count to mirror the original document page count."""
 
+    for d in docs:
+        if d.type_id == 'conflrev':
+            if len(d.related_that_doc('conflrev')) != 1:
+                continue
+            originalDoc = d.related_that_doc('conflrev')[0].document
+            d.pages = originalDoc.pages
+            
 def prepare_document_table(request, docs, query=None, max_results=200):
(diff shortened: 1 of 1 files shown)


rjsparks:
This is an improvement, and it will help IESGs, and I plan to merge it as-is.

I worry, though, about changing the meaning doc.pages this way - that will be surprising to future coders working on the templates, and it requires cultural training for the community (particularly new IESG members) to know that what the number means on the output page is _different_ for conflict review rows than it is for any other kind of documents.

We should consider a small project where we formally and carefully change the display element on the search result rows that currently holds a page count (which is what this PR affects) to explicitly hold document-specific metrics that have document-specific labels/layout in the search result row template.

It hasn't happened very often but it is possible to get a Conflict Review that by itself is very long, or requires a lot of reading of some _other_thing, even if the document being reviewed is short.

ietf/doc/utils_search.py
```diff
@@ -144,7 +144,17 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
         l.append(rel_rfc_aliases[rel.source_id])
         l.sort()
 
+def augment_docs_with_related_docs_info(docs):
+    """Augment all documents with related documents information.
+    At first, it handles only conflict review document page count to mirror the original document page count."""
 
+    for d in docs:
+        if d.type_id == 'conflrev':
+            if len(d.related_that_doc('conflrev')) != 1:
```
rjsparks, line 153: Right now, the other parts of the datatracker that set up conflict reviews will not let a conflict review be created that has anything other than exactly one reviewed document, so this line is a safety check against misconfiguration or changes to that assumption in the future. That's good.

The code could instead have looped over the reviewed documents and summed their pages, but that just makes another assumption about what having multiple related documents might mean might mean in the future.

PR #3689, opened 2022-03-19
fix: List blocked charters in AD dashboard that are in extrev

### LINKED ISSUE #3588: The AD Dashboard "Blocking positions held by..." section misses (some?) non-draft documents
This feature was requested in #2800 and implemented in https://github.com/ietf-tools/datatracker/commit/6688d37bb4b57f5f42e1ee38d4688b2788a1221e ; there are references in the code to document types of "charter" and "conflrev", but I do not believe the handling for those document types is currently functional.
In particular, I currently hold a BLOCK on charter-ietf-shmoo but it does not show up on my AD Dashboard page.

russhousley:
Fixes #3588

### DIFF
--- a/ietf/doc/views_search.py
+++ b/ietf/doc/views_search.py
@@ -51,7 +51,9 @@ from django.utils.cache import _generate_cache_key # type: ignore
 import debug                            # pyflakes:ignore
 
 from ietf.doc.models import ( Document, DocHistory, DocAlias, State,
-    LastCallDocEvent, NewRevisionDocEvent, IESG_SUBSTATE_TAGS, IESG_BALLOT_ACTIVE_STATES )
+    LastCallDocEvent, NewRevisionDocEvent, IESG_SUBSTATE_TAGS,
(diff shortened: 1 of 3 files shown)


ietf/doc/views_search.py
```diff
@@ -446,6 +447,10 @@ def docs_for_ad(request, name):
         if blocked_docs:
             blocked_docs.sort(key=lambda d: min(p.time for p in d.blocking_positions if p.balloter==ad), reverse=True)
 
+        for d in blocked_docs:
+           if d.get_base_name() == 'charter-ietf-shmoo-01-04.txt':
+              print('Is in list')
```
kaduk, line 452: This looks like debugging code that snuck in

russhousley: Ooops.  I just removed it from my branch.

rjsparks: And my error for letting it in.
See https://github.com/ietf-tools/datatracker/pull/3704

PR #3706, opened 2022-03-20
fix: add more explanations in BoF requests

evyncke:
Fixing issue #3517

### DIFF
--- a/ietf/templates/doc/bofreq/new_bofreq.html
+++ b/ietf/templates/doc/bofreq/new_bofreq.html
@@ -8,14 +8,6 @@
     <p>
         The IAB will also attempt to provide BoF Shepherds as described in their document on the subject only on request from the IESG. If you feel that your BoF would benefit from an IAB BoF Shepherd, please discuss this with your Area Director.
     </p>
-    <p>        To allow evaluation of your proposal, please include the following items:
-        <ul>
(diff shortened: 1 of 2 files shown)


ietf/templates/doc/bofreq/new_bofreq.html
```diff
@@ -5,12 +5,24 @@
 {% block content %}
     {% origin %}
     <h1>Start a new BOF Request</h1>
+    <p>
+        The IAB will also attempt to provide BoF Shepherds as described in their document on the subject only on request from the IESG. If you feel that your BoF would benefit from an IAB BoF Shepherd, please discuss this with your Area Director.
+    </p>
+    <p>        To allow evaluation of your proposal, please include the following items:
+        <ul>
+        <li>Please list any protocols or practices that already exist in this space.</li>
```
rjsparks, line 13: Why is this block outside edit box instead of being part of the text at 
https://github.com/ietf-tools/datatracker/blob/feat/bs5/ietf/templates/doc/bofreq/bofreq_template.md
?

rjsparks (changes requested):
Per offline discussion with éric, he wants to make some changes before this is merged.

rjsparks:
Note also https://github.com/ietf-tools/datatracker/runs/5617870056?check_suite_focus=true#step:6:48

PR #3709, opened 2022-03-20
fix: use better address calculation for schedule owner. Fixes #2611.

### LINKED ISSUE #2611: Poor choice of email used for identity when saving a copy of an agenda
`type_defect`   |    by rjsparks@nostrum.com

___

When I save a copy of an agenda in the drag-drop editor, it names it with an inactive email address:
```
/meeting/103/agenda/robert.sparks@tekelec.com/rjstest/edit
```
It should use the code that chooses the "best" address instead (which will return the address marked primary when there is one).

___
_Issue migrated from trac:2611 at 2022-03-04 06:54:05 +0000_

rjsparks:


jennifer-richards (approved):
Much better!

The failing test in the automated checks is one of the intermittent JS ones.

PR #3719, opened 2022-03-21
test: scroll swap timeslots button into view before clicking

### LINKED ISSUE #3718: test_past_swap_days_buttons failing frequently
### What happened?

On a couple of test runs on the feat/bs5 branch, the test_past_swap_days_buttons javascript test has failed. Intermittent failures similar to this have been seen in the past, but should investigate whether this is a real failure and/or whether it can be made more robust.

Frequently the problem has been layout changes pushing elements on top of each other or off screen in the viewport used by selenium.

Not Applicable

### Code of Conduct

jennifer-richards:
Fixes #3718

### DIFF
--- a/ietf/meeting/tests_js.py
+++ b/ietf/meeting/tests_js.py
@@ -475,6 +475,14 @@ class EditMeetingScheduleTests(IetfSeleniumTestCase):
         # option to swap. If we used the first or last day, a fencepost error in
         # disabling options by date might be hidden.
         clicked_index = 1
+        # scroll so the button we want to click is just below the navbar, otherwise it may
+        # fall beneath the sessions panel
+        navbar = self.driver.find_element_by_class_name('navbar')
+        self.driver.execute_script(
+            'window.scrollBy({top: %s, behavior: "instant"})' % (
+                    future_swap_days_buttons[1].location['y'] - navbar.size['height']
+            )
+        )
         future_swap_days_buttons[clicked_index].click()
         try:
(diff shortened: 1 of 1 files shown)


ietf/meeting/tests_js.py
```diff
@@ -475,6 +475,14 @@ def test_past_swap_days_buttons(self):
...
         # disabling options by date might be hidden.
         clicked_index = 1
+        # scroll so the button we want to click is just below the navbar, otherwise it may
+        # fall beneath the sessions panel
+        navbar = self.driver.find_element_by_class_name('navbar')
+        self.driver.execute_script(
+            'window.scrollBy({top: %s, behavior: "instant"})' % (
+                    future_swap_days_buttons[1].location['y'] - navbar.size['height']
+            )
+        )
```
larseggert, line 485: Does `self.scroll_to_element` not work?

jennifer-richards:
No, unfortunately. I think that only scrolls until the element is in the
viewport. The session panel is still floating over it, intercepting clicks.

ietf/meeting/tests_js.py
jennifer-richards: No, unfortunately. I think that only scrolls until the element is in the viewport. The session panel is still floating over it, intercepting clicks.

(Sorry to repeat, but this showed up as a top level comment rather than a reply to your question)

rjsparks (approved):

PR #3721, opened 2022-03-21
test: Add tests for submission file validation

russhousley:
Add length check to validate_submission_rev, add test for validate_submission_rev, and add test for validate_submission_name


rjsparks:
This commit went in as part of https://github.com/ietf-tools/datatracker/pull/3729 instead.

PR #3724, opened 2022-03-22
ci: Add vnu to dev docker image

larseggert:


larseggert:
Whoops - why is the diff so huge?

rjsparks:
This currently assumes an x86 architecture on the developer's machine.

larseggert:
Oh. Yes. The alternate way to do this is to install the JDK from `apt` and then just pull the JAR. Will do that.

NGPixel:
> Whoops - why is the diff so huge?

The original file was incorrectly created using CRLF instead of just LF (thanks Windows). So this is ok.

NGPixel (approved):

PR #3738, opened 2022-03-23
feat: improve release information in footer.

rjsparks:


ietf/__init__.py
```diff
@@ -17,8 +17,6 @@
...
 __patch__   = ""
 
-__date__    = "$Date$"
-
-__rev__     = "$Rev$ (dev) Latest release: Rev. 19974 "
-
-__id__      = "$Id$"
+# Testing - do not commit
+__release_hash__ = 'd489391b66bd67d8d702d82ef9f21006e9415385'
+__release_branch__ = 'feat/bs5'
```
NGPixel, line 22: Should these values be hardcoded, instead of left empty?

rjsparks: marking this draft until @larseggert  has a chance to look

rjsparks: nevermind waiting for lars.

PR #3739, opened 2022-03-23
chore: remove svn/trac infrastructure (#3649)

NGPixel:
Port of #3649 by @rjsparks for bs5 branch. I had to make some manual merging in `ietf/utils/tests.py` as there were conflicts, so please review this file in particular.


rjsparks (approved):
This all looks good. The only difference I see in the net effect of the changes to ietf/utils/tests.py are whitespace, and they're fine.

PR #3752, opened 2022-03-25
chore: merge updates from main into feat/bs5 branch

jennifer-richards:


docker/app.Dockerfile
```diff
@@ -131,6 +131,9 @@ ADD https://github.com/validator/validator/releases/download/latest/vnu.jar /
 ADD https://raw.githubusercontent.com/eficode/wait-for/v2.1.3/wait-for /usr/local/bin/
 RUN chmod +rx /usr/local/bin/wait-for
 
+ADD https://raw.githubusercontent.com/eficode/wait-for/v2.1.3/wait-for /usr/local/bin/
```
rjsparks, line 134: This looks like it duplicates a command already run?

ietf/templates/release/about.html
```diff
@@ -1,5 +1,6 @@
 {# Copyright The IETF Trust 2015-2022, All Rights Reserved #}
 {% extends "base.html" %}
+{# Copyright The IETF Trust 2015-2022, All Rights Reserved #}
```
rjsparks, line 3: Another add of a line that already exists.

rjsparks (changes requested):

ietf/templates/release/about.html
jennifer-richards: removing

rjsparks (approved):

NGPixel (approved):

PR #3759, opened 2022-03-29
fix: ignore exceptions when comparing imported to existing minutes

### LINKED ISSUE #3749: "Import minutes" fails if trying to replace PDF formatted minutes
### What happened?

Someone attempted to import minutes for a session that already had minutes. This failed with a top-level `UnicodeDecode` error, triggered when attempting to compare the version to be imported with the existing minutes.

The failure resulted from `ietf/meeting/views.py in import_session_minutes` on
```
                if import_contents == Note.preprocess_source(f.read()):
```
The local variables reported in the crash report show that the value of `import_contents` is valid, but the existing file, `f`, is a PDF. The `UnicodeDecode` is failing because it is trying to parse binary contents of the PDF header.

The import view needs to be smarter about trying to compare files here, and should catch unexpected exceptions instead of throwing a 500 error.

In the meantime, uploading a text file to replace the PDF, then importing should work around this issue. Directly copy/pasting the minutes instead of importing works around as well.

_No response_

### Code of Conduct

jennifer-richards:
Fixes #3749

### DIFF
--- a/ietf/meeting/tests_views.py
+++ b/ietf/meeting/tests_views.py
@@ -6112,7 +6112,32 @@ class ImportNotesTests(TestCase):
             self.assertContains(r, "This document is identical", status_code=200)
             q = PyQuery(r.content)
             self.assertEqual(len(q('button:disabled[type="submit"]')), 1)
-            self.assertEqual(len(q('button:not(:disabled)[type="submit"]')), 0)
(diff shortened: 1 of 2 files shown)


jennifer-richards:
Oops - will replace this with a new PR in a bit.

PR #3772, opened 2022-03-30
fix: left-align date columns in all tablesorter-classed tables.

rjsparks:


rjsparks:
@larseggert This is what we discussed aligning dates so that visual scanning works when tables contain both "yyyy-mm" and "yyyy-mm-dd"

jennifer-richards (approved):

PR #3774, opened 2022-03-31
fix: restore function of unassigned review management page

jennifer-richards:
Fixes ietf-tools#3685

N.B., the first three commits are just refactoring/StandardJS-ifying the affected javascript module. Latter two commits have the interesting changes.

### DIFF
--- a/ietf/static/js/manage-review-requests.js
+++ b/ietf/static/js/manage-review-requests.js
@@ -2,17 +2,17 @@
 $(function () {
   'use strict'
   const form = $('form.review-requests')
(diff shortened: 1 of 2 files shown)


jennifer-richards:
Converting to draft while I fix the failing test.

jennifer-richards:
I think that must have been a spurious failure. I don't see it locally and don't see how this change could have affected it.

rjsparks (approved):
I think we still need to work on the form layout, but this is so much better than what we had

PR #3777, opened 2022-04-01
fix: render meeting schedule editor icons using bootstrap-icons

### LINKED ISSUE #3557: Schedule editor constraint icons not shown
`type_defect`   |    by jennifer@painless-security.com

___

The small icons for session constraints in the schedule editor are no longer shown. I believe this is because they are still using the font-awesome classes. These should be updated to equivalent or similar bootstrap icons.

This affects the bootstrap5 revamp

___
_Issue migrated from trac:3557 at 2022-03-04 09:29:12 +0000_

jennifer-richards:
Moves the the HTML that had lived in the `editor_label` field in the DB into a Django template tag.

Fixes #3557


rjsparks (approved):

jennifer-richards:
[rjsparks commented](https://github.com/ietf-tools/datatracker/pull/3778#pullrequestreview-929574465)
> My only concern is that his introduces a schema change, so unless we also work it back into main, it wouldn't be possible to revert to (pre-bs5-merge) main in a crunch. It may not be worth the effort to do so, but choosing not to do so makes this the point where we close the door on the possibility.

[and](https://github.com/ietf-tools/datatracker/pull/3778#issuecomment-1086649764)
> (Yes, rolling back the migrations are supposed to be enough, but...)

The migrations are purely cleanup. We could separate those and make the schema change once we're confident that we won't need to roll back.

jennifer-richards:
The last two commits revert the schema changes. Those are available to be merged separately whenever it's safe over at [jennifer/drop-editor-label](https://github.com/painless-security/datatracker/tree/jennifer/drop-editor-label) on the painless-security fork.

rjsparks:
I think this could be merged then.

Why not make a PR from drop-editor-label and leave it as draft for now?

PR #3778, opened 2022-04-01
feat: persist timeslot types made visible in the schedule editor

### LINKED ISSUE #3590: Persistent meeting schedule editor view settings
Usability of the schedule editor can be improved by making the timeslot type / session purpose visibility settings persistent. At present, navigating away from the editor then returning will normally cause these settings to be reset to default. This adds many additional UI steps to tasks like editing session details.

jennifer-richards:
Fixes #3590


rjsparks (approved):
My only concern is that his introduces a schema change, so unless we also work it back into main, it wouldn't be possible to revert to (pre-bs5-merge) main in a crunch. It may not be worth the effort to do so, but choosing not to do so makes this the point where we close the door on the possibility.

rjsparks:
(Yes, rolling back the migrations are supposed to be enough, but...)

rjsparks:
and of course, those comments would be better attached to #3777

PR #3780, opened 2022-04-04
fix: put unassigned review assignment contols on one small line

rjsparks:


rjsparks:
Be great if Lars can review this short one as it's all about tweaking how the form uses bootstrap5

ietf/templates/group/manage_review_requests.html
```diff
@@ -180,14 +180,14 @@ <h1>
...
                             </div>
-                            <div class="reviewer-controls mx-1 flex-fill">
+                            <div class="reviewer-controls mx-1 flex-fill row">
                                 {% if r.form.review_type %}
-                                    {% bootstrap_field r.form.review_type layout="horizontal" size="sm" %}
+                                    {% bootstrap_field r.form.review_type layout="horizontal" wrapper_class="col" size="sm" %}
                                 {% endif %}
-                                {% bootstrap_field r.form.reviewer layout="horizontal" size="sm" %}
-                                {% bootstrap_field r.form.add_skip layout="horizontal" size="sm" %}
+                                {% bootstrap_field r.form.reviewer layout="horizontal" wrapper_class="col row" size="sm" %}
```
jennifer-richards, line 187: I believe the `row` in the `wrapper_class` is redundant with the `layout="horizontal"`.

jennifer-richards (approved):
Looks good to me. One minor tweak suggested inline.

The "Close" button opens a similar set of fields. Should those receive the same single-row treatment? (Don't recall whether they were single row in the bs3 implementation.)

ietf/templates/group/manage_review_requests.html
rjsparks: It isn't, though I agree it should perhaps have worked, but there are (I'm guessing) interactions with the fluid form classes (row and col) that defeat what layout='horizontal' does by itself.
To be clear - layout='horizontal' was already present, and was insufficient, even after adding "col".

rjsparks: Also - the 'close' form is one line on main, but I think having it take more than one line makes sense - it's not used quite as often, and it makes more sense for the reason field to take some space.

Issue #3757, opened 2022-03-28
Misleading "meeting" time zone on upcoming meetings page UI


jennifer-richards:
### What happened?

The upcoming meetings page (`/meeting/upcoming`) has a timezone selection UI similar to the meeting agenda page. It offers three fixed options ("Meeting", "Local", and "UTC"), plus a select box to choose a custom timezone. Selecting "Meeting" actually sets the time zone to the browser local time zone, just the same as the "Local" option.

The equivalent UI on the bootstrap 3 branch simply omits the "Meeting" time zone choice. This is probably the right thing to do - it's unclear why anyone would want to view a list of meetings with their times shown in potentially different meeting-local time zones.

Need to coordinate this with fixing #3737, which will also affect the time zone UI.

### What browser(s) are you seeing the problem on?

Not Applicable

### Code of Conduct

- [X] I agree to follow the IETF's Code of Conduct

Issue #3737, opened 2022-03-23
Timezone display / quick-change links are absent from agenda nav panel


jennifer-richards:
### What happened?

In #3698, @kivinen mentioned:
> Also the showing which timezone is used when shown and providing quick links to change timezones are missing.

These should be added back.

### What browser(s) are you seeing the problem on?

Not Applicable

### Code of Conduct

- [X] I agree to follow the IETF's Code of Conduct

PR #3792, opened 2022-04-07
fix: restore time zone widget to agenda RH nav panel

### LINKED ISSUE #3700: Timezone picker does not properly update timezone on popup
### What happened?

On both on http://localhost:8000/meeting/agenda/ and http://localhost:8000/meeting/upcoming page the timezone picker does not properly update the popup menu. I.e., if I select for example "Japan" from that list it will properly change the timezone to Japan. Now if I click on the UTC button the timezone popup stays unchanged, just the UTC item gets highlighted to indicate we have now selected UTC. The Popup still says japan. Old agenda before bs5 changed the popup text to match the "UTC" or whatever the timezone for local / meeting was, but bs5 version does not update the popup. 

This is fine, but now when I click on the popup again, I would expect it to be centered around the Japan, i.e., the text it was showing and so I could quickly pick it again, but it is actually centered over UTC or whatever I have picked from the left last.

So as I can't use the picker as one more option to quickly swap between then keeping the misleading value there is bad.

Either fix it so that the popup menu item shown is always updated when meeting / local / UTC is selected, so it is easy to see which timezone is used, or change it so that when you click the popup open it will be centered around the last entry user picked...

_No response_

### Code of Conduct

jennifer-richards:
This restores the functionality of the agenda time zone selection widget, extends it to support a compact mode, and adds that compact display to the RH nav panel. This restores a feature that was dropped in the initial bs5 facelift.

Creating as a draft because this builds on #3790. Once that is approved/merged, I'll straighten this out.

Fixes #3700, #3737, and #3757.


rjsparks (approved):
awesome!

larseggert:
Why is the righthand-nav a card now? That makes no sense structurally and also introduces styling oddities. (Of which there are other ones, e.g., the nav content is now vertically centered?)

Also, the timezone picker needs to be layed out vertically or at least wrap correctly
Screen Shot 2022-04-08 at 11 27 50
:

jennifer-richards:
What is the structural problem with a card? As far as I can tell, it's a utility component that makes content-in-a-box. What's there is (loosely) modeled on [the kitchen sink example](https://getbootstrap.com/docs/5.1/components/card/#kitchen-sink).

The vertical alignment is because of a flexbox in need of an align-items-start.

There's some styling strangeness tied to the timezone picker being too wide. What other oddities are you referring to?

I'm not thrilled at the prospect of making the timezone picker vertical because it will be less distinct from the nav links above it. It does need to behave better. It could also revert to being text links like it was before. That's not as snazzy, but the display immediately above it already indicates the timezone so the fancy highlighting is not essential.

larseggert:
Also see #3798.

Regarding card, I was under the impression that is mostly for displaying series of identical/similar things, with some convenient formatting that applies across the series. The added border and whitespace (to me) clash with the leftmenu, which doesn't have those.

Regarding oddities, yes, that the content no longer wraps. I agree it doesn't look great when the timezone picker does, but at least it would remain usable that way.

PR #3797, opened 2022-04-08
fix: Update the document shepherd template

larseggert:
Along the lines of https://github.com/ietf-chairs/chairs.ietf.org/pull/5


rjsparks (changes requested):
There are a lot of nasty breaks in the text that are supposedly setting up links. Breaking between RFC or BCP and the number.

This is used as a text blob, not as markdown (yet). It might be better to not try to use that markdown formatting until what this captures is rendered as markdown.

rjsparks (approved):

PR #3812, opened 2022-04-12
Fix urlize ietf docs

larseggert:
The old code would first substitute the `rfx<nnnn>` part, and then again
on the full draft name. This now only ever applies one substitute,
and does longer ones first.

### DIFF
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -192,23 +192,15 @@ def urlize_ietf_docs(string, autoescape=None):
     """
     if autoescape and not isinstance(string, SafeData):
         string = escape(string)
-    patterns = [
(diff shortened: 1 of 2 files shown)


larseggert:
You're right; I didn't check a block of text with multiple replaceable things. Let me rework this.

larseggert:
@rjsparks take another look?

rjsparks:
Still not where you want to be I think. You (or I could if you want) should really add a unittest for this urlize function.
```
>>> re.sub(r"(?<!>)\b(RFC\s*?)0{0,3}(\d+)\b", "<a href=\"/doc/rfc\\2/\">\\1\\2</a>", "draft-ietf-rfc9999-bis-01", flags=re.IGNORECASE)
'draft-ietf-<a href="/doc/rfc9999/">rfc9999</a>-bis-01'
```

larseggert:
@rjsparks over to you :-)

rjsparks (approved):

PR #3819, opened 2022-04-13
chore: switch to Yarn 3

NGPixel:
Replaces NPM with Yarn 3 as the package manager.

This new setup enables zero-install and PNP capabilities:
- Nothing to install (no npm install) for new users or when fetching new code. Enables fully offline use.
- Precompiled dependencies for any combination of macOS, Linux, Windows + x64, arm64 are included.
- No node_modules black hole:
image


phitattoo (approved):
ngpixel:yarn-berry

phitattoo (approved):

phitattoo (approved):

ietf/static/js/datepicker.js
```diff
@@ -1 +1,2039 @@
-import "bootstrap-datepicker/js/bootstrap-datepicker.js";
```
larseggert, line 1: This diff seems to pull in the entire source, rather than include by reference?

NGPixel: Yes, as it's been causing issues in some scenarios.

Because that package is being fetched directly from GitHub, it uses ssh when cloning the repo. That process sometime fails because of incorrect ssh settings on the host.

There's also the issue that there's no versioning as it always fetch whatever is currently on GitHub. A change there could easily break the build.

So while not ideal, including the source directly is the best option for now to address both issues. These files will eventually go away once we get rid of jquery.

larseggert: We are planning to get rid of jquery?

NGPixel: At some point yes, as pages get refactored to more modern frameworks like Vue.js.

There's really no reason nowadays to use jquery. All browsers implement the querySelector methods. In fact, Bootstrap 5 no longer relies on jquery either.

Other bits that we currently use like scroll to element and scroll spy have native implementations now via scrollIntoView() and the IntersectionObserver class. Both offers much better performance than their jquery counterparts.

rjsparks (approved):

larseggert:
I just realized that this commit removed the caret from the version numbers in `package.json`. @NGPixel, how do we now upgrade the assets? A bunch of them had bugfix releases in the meantime.

NGPixel:
@larseggert You can upgrade packages using the command `yarn upgrade-interactive`, which will display something similar to:

image

PR #3821, opened 2022-04-13
chore: tweak sqlite timeout during tests

rjsparks:


rjsparks:
not the right path to pursue

PR #3822, opened 2022-04-14
fix: add `flex-wrap` to `pagination` elements

larseggert:
They seem to be one of the only bs5 elements that do not by default wrap, which
causes display issues on narrow screens when there are lots of page elements,
e.g., on http://127.0.0.1:8000/doc/draft-ietf-ipwave-vehicular-networking/

Adding `flex-wrap` (per https://github.com/twbs/bootstrap/issues/23504) makes
this better, but unfortunately doesn't fix it completely, because the layout
still jumps when the window is narrowed by the user. It does fix the issue on
first load and after a reload though.


rjsparks (approved):

mmm8955405:
https://github.com/twbs/bootstrap/issues/24159
Why not try more paging schemes? Your paging primary key hasn't changed for many years

larseggert:
@mmm8955405 I don't quite understand what the suggestion is? What other paving schemes?

PR #3824, opened 2022-04-14
style: indent_style = space

larseggert:


NGPixel:
@larseggert just a quick note on the "fix" commit prefix usage. It only applies to app code that actually ends up into production; not dev / test / build configuration files.

In this case, either "style" or "chore" prefix would be more appropriate. See https://github.com/ietf-tools/.github/blob/main/CONTRIBUTING.md#git-commit-messages for an explanation of each option.

PR #3831, opened 2022-04-15
fix: improve ietf doc string urlization. Fixes #3828.

### LINKED ISSUE #3828: new urlizer breaks on draft-xxxx-xxxx-00.txt
At /doc/draft-calhoun-enh-radius-filter/history/
```
 <td>New version available: <b><a href="/doc/draft-calhoun-enh-radius-filter-00/">draft-calhoun-enh-radius-filter-00</a>.txt</b></td>
```

rjsparks:


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -185,15 +185,44 @@ def rfceditor_info_url(rfcnum : str):
...
+    elif re.search(exp2, string): 
+        string = re.sub(
+            exp2,
+            lambda x: f'<a href="/doc/{x[1][:-1]}/{x[2]}/">{x[0]}</a>',
+            string,
+            flags=re.IGNORECASE | re.ASCII,
+        )
+    string = re.sub(
+        r"\b(?<![/-])(((?:draft-|bofreq-|conflict-review-|status-change-)(?:[\d\w\.+]+-)*)([\d\w\.+]+?)(\.txt)?)\b",
+        lambda x: link_non_charter_doc_match(x),
```
jennifer-richards, line 219: Could do away with the lambda here and simplify to `link_non_charter_doc_match`

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -185,15 +185,44 @@ def rfceditor_info_url(rfcnum : str):
...
+        string,
+        flags=re.IGNORECASE | re.ASCII,
+    )
     string = re.sub(
-        r"\b((RFC|BCP|STD|FYI|(?:draft-|bofreq-|conflict-review-|status-change-|charter-)[-\d\w.+]+)\s*0*(\d+))\b",
+        # r"\b((RFC|BCP|STD|FYI|(?:draft-|bofreq-|conflict-review-|status-change-|charter-)[-\d\w.+]+)\s*0*(\d+))\b",
+        r"\b(?<!-)((RFC|BCP|STD|FYI)\s*0*(\d+))\b",
```
jennifer-richards, line 225: Intentional that the exclusion is just `(?<!-)` vs `(?<![/-])` in the previous case?

jennifer-richards (approved):
A comment and a question, but looks good.

ietf/doc/templatetags/ietf_filters.py
rjsparks: yeah - silly

ietf/doc/templatetags/ietf_filters.py
rjsparks: so far, yes. Trying not to complicate the matches without positive need to do so.
Adding / in the negative lookbehind here, would cause any text that had a string like this
```
According to RFC9998/BCP99 the foobars are
```
to not link the second thing.

PR #3835, opened 2022-04-18
fix: do not prematurely dereference status change RelatedDocuments

jennifer-richards:
Fixes crash when displaying RFCs with status change documents.

Also adds tests to cover this and the change to the `person_link` tag in #3832.

### DIFF
--- a/ietf/templates/doc/document_draft.html
+++ b/ietf/templates/doc/document_draft.html
@@ -65,10 +65,10 @@
                         {% if obsoletes %}<div>Obsoletes {{ obsoletes|urlize_related_target_list|join:", " }}</div>{% endif %}
                         {% if updates %}<div>Updates {{ updates|urlize_related_target_list|join:", " }}</div>{% endif %}
                         {% if status_changes %}
-                            <div>Status changed by {{ status_changes|join:", "|urlize_related_source_list }}</div>
+                            <div>Status changed by {{ status_changes|urlize_related_source_list|join:", " }}</div>
                         {% endif %}
                         {% if proposed_status_changes %}
-                            <div>Proposed status changed by {{ proposed_status_changes|join:", "|urlize_related_source_list }}</div>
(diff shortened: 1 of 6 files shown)


ietf/templates/doc/document_draft.html
```diff
@@ -65,10 +65,10 @@
...
                         {% if updates %}<div>Updates {{ updates|urlize_related_target_list|join:", " }}</div>{% endif %}
                         {% if status_changes %}
-                            <div>Status changed by {{ status_changes|join:", "|urlize_related_source_list }}</div>
+                            <div>Status changed by {{ status_changes|urlize_related_source_list|join:", " }}</div>
                         {% endif %}
                         {% if proposed_status_changes %}
-                            <div>Proposed status changed by {{ proposed_status_changes|join:", "|urlize_related_source_list }}</div>
+                            <div>Proposed status changed by {{ proposed_status_changes|urlize_related_source_list|join:", " }}</div>
                         {% endif %}
                         {% if rfc_aliases %}<div>Also known as {{ rfc_aliases|join:", "|urlize_ietf_docs }}</div>{% endif %}
```
rjsparks, line 73: Same bug on this line, though currently it won't be irritated as the length of rfc_aliases will be 0 or 1 until we handle bcps correctly

rjsparks (approved):

ietf/templates/doc/document_draft.html
rjsparks: eh - no - urlize_ietf_docs will work just fine on a list.

jennifer-richards: Yeah, I chased that one for a bit - I think it's used a few places like this. It's a bit of a shame to convert a full-blown Document into its names just to parse those back into links that could have been generated directly from the Document, but it works. It'd be a nice thing to straighten out some time.

PR #3837, opened 2022-04-19
fix: remove urlization from document history page.

rjsparks:


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -426,8 +426,9 @@ def ad_area(user):
 def format_history_text(text, trunc_words=25):
     """Run history text through some cleaning and add ellipsis if it's too long."""
     full = mark_safe(text)
-    if "</a>" not in full:
-        full = urlize_ietf_docs(full)
+    # The content of docevent.desc is full of things that urlize incorrectly, producing a massive number of new 404s
+    #if "</a>" not in full:
+    #    full = urlize_ietf_docs(full)
```
jennifer-richards, line 431: Looking at the history, I think the `urlize_ietf_docs()` has been there for a long time and the `bleach.linkify()` is new. Are you sure it's the `urlize` that's causing the 404s? On the `doc/draft-ietf-ospf-transition-to-ospfv3/history/` example there were some bad links caused by the `linkify` where an email address wrapped. These probably did not show up as 404s though (the ones I noticed were `mailto`)

The `if` around the `urlize` is also new - it was previously just an unconditional call.

rjsparks: yes, urlize is responsible for _most_ of the 404s
There are some many remaining, but the percentage went _way_ down.

jennifer-richards (approved):
Looks ok to me.

larseggert:
I fixed a bunch of urlization and 404 issues in #3891 - take a look?

The main remaining issue is that `person_link` creates links to some non-existing DT profile pages.

larseggert (commented):
I think this PR is OBE?

rjsparks:
agreed.

PR #3842, opened 2022-04-20
fix: Fix OIDC authorization bug

kesara:


rjsparks:
@larseggert : see  https://github.com/juanifioren/django-oidc-provider/blob/f0daed07b2ac7608565b80d4c80ccf04d8c416a8/oidc_provider/views.py#L177

larseggert:
Ah, sorry for the breakage! I auto-converted a bunch of `input[type=button]` elements to `button`, which caused this. We need a test case :-)

larseggert (approved):

rjsparks (approved):

PR #3862, opened 2022-04-21
feat: add analytics infrastructure

rjsparks:


rjsparks:
This adds generic plumbing for a number of analytics frameworks. We will only use the ietf Matomo instance, by configuring MATOMO_* variables in settings_local on production. Without these settings, the framework does nothing.

jennifer-richards (approved):

PR #3883, opened 2022-04-26
feat: Make the "report a bug" links more visible

larseggert:
Also rework the page footer somewhat, while we're here.


rjsparks:
I'm not sure this is enough.
(I can live with the suggested footer rework).
I think we may need (for awhile) to put report-a-bug in the menus somewhere, or at least on the menu bar

larseggert:
It's in the "other" menu

rjsparks:
and yet we _still_ have chairs asking "how do I report a bug?" regularly.

I'm thinking of something more like:

image

unless you think that's just too obnoxious.

rjsparks:
If we go that route, we could place it only for logged in users to start with

rjsparks (approved):

PR #3891, opened 2022-04-26
More HTML validation & fixes

larseggert:
Optionally, also validates with vnu, which finds more issues but is also a lot slower, hence not enabled by default.

PR also fixes a bunch of HTML issues vnu found.


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -765,3 +833,16 @@ def absurl(viewname, **kwargs):
     Uses settings.IDTRACKER_BASE_URL as the base.
     """
     return urljoin(settings.IDTRACKER_BASE_URL, urlreverse(viewname, kwargs=kwargs))
+
+
+@register.filter
+def is_valid_url(url):
+    """
+    Check if the given URL is syntactically valid/
```
larseggert, line 841:
```suggestion
    Check if the given URL is syntactically valid
```

ietf/doc/templatetags/tests_ietf_filters.py
```diff
@@ -9,6 +9,7 @@
 
 class IetfFiltersTests(TestCase):
     def test_urlize_ietf_docs(self):
+        self.maxDiff = None
```
larseggert, line 12:
```suggestion
```

rjsparks (approved):
There's a lot in ietf_filters that isn't being tested (the "documents always exist when testing" logic branch bypasses them).  Please look at adding unit tests (or mock around the condition) so we know when this code breaks.

PR #3907, opened 2022-04-28
doc: update manual deploy instructions

rjsparks:


dev/INSTALL
```diff
@@ -19,7 +19,7 @@ General Instructions for Deployment of a New Release
  2. Fetch the release tarball from github 
     (see https://github.com/ietf-tools/datatracker/releases)::
 
-      wget https://github.com/ietf-tools/datatracker/archive/refs/tags/${releasenumber}.tar.gz
+      wget https://github.com/ietf-tools/datatracker/releases/download/${releasenumber}/release.tar.gz
       tar xzvf ${releasenumber}.tar.gz
```
jennifer-richards, line 23: now `tar xzvf release.tar.gz`

jennifer-richards (approved):
I believe one filename needs to be updated (see inline comment). Approved otherwise.

larseggert:
There is also `--ignore=bower.json` somewhere in that doc that seems superfluous now.

PR #3908, opened 2022-04-28
chore: use and expect python39

rjsparks:


NGPixel (approved):

jennifer-richards (approved):

larseggert:
Nice! Why not 3.10 though?

larseggert:
Might also want to run [`pyupgrade`](https://github.com/asottile/pyupgrade) over the tree?

rjsparks:
3.10 isn't available on the production system.
We'll move there later, but it will likely be when we containerize the datatracker for production deployment.

PR #3914, opened 2022-04-29
fix: re-align html and view for submit-to-iesg. Fixes #3913

### LINKED ISSUE #3913: Submit to IESG is not working
When chairs try to "Submit to IESG for publication", nothing happens.

rjsparks:


jennifer-richards (approved):

rjsparks:
See the corresponding line in the view at https://github.com/rjsparks/datatracker/blob/fix-3913/ietf/doc/views_draft.py#L581

PR #3915, opened 2022-04-29
fix: guard against reviews without reviewassignments. Fixes #3906.

### LINKED ISSUE #3906: Link to review document without corresponding assignment returns server error
### What happened?

Crash reports came in for [/doc/review-ietf-v6ops-slaac-renum-03-secdir-lc-wierenga-2020-09-24/](https://sandbox.ietf.org/doc/review-ietf-v6ops-slaac-renum-03-secdir-lc-wierenga-2020-09-24/).  (This is a sandbox link - effectively the same behavior is seen on production)

The error reported on production is as follows (sandbox is the same, except with a debug message instead of an empty `name` keyword argument)
```
django.urls.exceptions.NoReverseMatch: Reverse for 'ietf.doc.views_doc.document_main' with keyword arguments '{'name': ''}' not found. 1 pattern(s) tried: ['doc/(?P<name>[A-Za-z0-9._+-]+?)(?:/(?P<rev>[0-9]{1,2}(-[0-9]{2})?))?/$']
```

Poking, it seems that `assignments = ReviewAssignment.objects.filter(review__name=doc.name)` on l.708 of `views_doc.py` is empty, which makes its way to `review_req` being `None`.

Not Applicable

### Code of Conduct

rjsparks:


rjsparks:
Also fixes a typo that broke things but no-one complained. I wonder if we can find such typos programmatically yet?

jennifer-richards (approved):

PR #3920, opened 2022-05-02
ci: add db migrations checks to tests and build workflows

### LINKED ISSUE #3919: cicd: PR checks should look to see if the PR changes models without creating corresponding migrations
The old mkrelease script had this check:
```
note "Checking that all model changes have been captured in migrations ..."
$do ietf/manage.py makemigrations | tee /dev/stderr | $do grep -q "^No changes detected$" || die "Model changes without migrations found."
```

We should have something similar in the tests that run over PRs, and again when cutting releases.

NGPixel:
closes #3919

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,6 +128,8 @@ jobs:
       run: |
         echo "Running checks..."
         ./ietf/manage.py check
+        echo "Validating migrations..."
+        $do ./ietf/manage.py makemigrations | tee /dev/stderr | $do grep -q "^No changes detected$" || die "Model changes without migrations found."
         echo "Running tests..."
         ./ietf/manage.py test --settings=settings_sqlitetest
         coverage xml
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -128,6 +128,8 @@ jobs:
       run: |
         echo "Running checks..."
         ./ietf/manage.py check
+        echo "Validating migrations..."
+        $do ./ietf/manage.py makemigrations | tee /dev/stderr | $do grep -q "^No changes detected$" || die "Model changes without migrations found."
```
rjsparks, line 132: $do (in several places here) no longer makes sense?

rjsparks (changes requested):

.github/workflows/build.yml
rjsparks: die is also locally defined in the source script.

NGPixel: ok, I was wondering what $do was used for. I assumed it was some special magic bash keyword...
I rework the code.

rjsparks (approved):

PR #3934, opened 2022-05-04
More HTML nitfixing

larseggert:


rjsparks (approved):

rjsparks:
The test failure at event wrapping already exists on main

PR #3945, opened 2022-05-10
chore: Remove some unneeded things from .gitignore

larseggert:


.gitignore
```diff
@@ -26,39 +10,16 @@ datatracker.sublime-workspace
 /.Python
 /.settings
 /.tmp
-/attic
-/bin
-/bootstrap/node_modules
 /data
 /docker/docker-compose.extend-custom.yml
-/env
```
rjsparks, line 34: I'm not aware of anyone attempting native installs for development anymore, but if there is anyone, this will be annoying

.gitignore
```diff
@@ -26,39 +10,16 @@ datatracker.sublime-workspace
...
-/etc
 /ghostdriver.log
-/htmlcov
```
rjsparks, line 37: This should come back. It's the output of `coverage html` for when you want to look at the coverage results locally.

rjsparks, line 37: This should come back. It's the output of `coverage html` for when you want to look at the coverage results locally.

rjsparks (approved):

PR #3946, opened 2022-05-10
refactor: update syntax to python-3.9

larseggert:
Via `pyupgrade --py39-plus $(find * -name \*py)`.

In case we see a value here - up for discussion.


rjsparks:
Kesara, Nick, Jennifer, Ryan and I talked about this today during a dev-focused call. Our collective take is that, like running black on the codebase, it will be too destructive to history (git blame) for the benefit it would bring.

Nick will be looking at whether we could run it over just changed files (lines would probably be too much to ask) and nudge the PR submitter to consider improvements.

larseggert:
I'm fine with this decision, but I'd personally rather bite the bullet and do a one-time sweeping change for this and to reformat, and then enforce that new PRs conform. YMMV.

PR #3952, opened 2022-05-12
fix: escape the ballot comments

larseggert:
I wonder if other fields that have class `pasted` would need to be escaped, too?


rjsparks:
do you think this was a bs5 regression?
Double-escaping was a huge problem several years ago. I hope we're not back into a period of fighting that bit of annoyance.
We really should start working towards a future where things like ballots (charter, etc) are markdown...

larseggert:
> do you think this was a bs5 regression?

I don't know. It might be.

rjsparks (approved):

Issue #3841, opened 2022-04-19
Selecting last letter of WG chairs last name doesn't scroll on Chrome


paulehoffman:
### What happened?

On [chair photos](https://datatracker.ietf.org/wg/photos/), select a letter such as "H". The URL correctly changes to [this](https://datatracker.ietf.org/wg/photos/#H), but the page does not scroll.

### What browser(s) are you seeing the problem on?

Chrome / Edge

### Code of Conduct

- [X] I agree to follow the IETF's Code of Conduct

paulehoffman:
It did use to work on Chrome, my (faulty) memory tells me. It works on Firefox for me as well.

paulehoffman:
It is failing for me in Chrome 100.0.4896.127 on Mac.

Also fails in Microsoft Edge version 100.0.1185.36 (Official build) (64-bit) on Windows 11.

larseggert:
@paulehoffman could you check if the problem is still there?

paulehoffman:
Thanks for the fix; now works for me!

PR #3958, opened 2022-05-13
fix(ui): debounce the scroll event to make chrome happy

### LINKED ISSUE #3893: "find in page" misbehaves with Chrome/Mac
From Amanda Baber:
> This is happening in Chrome with Mac versions 11 and 12. (Safari is fine.)
> 
> When I use command-F to find text at https://datatracker.ietf.org/iesg/agenda/ or https://datatracker.ietf.org/wg/, Chrome tells me how many occurrences there are, and if I scroll down the page, they’re highlighted. However, it won’t jump to them. The “previous” and “next” up/down arrows don’t work.

larseggert:
Fixes #3893, #3841


ietf/static/js/ietf.js
```diff
@@ -244,15 +246,17 @@ $(function () {
...
+                // otherwise the main view doesn't scroll?
+                .on("scroll", debounce(function () {
                     const item = $('#righthand-nav')
                         .find(".active")
                         .last();
                     if (item.length) {
                         item[0].scrollIntoView({ block: "center" });
                     }
-                });
+                }, 20));
```
NGPixel, line 259: I would suggest using a value like `100` ms debounce if it achieves the same result. I've found using values under 100 ms to be unreliable in many scenarios.

package.json
```diff
@@ -18,6 +18,7 @@
     "jquery-ui-dist": "1.13.0",
     "js-cookie": "3.0.1",
     "list.js": "2.3.1",
+    "lodash.debounce": "4.0.8",
```
NGPixel, line 21: As suggested above, replace with the full `lodash` library instead.

NGPixel (changes requested):

NGPixel (approved):

larseggert:
The failing check should be fixed by #3975

PR #3975, opened 2022-05-16
test: fix more validation issues

larseggert:
Including the ones @rjsparks sent me in email


larseggert:
Not sure why the tests fail, they succeed locally

ietf/nomcom/views.py
```diff
@@ -57,7 +57,7 @@ def index(request):
         else:
             nomcom.url = None
         if year >= 2002:
-            nomcom.ann_url = "/nomcom/ann/#%4d" % year
+            nomcom.ann_url = "/nomcom/ann/#nomcom-%4d" % year
```
rjsparks, line 60: We'll go with this for now, but consider dropping the hyphen since nomcom2021 is a group.

rjsparks (approved):

ietf/static/css/liaisons.css
```diff
@@ -1,67 +1,4 @@
...
-.baseform #id_organization,
-.baseform #id_to_poc,
-.baseform #id_response_contacts,
-.baseform #id_technical_contact,
-.baseform #id_cc1,
-.attach_titleField input,
-.baseform textarea {
-    width: 80%;
-}
+/* TODO: this should use bs5 styling and ideally be integrated into the code */
```
rjsparks, line 1: What tooling are you using that shows the removed styling was not used?

larseggert: I grepped for "baseform" in the repo :-)

PR #3977, opened 2022-05-16
fix(ui): import moment.js library with moment-timezone support in wrapper

jennifer-richards:
Packaging changes broke the mechanism used to make moment.tz available to javascript code. This eliminates the separate moment.js and moment-timezone-with-data-10-year-range.js wrappers and instead imports the timezone-augmented version directly. This provides both the timezone support and the regular moment features.


NGPixel (approved):

jennifer-richards:
The failing checks are due to validation errors that I believe are addressed by #3975

larseggert (approved):

PR #3978, opened 2022-05-16
chore(dev): fix chromedriver crash during tests in devcontainer

NGPixel:
Fixes chromedriver crashing during tests by setting `1777` permissions to `/dev/shm`


.devcontainer/devcontainer.json
```diff
@@ -74,9 +74,9 @@
 		"mrmlnc.vscode-duplicate",
 		"ms-azuretools.vscode-docker",
 		"ms-python.python",
-		"ms-python.vscode-pylance",
+	    	"ms-python.vscode-pylance",
```
rjsparks, line 77: Is this intended?

.vscode/extensions.json
```diff
@@ -1,6 +1,3 @@
 {
-  "recommendations": [
-    "arcanis.vscode-zipfs",
-    "dbaeumer.vscode-eslint"
-  ]
+  "recommendations": []
```
rjsparks, line 2: This also seems unrelated to this particular problem?

rjsparks (changes requested):

.vscode/extensions.json
NGPixel: It's unrelated but it's to silence some popups when opening vscode. I can make a separate PR if you prefer, but it's quite a tiny change with no impact.

PR #3979, opened 2022-05-16
musing: Don't show a scrollbar for the righthand nav

larseggert:
This means that the scrollbars for the nav and the main content will no longer
overlap. But it also means the nav can't be directly scrolled anymore (just via
swiping, etc.)

I'm not sure this is a good tradeoff, but I think someone complained about the
overlap. FWIW, there doesn't seem to be a CSS way to move the scrollbars of the
nav to the left.

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -176,7 +176,16 @@ table tbody.meta {
 #righthand-nav {
     height: 70vh;
     width: inherit;
+
+    -ms-overflow-style: none; // Hide scrollbar for IE and Edge and Firefox
+    scrollbar-width: none; // Hide scrollbar for Firefox
 }
+
+// Hide scrollbar for Chrome, Safari and Opera
+#righthand-nav::-webkit-scrollbar {
+    display: none;
+}
+
 
 // Add some padding when there are multiple buttons in a line than can wrap
 .buttonlist .btn {


jennifer-richards:
I noted the overlap. Not quite a complaint but it's probably dubious ui as it is. I'll look at this when I'm online tomorrow, but I think a hidden scroll bar sounds preferable to a dysfunctional one.

For the css, I was imagining there must be a way to hide the container that gets the scrollbar in a slightly wider container with the same background. The menu would then look like it reached the right margin but not actually do so. I don't know whether this would work out in practice and am even less sure it'll be a common enough issue to warrant a lot of effort to find out.

larseggert:
Agreed.

There are some CSS tricks that people claim will move the scrollbar to the left (playing with ltr/rtl settings and scaling the frame by 180 degrees twice), but I couldn't make the work in FF.)

rjsparks:
Is this intended for merge, or should it be marked draft?

larseggert:
I don't think it should be merged based on the feedback, but I wasn't sure if draft "musing" PRs would even get discussed.

larseggert:
> For the css, I was imagining there must be a way to hide the container that gets the scrollbar in a slightly wider container with the same background. The menu would then look like it reached the right margin but not actually do so. I don't know whether this would work out in practice and am even less sure it'll be a common enough issue to warrant a lot of effort to find out.

I opened #4112 for a future enhancement along those lines; will close this one.

PR #3980, opened 2022-05-16
fix(ui): prevent righthand-nav from overlapping the footer on short screens

larseggert:


rjsparks (approved):

ladygem (commented):
larseggert:fix-footer-overlap

ladygem (commented):
larseggert:fix-footer-overlap

PR #3993, opened 2022-05-20
fix: improve person selection when receiving IANA email

rjsparks:


rjsparks:
Misconfiguration (in the form of two identical Alias records pointing to the same Person) led to this fail:
```
Remote Server returned '554 5.3.0 < #5.3.0 x-unix; Traceback (most recent call last): File "<elided>/ietf/bin/iana-review-email", line 25, in <module> doc_name, review_time, by, comment = parse_review_email(msg) File "<elided>/web/ietf/sync/iana.py", line 254, in parse_review_email by = Person.objects.get(alias__name=name, role__group__acronym="iana") File "<elided>/web/env/lib64/python3.9/site-packages/django/db/models/manager.py", line 82, in manager_method return getattr(self.get_queryset(), name)(*args, **kwargs) File "<elided>/web/env/lib64/python3.9/site-packages/django/db/models/query.py", line 410, in get raise self.model.MultipleObjectsReturned( ietf.person.models.MultipleObjectsReturned: get() returned more than one Person -- it returned 2!>'
```
The data has been repaired. This fix would have allowed the code to succeed. There are corners the code does _not_ handle, but they are obscure enough to continue to risk. It would produce unexpected results if we gave two people with the same Alias roles in the IANA group. If that ever arises, the code or data would need adjusting.

jennifer-richards (approved):

UbaisAli:
Ubi

PR #4014, opened 2022-05-24
ci: Add watchman to image

larseggert:
This speeds up repeated runs of `parcel`, but unfortunately not the initial one.


NGPixel (approved):

NGPixel:
@larseggert I had to remove watchman as the parcel build was failing with `Error: invalid clockspec`

https://github.com/parcel-bundler/parcel/issues/7871

larseggert:
Weird, it ran fine in my image yesterday, but not anymore.

Without `watchman`, `parcel` in the container is unusable slow though.

PR #4017, opened 2022-05-24
fix: show session name on meeting/upcoming. Fixes #4010.

### LINKED ISSUE #4010: meeting/upcoming should show Session.name when it is not empty
Right now, meeting/upcoming only shows the session's tokenized _meeting_ name (e.g. `interim-2022-tools-07`. If the session object's `name` attribute is not empty, that should be shown instead.

(discovered during a conversation with @cindymorgan)

rjsparks:


ietf/templates/meeting/upcoming.ics
```diff
@@ -4,7 +4,7 @@ METHOD:PUBLISH
 PRODID:-//IETF//datatracker.ietf.org ical upcoming//EN
 {{vtimezones}}{% for item in assignments %}BEGIN:VEVENT
 UID:ietf-{{item.session.meeting.number}}-{{item.timeslot.pk}}
-SUMMARY:{% if item.session.name %}{{item.session.name|ics_esc}}{% else %}{{item.session.group.acronym|lower}} - {{item.session.group.name}}{%endif%}
+SUMMARY:{% if item.session.name %}{{item.session.group.acronym|lower}} - {{item.session.name|ics_esc}}){% else %}{{item.session.group.acronym|lower}} - {{item.session.group.name}}{%endif%}
```
jennifer-richards, line 7: There seems to be a stray `)` in this line. On my test, where I gave a session the name "Named Session," I get:
```
[...]
BEGIN:VEVENT
UID:ietf-interim-2022-cellar-09-14114
SUMMARY:cellar - Named Session)
STATUS:CONFIRMED
[...]
```

jennifer-richards (changes requested):
Seems to be stray punctuation in the ics template.

ietf/templates/meeting/upcoming.ics
rjsparks: good catch

jennifer-richards (approved):

PR #4018, opened 2022-05-25
chore: make /assets a named volume for dev container

### LINKED ISSUE #4012: Provide `rsync`'ed assets via a container image
### Description

I recently started `rsync`ing the extra assets via `docker/scripts/app-rsync-extras.sh`, which places a lot of files into `data/developers`. That caused `parcel` to run extremely slowly inside the container, due to https://github.com/parcel-bundler/parcel/issues/8128.

Would it be possible to provide those extra assets via a CI-built container image that we hook into the `docker-compose` setup? That way, they wouldn't need to be under `/workspace` and hence would not slow `parcel` down.

An alternative would be to go somehow re-enable running `yarn` and `parcel` on the host rather than in the container, which is what my original Dockerfile did, but I think @NGPixel found some downsides with that?

### Code of Conduct

NGPixel:
This PR adds a named volume "app-assets" mapped to `/assets` inside the dev container.

Configs and scripts were modified to point to this new location.

fixes #4012

### DIFF
--- a/.devcontainer/docker-compose.extend.yml
+++ b/.devcontainer/docker-compose.extend.yml
@@ -8,6 +8,7 @@ services:
             - .:/workspace
             - /workspace/.parcel-cache
             - /workspace/__pycache__
+            - app-assets:/assets
             - datatracker-vscode-ext:/root/.vscode-server/extensions
         # Runs app on the same network as the database container, allows "forwardPorts" in devcontainer.json function.
         network_mode: service:db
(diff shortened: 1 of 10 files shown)


larseggert:
Can we remove the `data` directory entirely?

larseggert (commented):
Can we remove the `data` directory entirely?

Also, check where else `data/developers` is used, I think at least `docker/scripts/app-create-dirs.sh` also needs to change.

NGPixel:
`app-create-dirs.sh` fixed. Search didn't return any other files still using `data/developers`.

rjsparks (approved):

PR #4021, opened 2022-05-25
chore: Update deps

larseggert:


rjsparks:
fwiw - I am increasingly discomfortable with how little test we have over the js using these dependencies.

NGPixel (approved):

PR #4027, opened 2022-05-26
chore: add .dockerignore

NGPixel:
Add dockerignore to speed up the docker build process by excluding the quite large `.git` directory.


jennifer-richards (approved):

larseggert:
Should we add python and other caches that might exist on the host?

PR #4030, opened 2022-05-30
refactor: replace flot with highcharts

larseggert:
Since flot hasn't been updated since 2014 and was only used in one
place. Simplify how highcharts is initialized and used, and re-enable
pre-bs5 export functionality.


ietf/templates/stats/review_stats.html
```diff
@@ -333,9 +332,41 @@ <h2>
...
-            <script src="{% static 'ietf/js/flot.js' %}">
-            </script>
-            <script src="{% static 'ietf/js/review-stats.js' %}">
-            </script>
+        <script src="{% static 'ietf/js/highcharts.js' %}"></script>
+        <script>
+            $(document)
+                .ready(function () {
+                    if (window.timeSeriesData) {
+                        console.log(window.timeSeriesData);
```
NGPixel, line 340: console.log leftover

NGPixel (changes requested):

NGPixel (approved):

PR #4037, opened 2022-05-31
feat: Celery support and asynchronous draft submission API

### LINKED ISSUE #3286: Add replaces to submissions API
`type_enhancement`   |    by mt@lowentropy.net

___

The automated API at https://datatracker.ietf.org/api/submit allows for the submission of drafts, but the information that this accepts is limited.

It would be good if this could be used to include "replaces" information.  I would use this for individual drafts that are renamed by their authors and for drafts that are adopted by working groups.

Right now, we rely on chairs and ADs to provide this information, but it could be automated.  That would both save time and be more reliable.

This is not a high priority item, but it would be nice to have.

___
_Issue migrated from trac:3286 at 2022-03-04 08:40:52 +0000_

### LINKED ISSUE #3791: gateway timeout results in half-posted document
### What happened?

I submit an ID through the API, but it takes too long, I guess, and I get an error from cloudflare.
So I try again, and it seems that the submission has already taken place.

a) maybe the error should include the hostname as part of the URL?  Yes, it's relative, but the error is an API return...
b) maybe it's okay if I submit a second time, and the input file is identical, I get the same reply, not an error?
c) maybe, if it's going to take more than a few s, that it should go into a job queue, and I should get a reference to a job activity that I can query for status.

I know how to deal with this problem, but it may be very confusing to others.

```
curl -S -F "user=mcr+ietf@sandelman.ca" -F "xml=@constrained-voucher.xml" https://datatracker.ietf.org/api/submit
<html>
<head><title>504 Gateway Time-out</title><script async src='/cdn-cgi/bm/cv/669835187/api.js'></script></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx/1.16.1</center>
<script type="text/javascript">(function(){window['__CF$cv$params']={r:'6f8391f5ca7aa228',m:'hcH7aS7BmeAwZupk.T.s.agLaCN2ok3q5PAjruxmcRo-1649343030-0-Ade8JFZjpQuHgOF2ZIwhmQSTZTxw+WQj1rPDXZ7vVyeF1SO5YSoL8qBIdlkrsq+6e5nHcKL59AiIHZmR1MG8gWb7wng3D48ubYutQbl28Gms6x66hksy1ET8IdUfT3hwQQ==',s:[0x95d6cbcc72,0xc08ac8c8f1],}})();</script></body>
</html>
[make submit]: u:5.37s s:0.46s r:67.44s 8% T:0 D:0 T:0 M:55
0sw 10+126667pf 8328i 1000o 0r 0s 0sig 553waits 278csw
obiwan-[work/anima/constrained-voucher](2.6.6) mcr 10087 %make submit
curl -S -F "user=mcr+ietf@sandelman.ca" -F "xml=@constrained-voucher.xml" https://datatracker.ietf.org/api/submit
Validation Error: {'__all__': ['A submission with same name and revision is currently being processed. <a href="/submit/status/125597/">Check the status here.</a>']}%                     
```

Not Applicable

### Code of Conduct

jennifer-richards:
This rather large pull request adds support for Celery asynchronous task handling and uses this to implement an asynchronous draft submission API.

The new API has two endpoints. Most of it is in `api_submission()`, which is a refactored and slightly renamed `api_submit()`. The other piece is `api_submission_status()`, which provides a way to check whether a submission is still in the new `validating` state.

For development, docker containers for rabbitmq and the celery worker are started up alongside the app and db containers. The celery container is currently pulled from the github package repo under the painless-security fork, to be changed when this is merged. That docker image is based on the existing test image.

Fixes #3791. Fixes #3286.


ietf/submit/views.py
```diff
@@ -923,4 +1032,9 @@ def get_submission_or_404(submission_id, access_token=None):
     if access_token and not key_matched:
         raise Http404
 
-    return submission
\ No newline at end of file
+    return submission
+
+
+def async_poke_test(request):
```
martinthomson, line 1038: Vestigial?

jennifer-richards: Nearly, but plan to hang on to it as a test point for when we deploy the async job handling to the production system.

jennifer-richards:
I've added instructions for deploying updates that should keep the container up to date. I've also added support for the "replaces" field, which resolves #3286.

rjsparks (approved):

PR #4038, opened 2022-05-31
fix: skip assets volume in cleandb script + add confirm prompt to cleanall

NGPixel:
- Change the `cleandb` script to only remove the DB volume instead of all volumes.
- Add a confirm prompt to the `cleanall` script.

### DIFF
--- a/docker/cleanall
+++ b/docker/cleanall
@@ -1,9 +1,14 @@
 #!/bin/bash
 
-cd ..
-echo "Shutting down any instance still running and purge images..."
-docker compose down -v --rmi all
-echo "Purging dangling images..."
-docker image prune
-cd docker
-echo "Done!"
+read -p "Stop and remove all containers, volumes and images for this project? [y/N] " -n 1 -r
+echo 
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    cd ..
+    echo "Shutting down any instance still running and purge images..."
+    docker compose down -v --rmi all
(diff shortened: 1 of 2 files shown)


docker/cleanall
```diff
@@ -1,9 +1,14 @@
 #!/bin/bash
 
-cd ..
-echo "Shutting down any instance still running and purge images..."
-docker compose down -v --rmi all
-echo "Purging dangling images..."
-docker image prune
-cd docker
-echo "Done!"
+read -p "Stop and remove all containers, volumes and images for this project? [y/N] " -n 1 -r
```
rjsparks, line 3: I think this will prune dangling images from my other projects too won't it?

rjsparks: Not sure it's pruning volumes?

NGPixel: - Yes that would also prune dangling images from elsewhere as well, but since they are images not used by any container, is that a concern?

- The `-v` flag removes both anonymous volumes attached to the container and named volumes specified in the docker-compose, in this case the `mariadb-data` and `app-assets` volumes.

rjsparks: I have been bitten by having images that are expensive to build being removed by such a global prune. The string you have above claims the scope of the action will be this project. They will be upset if it does something more than it says it does. If you can't scope the prune to this project, the warning question text needs to change.

rjsparks (approved):

PR #4066, opened 2022-06-08
fix: Fix some HTML validation errors that came in via #4035

larseggert:
Also make validation errors display a bit better.

### DIFF
--- a/ietf/templates/base/icons.html
+++ b/ietf/templates/base/icons.html
@@ -12,7 +12,7 @@
 <link rel="manifest" href="/site.webmanifest">
 <link rel="mask-icon"
       href="{% static 'ietf/images/ietf-logo-nor-mask.svg' %}"
-      content="{% if server_mode and server_mode == 'production' %}#ffffff{% else %}#ffc4c4{% endif %}">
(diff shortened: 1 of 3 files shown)


ietf/templates/base/icons.html
```diff
@@ -12,8 +12,8 @@
 <link rel="manifest" href="/site.webmanifest">
 <link rel="mask-icon"
       href="{% static 'ietf/images/ietf-logo-nor-mask.svg' %}"
-      {% if server_mode and server_mode == 'production' %}color="#ffffff"{% else %}color="#ffc4c4"{% endif %}>
+      content="{% if server_mode and server_mode == 'production' %}#ffffff{% else %}#ffc4c4{% endif %}">
```
jennifer-richards, line 15: I think the `content` should be `color` for the `link rel="mask-icon"` ([Apple developer docs for pinned tabs](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html)).

jennifer-richards (changes requested):

ietf/templates/base/icons.html
larseggert: Good catch, thanks

jennifer-richards (approved):

PR #4067, opened 2022-06-08
feat: Replace graphviz with d3

larseggert:
This eliminates the datatracker's dependency on graphviz, by using d3 to draw the document dependency graph.


larseggert:
Need to figure out why @rjsparks sees different data in the graphs, and why Chrome doesn't like my SVG.

larseggert:
@rjsparks maybe gives this another look? Made a bunch of changes.

rjsparks:
I'm not getting any hover effects on the graph or the legend, even after a complete rebuild. I'll ask Nick to look.

NGPixel:
I'm not getting any hover effects on the bubbles either.

rjsparks (approved):

rjsparks:
This doesn't look right, but I think we're close enough to bring it in, and can address this (and some protonopia issues I have with the colors) later.

PR #4069, opened 2022-06-08
fix: Make attention indicators less subtle again

### LINKED ISSUE #4009: Document needs attention indicators have become too subtle.
The IESG would like to have the datatracker be more assertive in pointing out that a document has been in a state for a long time. The egg-timer takes less screen space, but doesn't capture enough attention to drive action.

Conversation about this began with the AD dashboard, but it was noted that any search result page should have stronger indication when documents are in such states.

larseggert:
Fixes #4009

### DIFF
--- a/ietf/templates/doc/search/status_columns.html
+++ b/ietf/templates/doc/search/status_columns.html
@@ -77,7 +77,7 @@
             <br>
             Action Holder{{ doc.documentactionholder_set.all|pluralize }}:
             {% for action_holder in doc.documentactionholder_set.all %}
-                {% person_link action_holder.person title=action_holder.role_for_doc %}{% if action_holder|action_holder_badge %} {{ action_holder|action_holder_badge }}{% endif %}{% if not forloop.last %},{% endif %}
+                {% person_link action_holder.person title=action_holder.role_for_doc %}{% if forloop.last %}{% if action_holder|action_holder_badge %} {{ action_holder|action_holder_badge }}{% endif %}{% else %},{% endif %}
             {% endfor %}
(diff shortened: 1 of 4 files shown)


ietf/templates/doc/search/status_columns.html
```diff
@@ -77,7 +77,7 @@
             <br>
             Action Holder{{ doc.documentactionholder_set.all|pluralize }}:
             {% for action_holder in doc.documentactionholder_set.all %}
-                {% person_link action_holder.person title=action_holder.role_for_doc %}{% if action_holder|action_holder_badge %} {{ action_holder|action_holder_badge }}{% endif %}{% if not forloop.last %},{% endif %}
+                {% person_link action_holder.person title=action_holder.role_for_doc %}{% if forloop.last %}{% if action_holder|action_holder_badge %} {{ action_holder|action_holder_badge }}{% endif %}{% else %},{% endif %}
```
rjsparks, line 80: This doesn't work (as the test failure at ietf.group.tests_info.GroupPagesTests.test_group_documents points out).

It is possible that not all the action holders have the same start time.

In this test, the first action holder was the one that had been added a long time ago, the second had been just added, so no badge got displayed.

rjsparks (changes requested):

ietf/templates/doc/search/status_columns.html
rjsparks: (To be fair, my suggestion on the associated ticket implied one badge here might work, but on review of the code, it doesn't)

rjsparks (approved):

PR #4075, opened 2022-06-09
fix(ui): don't let the sort chevron wrap

### LINKED ISSUE #4071: Meeting requests column headings formatted poorly
### Describe the issue

On the [meeting requests page](https://datatracker.ietf.org/meeting/requests), the first few column headings are formatted poorly. The columns are too narrow, taking up two lines with the sort button meant to be to beside the heading text on the next line. This occurs even with a 2500px wide browser until I reduce the font size to something unreadable.

Seeing this on Firefox 101.0 on Linux. Screenshot shows the first few - the AD column is correct, but all those to its left show the problem.

Screenshot from 2022-06-08 13-31-35

### Code of Conduct

larseggert:
I couldn't figure out a way to not let it wrap *and* float it to the right,
however,

Fixes #4071

### DIFF
--- a/ietf/static/css/list.scss
+++ b/ietf/static/css/list.scss
@@ -1,8 +1,6 @@
 // Import bootstrap helpers
 @import "bootstrap/scss/functions";
 @import "bootstrap/scss/variables";
-// @import "bootstrap/scss/mixins";
-// @import "bootstrap/scss/utilities";
 
 table .sort {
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
Looks good to me. I don't see an easy way to float the icon right short of rewriting the page, and I think it's ok next to the heading.

PR #4076, opened 2022-06-09
fix(ui): badge -> badge rounded-pill

larseggert:
To make badges look a little less like small buttons.

### DIFF
--- a/ietf/templates/base.html
+++ b/ietf/templates/base.html
@@ -97,7 +97,7 @@
                             {{ message.message }}
                         </div>
                     {% endfor %}
-                    {% block content %}{{ content|safe }}{% endblock %}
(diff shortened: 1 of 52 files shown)


ietf/templates/base.html
```diff
@@ -97,7 +97,7 @@
                             {{ message.message }}
                         </div>
                     {% endfor %}
-                    {% block content %}{{ content|safe }}{% endblock %}
+                    {% block content %}{{ content|safe }}{% endblock %}<span class="badge bg-secondary">New</span><span class="badge bg-secondary">Secondary</span>
```
rjsparks, line 100: It's not clear what these spans are supposed to be doing?

rjsparks (approved):

Issue #3653, opened 2022-03-16
Pdfized rendering pagination incorrect on first page break


jimfenton:
### What happened?

When rendering an Internet Draft (specifically, draft-ietf-jmap-blob-09) using the *pdfized* format on the draft's information page, the first page includes, at the bottom, the heading from page 2. The rest of the pages are paginated correctly.

### What browser(s) are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow the IETF's Code of Conduct

larseggert:
This should be fixed as part of #4082. Note that it also needs https://github.com/ietf-tools/rfc2html/pull/12.

larseggert:
@jimfenton can you confirm if this is fixed now?

jimfenton:
@larseggert Has the pdfized format option gone away? I don't see it on the draft information page any more. IIRC, it was basically a rendering of the .txt file into a PDF (with page numbering, headers, and footers). If that option was eliminated for some other reason, then this issue has been overcome by events.

The pdf format looks quite good although without headers, footers, and page numbers that the .txt format has. I looked at draft-ietf-jmap-blob-18, and it looks like maybe the widow/orphan control got out of hand on Section 4.1.2, where there is a page with only the section header followed by a long section of json. But that's probably a separate issue.

larseggert:
> @larseggert Has the pdfized format option gone away? I don't see it on the draft information page any more.

Er, no? See the "pdf" button.

<img width="867" alt="Screenshot 2023-03-06 at 19 50 20" src="https://user-images.githubusercontent.com/200328/223190769-180b7628-2a0e-49e2-8805-fa5ccb678649.png">

jimfenton:
I thought there was a separate button, I think called "pdfized", that rendered a PDF that looks like the txt format (with page numbering, headers, and footers). I tried to start my old development version of datatracker to see if it was there, but it doesn't work (I changed laptops in the interim, among other things). But if pdfized went away, or the behavior of the pdf version changed to eliminate the page numbering, etc. then it's OK to call this fixed as OBE.

rjsparks:
I wonder if line 17 below:
https://github.com/ietf-tools/datatracker/blob/996e6c2011d8537cdfd2ab5a00599946126f5b87/ietf/templates/doc/document_format_buttons.html#L16-L17
should be {{ label }}, similar to line 19?

jimfenton:
[...]

PR #4082, opened 2022-06-14
feat: Use bs5 for htmlized doc

### LINKED ISSUE #4057: Unexpected type size behavior in htmlized drafts
### Describe the issue

I am a simple routing guy and don't know too much about no htmls and stylesheets and whatnot, but the type size behavior of the htmlized drafts is unexpected -- if I make my window super wide, the type gets real small. As I make it narrower, the type jumps several sizes bigger, to roughly fit the window's width. Narrower still, horizontal scroll appears and the type stays the same. A little narrower than that, and the type drops down to medium-ish. Still narrower and the type gets freaky small again.

(Changing the type size in the very-large-window rendering with my browser's (Safari 15.5) "make text bigger/smaller" options produces unexpected results also (some sizes are skipped over) but it's not clear to me if this is a deficiency in the browser or what.)

Most of the behavior seems to be trying to fit the line width to the window width... except when the window is super wide in which case the counter-intuitive "make the type teensy" choice is used.

I tend to use wide windows because reasons, so this weirded me out until I figured out what was happening.

### Code of Conduct

larseggert:
This gets rid of a bunch of old code by re-using `document_main` to render htmlized docs into a new set of bs5-based templates.

Also fixes #4057 and #3653.

This depends on https://github.com/ietf-tools/rfc2html/pull/11 and https://github.com/ietf-tools/rfc2html/pull/12.


rjsparks:
This is a pretty big change of what's really the most heavily fetched format from the datatracker. We may want to phase this in? (and possible fix the html brokenness in rfc2html at the same time).

As mentioned out of band, when I first moved this to the datatracker, I didn't preserve the colored header and got immediate feedback that people wanted it back.

Please look at #3466.

larseggert:
I'm adding the colored header - as a colored badge.

larseggert:
> We may want to phase this in?

We should definitely deploy on sandbox first and ask for feedback.

larseggert:
@rjsparks should we chat about how to make this available for community feedback?

rjsparks:
We can do a deployment of the branch on the sandbox server using a new DNS name (such as htmlize-preview.ietf.org), or you could look at rewiring the PR to put this htmlization at a different endpoint than the current, so that they are both available in production for some time (/doc/htmlpreview/ perhaps). If we go with deploying the branch, it would be good to keep it close to main.

larseggert:
I'll look I to adding another button for deployment testing via `main`.

larseggert:
@rjsparks I think this is ready for you to take a look (htmlization and pdfization of varius documents). If you think it's ready, maybe we do public testing?

ietf/static/css/document_html_txt.scss
```diff
@@ -0,0 +1,299 @@
...
+  top: 0;
+  right: 0;
+  margin: 0;
+}
+#identifiers dd.updates {
+  margin-left: 9ch;
+}
+#identifiers dd.obsoletes {
+  margin-left: 11ch;
+}
```
martinthomson, line 99: Replace this bit...

```suggestion
#identifiers dt:is(.label-stream, .label-rfc, .label-std, .label-bcp, .label-internet-draft,
                   .label-workgroup, .label-intended-status, .label-obsoletes, .label-updates,
                   .label-published, .label-expires, .label-category, .label-issn,
                   .label-authors), .ears {
  display: none;
}
#identifiers {
  margin: 0;
  display: grid;
  grid-template-columns: 47ch 24ch;
  column-gap: 1ch;
  grid-auto-flow: column;
  grid-auto-rows: auto;
  row-gap: 0;
}
#identifiers dt {
  margin: 0 1ch 0 0;
  float: revert;
  display: inline-block;
}
#identifiers dd {
  margin: 0;
  width: 47ch;
}
#identifiers dd::before {
  margin: 0 1ch 0 0;
}
#identifiers dd.rfc::before {
  content: "Request for Comments:";
}
#identifiers dd.std::before {
  content: "STD:";
}
#identifiers dd.bcp::before {
  content: "BCP:";
}
#identifiers dd.internet-draft::before {
  content: "Internet-Draft:";
}
#identifiers dd.workgroup::before {
  content: "Workgroup:";
}
#identifiers dd.intended-status::before {
  content: "Intended Status:";
}
#identifiers dd.obsoletes::before {
  content: "Obsoletes:";
  margin-left: -11ch;
}
#identifiers dd.obsoletes {
  padding-left: 11ch;
  width: 36ch;
}
#identifiers dd.updates::before {
  content: "Updates:";
  margin-left: -9ch;
}
#identifiers dd.updates {
  padding-left: 9ch;
  width: 38ch;
}
#identifiers dd.updates::first-line {
  padding-left: 0;
}
#identifiers dd.published::before {
  content: "Published:";
}
#identifiers dd.expires::before {
  content: "Expires:";
}
#identifiers dd.category::before {
  content: "Category:";
}
#identifiers dd.issn::before {
  content: "ISSN:";
}
#identifiers dd.authors {
  grid-area: 1 / 2 / 100 / 3;
  width: 24ch;
  text-align: right;
}
```

Yes. this is an abomination.  Sorry about that.

ietf/static/css/document_html_txt.scss
```diff
@@ -0,0 +1,348 @@
...
+  content: "Updates:";
+  margin-left: -9ch;
+}
+#identifiers dd.updates {
+  padding-left: 9ch;
+  width: 38ch;
+}
+#identifiers dd.updates::first-line {
+  padding-left: 0;
+}
```
martinthomson, line 131:
```suggestion
#identifiers dd.obsoletes::before {
  content: "Obsoletes:";
  margin: 0 0 0 -10ch;
}
#identifiers dd.obsoletes {
  padding-left: 10ch;
  width: 37ch;
}
#identifiers dd.updates::before {
  content: "Updates:";
  margin: 0 0 0 -8ch;
}
#identifiers dd.updates {
  padding-left: 8ch;
  width: 39ch;
}
#identifiers dd:is(.updates, .obsoletes) {
  /* HAXX: this gets around the lack of text-content-trim support */
  display: flex;
  flex-wrap: wrap;
}
#identifiers dd:is(.updates, .obsoletes) a {
  margin: 0 0 0 1ch;
}
```

The abomination grows even worse.  Now we have flexbox inside grid for maximum awesomeness.  I am working around a "bug" in xml2rfc here with this change.

ietf/doc/views_doc.py
```diff
@@ -138,12 +140,12 @@ def interesting_doc_relations(doc):
 
     return interesting_relations_that, interesting_relations_that_doc
 
-def document_main(request, name, rev=None):
+def document_main(request, name, rev=None, document_html=False):
```
rjsparks, line 143: This need discussion - If we change the document main view to allow choosing what to display to be an option, we should make this generic. It'll make a big difference when we replace document_main with a better front-end.

ietf/doc/views_doc.py
```diff
@@ -444,8 +445,21 @@ def document_main(request, name, rev=None):
         else:
             stream_desc = "(None)"
 
-        return render(request, "doc/document_draft.html",
+        html = doc.html_body()
```
rjsparks, line 448: For this implementation, please guard this with a check on document_html - don't burn the compute if the template isn't going to use the result.

ietf/doc/views_doc.py
larseggert: What do you mean by "make this generic"?

larseggert:
@martinthomson  the recent CSS changes don't work in the Weasyprint PDFization:
Screen Shot 2022-08-31 at 12 50 07

ietf/doc/views_doc.py
rjsparks: I think ultimately a callable, but I'm not sure it should be part of the signature of the function itself. The question is if we want to switch this based on URLs, previously expressed user preferences, or sitewide configuration.

larseggert: Sorry, I'm dense - switch between _what_?

martinthomson:
Ugh, that's clearly a lack of CSS grid support.

I do have to ask why you are feeding this stylesheet in for printing to PDF though.  Isn't the regular style sheet better for that?

larseggert:
Because so far, the htmlized text was perused by the datatracker. Maybe that should change. I think the RFC Ed Offiziers the html?

martinthomson:
Hmm, I can maybe try to use @supports to provide some fallback styling (which won't look like the original).  Or you can do what rfc-editor.org does.  For example: https://www.rfc-editor.org/rfc/rfc9261.pdf

larseggert:
The latter probably makes most sense - @rjsparks any preference here?

rjsparks:
I think you're getting confused between pdf output from xml2rfc v3 and pdf of htmlized documents.
Martin is pointing to v3 pdf output with his link above.

As with most places where we have confusion across the v3 boundary wrt -ized documents, we need to consider whether people will expect the fundamental structure of what they're getting back to be the same for pre-v3 and post-v3 documents. For the endpoints that advertise themselves as -ized, I think they should return the same kind of thing.

rjsparks:
Moving this into a branch at ietf-tools to assist with a feedback deployment

PR #4089, opened 2022-06-16
fix: rsync fewer files

larseggert:
It's still super slow, however.

### DIFF
--- a/docker/scripts/app-rsync-extras.sh
+++ b/docker/scripts/app-rsync-extras.sh
@@ -33,33 +33,66 @@ done
 
 echo "Using destination $DEST_ROOT"
 
+EXCLUDE="$(mktemp)"
+cat << EOF > "$EXCLUDE"
+*#
+*%
+*.diff
+*.html
+*.json
+*.new
+*.p7s
+*.pdf
+*.ps
+*.tar
+*.xml
+*~
+/bcp
+/beta
+/fyi
+/ien
+/inline-errata
+/pending-errata
+/prerelease
+/std
+/v3test
+rfc[0-9]
+rfc[0-9][0-9]
+rfc[0-9][0-9][0-9]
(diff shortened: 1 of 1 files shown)


docker/scripts/app-rsync-extras.sh
```diff
@@ -33,33 +33,66 @@ done
...
 mkdir -p "$dest"
 echo "Fetching $dest ..."
-rsync -auz ${PROGRESS:+--info=progress2} rsync.ietf.org::id-archive/ $dest/
+rsync $OPTS rsync.ietf.org::id-archive/ $dest/
 
 dest="$DEST_ROOT/www6s/proceedings"
 mkdir -p "$dest"
 echo "Fetching $dest ..."
-rsync -auz ${PROGRESS:+--info=progress2} rsync.ietf.org::proceedings/ $dest/
+rsync $OPTS rsync.ietf.org::proceedings/ $dest/
```
rjsparks, line 96: Do we really need to sync all the proceedings?

rjsparks (approved):

docker/scripts/app-rsync-extras.sh
larseggert: I don't think so. Do we need _any_?

rjsparks: yes, it's where the minutes/agendas/slides are currently stored, so getting the most recent and upcoming meetings would likely be useful by default.

(We need to change where these are stored on disk in the long run)

larseggert: So one quick fix would be not not rsync proceedings of IETFs <100 (which is possible using the limited patterns that rsync allows) - should I add that?

rjsparks: Try it and look at the size. I think you could get 11* easily too?

PR #4100, opened 2022-06-20
chore: remove ietf/group/colors.py, which is unused

larseggert:


rjsparks:
that's reasonable - but the problem I'm pointing to (which isn't really the responsibility of this PR, but of changes leading up to it) then exists with how views use the css. I think we're diverging (Nick's agenda pages use the base colors, but darken them in a wash for instance). Have we reconciled group colors between the agenda and agenda editor yet?

rjsparks (approved):

PR #4101, opened 2022-06-20
chore: Remove various unused template tags

larseggert:


rjsparks:
The title of this PR implies it is simply cleaning away unused cruft, but instead, it makes changes to the output of several templates.

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -142,45 +125,11 @@ def bracketpos(pos,posslug):
...
-    added
-    """
-    string = str(string)
-    if string[:3].lower() == "rfc" and string[3] != " ":
-        return string[:3].upper() + " " + string[3:]
-    else:
-        return string
-
-@register.filter(name='rfcnospace')
-def rfcnospace(string):
```
rjsparks, line 159: The ability to have more than one representation (with and without space) was intentional. While we moved most of the datatracker to having the space, there was concern about disrupting the workflow of external entities using the IPR system. If changing the output wasn't what you meant to do, there's some further suggestion below.

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -142,45 +125,11 @@ def bracketpos(pos,posslug):
...
-    """
-    string = str(string)
-    if string[:3].lower() == "rfc" and string[3] == " ":
-        return string[:3] + string[4:]
-    else:
-        return string
-
 @register.filter
 def prettystdname(string):
     from ietf.doc.utils import prettify_std_name
```
rjsparks, line 130: This function takes an argument named `spacing`, and the filter could be modified to take the argument as well. In the places this filter is used to replace `rfcnospace`, the filter could then be called with `prettystdname:""`

ietf/templates/api/index.html
```diff
@@ -284,7 +284,7 @@ <h2 id="signing-keys">
         <p>
             When sending notifications to other APIs, the datatracker may sign
             information with a
-            <a href="https://datatracker.ietf.org/doc/html/rfc7515">
+            <a href="/doc/html/rfc7515">
```
rjsparks, line 287: All of these should instead be changed to use the url resolver, e.g. `<a href="{% url 'ietf.doc.views_doc.document_html' name='rfc7515' %}>`

rjsparks (changes requested):

ietf/templates/api/index.html
larseggert: There are many more that don't use the URL resolver, which is why I made the change in this way - want me to change those as well?

rjsparks: Yes, we should move away from having the templates assume they know what the URL structure looks like.

ietf/templates/liaisons/field_help.html
```diff
@@ -34,7 +34,7 @@ <h1>Liaison statement field help</h1>
     <p>
         For definitive information on generating liaison statements, please
         see
-        <a href="{{ "4053"|rfcurl }}">RFC 4053 (BCP 103) "Procedures for Handling Liaison Statements to and from the IETF."</a>
+        <a href="/doc/html/rfc4053">RFC 4053 (BCP 103) "Procedures for Handling Liaison Statements to and from the IETF."</a>
```
rjsparks, line 37: still a bunch of these in the PR?

rjsparks (changes requested):

rjsparks (approved):

PR #4114, opened 2022-06-22
fix: Make the IPR search form work when enter is pressed

larseggert:


ietf/static/js/ipr-search.js
```diff
@@ -1,8 +1,8 @@
 $(document)
     .ready(function () {
         // hack the "All States" check box
-        $("#id_state")
-            .addClass("list-inline");
+        $("#id_state .form-check")
+            .addClass("form-check-inline");
```
jennifer-richards, line 5: Can this not be done in the template?

jennifer-richards (approved):
Looks good to me. One question about the state input hack, but the answer may be "no" and it's not worth holding up over.

ietf/static/js/ipr-search.js
larseggert: These forms would need some serious work to better work with bs.

PR #4115, opened 2022-06-22
fix: Add popup to explain colors for liaisons needing action. Rework how aria-label is supplied for several input fields. Update vnu.jar. Adjust some field styling.

### LINKED ISSUE #4078: "Action Needed" liaison color is not labelled or explained
### Describe the issue

On this page: https://datatracker.ietf.org/liaison/posted/?tags=required

The "Action needed" is either yellow or blue, but this color isn't explained. Can we have a tool tip? Or change the text to explain what each one means?

### Code of Conduct

larseggert:
Also fix the search field styling while I'm here.

Fixes #4078.

### DIFF
--- a/ietf/secr/templates/groups/people.html
+++ b/ietf/secr/templates/groups/people.html
@@ -1,5 +1,5 @@
 {% extends "base_site.html" %}
-{% load staticfiles %}
+{% load staticfiles widget_tweaks %}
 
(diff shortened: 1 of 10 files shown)


rjsparks:
I can't quickly find what this PR claims it is doing - where is the popup?
There's so much _else_ here - why isn't it in another PR?

rjsparks (approved):

PR #4124, opened 2022-06-24
fix: display the agenda warning note more prominently

### LINKED ISSUE #4123: Meeting agenda_warning_note not shown prominently
### Describe the issue

The meeting agenda is meant to show a prominent warning when the `agenda_warning_note` field is set, e.g., to "PRELIMINARY" while the agenda is under development. In the main HTML agenda template, it is currently shown only in the heading for the `#weekview` section. As a result, it is entirely hidden unless the user is applying filters to their view.

Need to show it prominently whether or not the week view is open.

### Code of Conduct

jennifer-richards:
Fixes #4123

### DIFF
--- a/ietf/templates/meeting/agenda.html
+++ b/ietf/templates/meeting/agenda.html
@@ -102,6 +102,11 @@
     <h2 class="mt-3">
         {% if personalize %}Personalize{% endif %}
         Detailed Agenda
+        {% if schedule.meeting.agenda_warning_note %}
+            <span class="badge rounded-pill bg-danger">
+              {{ schedule.meeting.agenda_warning_note|removetags:"h1" |safe }}
+            </span>
+        {% endif %}
     </h2>
     {% if personalize %}
         <p>
--- a/ietf/templates/meeting/meeting_heading.html
+++ b/ietf/templates/meeting/meeting_heading.html
@@ -3,11 +3,16 @@
 {% origin %}
 {# assumes meeting is in context #}
 {% load origin %}
-{% load ietf_filters %}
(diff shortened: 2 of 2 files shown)


ietf/templates/meeting/agenda.html
```diff
@@ -102,6 +102,11 @@ <h2>
     <h2 class="mt-3">
         {% if personalize %}Personalize{% endif %}
         Detailed Agenda
+        {% if schedule.meeting.agenda_warning_note %}
+            <span class="badge rounded-pill bg-danger">
+              {{ schedule.meeting.agenda_warning_note|removetags:"h1" |safe }}
```
rjsparks, line 107: Is this "no h1 but everything else is safe" dance reintroduction of legacy code?
I wonder if the right thing to do, instead, is to make agenda_warning_note be either plain text (never marked safe) or be made markdown (migrating any old content, though such probably shouldn't exist at all).

jennifer-richards: It's copy-pasted verbatim from the instance of the flag that was left on the week-view header. This note really needs to be plaintext - it's a badge on a page heading, so more than a word or two is a recipe for yuck.

jennifer-richards: Just checked, and aside from my own debugging there are no non-empty `agenda_warning_notes` in the dev DB.

rjsparks: per further discussion, we'll turn this into a boolean in the long run, but we can work towards bringing the fix in this PR in for now.

rjsparks (approved):

PR #4140, opened 2022-06-29
feat: isolate py27 code interfacing with mailman from py39 code interfacing with django.

rjsparks:


ietf/mailinglists/management/commands/import_mailman_listinfo.py
```diff
@@ -1,25 +1,19 @@
...
-try:
-    from Mailman import Utils
-    from Mailman import MailList
-    from Mailman import MemberAdaptor
-    have_mailman = True
-except ImportError:
-    pass
 
 from ietf.mailinglists.models import List, Subscribed
 from ietf.utils.log import log
```
jennifer-richards, line 19: I think the `import decode` on the next line can go away

jennifer-richards (approved):
This looks ok to me. A couple minor suggestions inline.

PR #4145, opened 2022-07-01
fix: dont serialize Submission.first_two_pages through /api/v1.

### LINKED ISSUE #4083: v1 api crashes on some content when serializing to xml
See the very old ticket at https://github.com/django-tastypie/django-tastypie/issues/1107.

submission.first_two_pages can contain formfeeds. These break tastypie's xml serialization. Json serialization succeeds.

The quick fix is to stop exposing first_two_pages through the API.

rjsparks:
Fixes #4083

### DIFF
--- a/ietf/submit/resources.py
+++ b/ietf/submit/resources.py
@@ -68,6 +68,7 @@ class SubmissionResource(ModelResource):
             "group": ALL_WITH_RELATIONS,
             "draft": ALL_WITH_RELATIONS,
         }
+        excludes = ('first_two_pages',)
 api.submit.register(SubmissionResource())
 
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
Looks good. My only concern was whether the `makeresources` management command will interact with this. I think it will ignore models that already have resources even if those differ from what it would generate, but thought I'd raise this just in case my understanding is incorrect.

rjsparks:
```dev ➜ /workspace (no2pages@api ✗) $ ietf/manage.py makeresources submit

Inspecting ietf.submit ..
  nothing to do for ietf.submit`

PR #4156, opened 2022-07-04
fix: Better agenda display on phone screens

larseggert:


larseggert:
@rjsparks I figure out how to retain the existing columns

rjsparks (approved):

PR #4161, opened 2022-07-05
fix: Avoid crashes in `urlize_ietf_docs`

### LINKED ISSUE #4138: urlize_ietf_docs breaks on input that contains draft-foo-bar-2016
Some history and shepherd writeup pages are 500-ing as content is being fed through urlize_ietf_docs which incorrectly dissects a draft string and attempts a reverse lookup that matches no URL patterns.

The issue is in the regular expression for the link_non_charter_doc_match substitution. The below shows the root of the malbehavior:

```
>>> a
'\'The following Last Call announcement was sent out:&lt;br&gt;&lt;br&gt;From: The IESG &lt;iesg-secretary@ietf.org&gt;\nTo: "IETF-Announce" &lt;ietf-announce@ietf.org&gt;\nCC: alexey.melnikov@isode.com, draft-pal-eidr-urn-2016@ietf.org\nReply-To: ietf@ietf.org\nSender: &lt;iesg-secretary@ietf.org&gt;\nSubject: Last Call: &lt;draft-pal-eidr-urn-2016-01.txt&gt; (Entertainment Identifier Registry (EIDR) URN Namespace Definition) to Informational RFC\n\n\nThe IESG has received a request from an individual submitter to consider\nthe following document:\n- \'Entertainment Identifier Registry (EIDR) URN Namespace Definition\'\n  &lt;draft-pal-eidr-urn-2016-01.txt&gt; as Informational RFC\n\nThe IESG plans to make a decision in the next few weeks, and solicits\nfinal comments on this action. Please send substantive comments to the\nietf@ietf.org mailing lists by 2016-07-01. Exceptionally, comments may be\nsent to iesg@ietf.org instead. In either case, please retain the\nbeginning of the Subject line to allow automated sorting.\n\nAbstract\n\n\n   Entertainment Identifier Registry (EIDR) Identifiers are used for the\n   globally unique identification of motion picture and television\n   content.  This document defines the formal Uniform Resource Name\n   (URN) Namespace Identifier (NID) for EIDR Identifiers.\n\n   This document revises Request For Comments (RFC) 7302 -- see\n   Appendix A for a summary of changes.\n\n\n\n\nThe file can be obtained via\nhttps://datatracker.ietf.org/doc/draft-pal-eidr-urn-2016/\n\nIESG discussion can be tracked via\nhttps://datatracker.ietf.org/doc/draft-pal-eidr-urn-2016/ballot/\n\n\nNo IPR declarations have been submitted directly on this I-D.\n\n\n\''
>>> re.search(r"\b(?<![/\-:=#])((?:draft-|bofreq-|conflict-review-|status-change-)[\d\w\.+-]+(?![-@]))",a)
<re.Match object; span=(200, 222), match='draft-pal-eidr-urn-201'>
```
This ultimately leads to
```
NoReverseMatch at /doc/rfc7972/history/

Reverse for 'ietf.doc.views_doc.document_main' with keyword arguments '{'name': 'draft-pal-eidr-urn', 'rev': '201'}' not found. 1 pattern(s) tried: ['doc/(?P<name>[A-Za-z0-9._+-]+?)(?:/(?P<rev>[0-9]{1,2}(-[0-9]{2})?))?/$']
```

In addition to improving the dissection of the input string, perhaps urlize_ietf_docs should internally try-catch wrap its operations and just return the original input unchanged if attempting to operate on it would cause a crash.

larseggert:
Fixes #4138.

Also refactors the use of `bleach` and adds URLization and better newline handling to Markdown content.

### DIFF
--- a/ietf/templates/doc/document_material.html
+++ b/ietf/templates/doc/document_material.html
@@ -126,7 +126,7 @@
         <div class="card-body">
             {% if doc.rev and content != None %}
                 {% if content_is_html %}
-                    {{ content|sanitize|safe }}
+                    {{ content }}
                 {% else %}
                     <pre>{{ content|maybewordwrap|urlize_ietf_docs|linkify }}</pre>
                 {% endif %}
--- a/ietf/templates/doc/document_material.html
+++ b/ietf/templates/doc/document_material.html
@@ -126,7 +126,7 @@
         <div class="card-body">
             {% if doc.rev and content != None %}
                 {% if content_is_html %}
-                    {{ content }}
+                    {{ content|sanitize|safe }}
                 {% else %}
                     <pre>{{ content|maybewordwrap|urlize_ietf_docs|linkify }}</pre>
                 {% endif %}
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -17,6 +17,7 @@ from django.utils.encoding import force_str # pyflakes:ignore force_str is used 
 from django.urls import reverse as urlreverse
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
+from django.urls import NoReverseMatch
 
 import debug                            # pyflakes:ignore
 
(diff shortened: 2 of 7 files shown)


ietf/templates/doc/document_material.html
```diff
@@ -126,7 +126,7 @@
         <div class="card-body">
             {% if doc.rev and content != None %}
                 {% if content_is_html %}
-                    {{ content|sanitize|safe }}
+                    {{ content }}
```
rjsparks, line 129: I think this is not going to have the effect you intended. If I follow you correctly, you're changing this to count on the content string _not_ being marked safe, and would get the default treatment for an unsafe string. That default will escape all the html tags. So (for instance), https://datatracker.ietf.org/meeting/80/materials/agenda-80-dispatch would not render legibly?
If I'm missing the path that would cause the right thing to happen with such input, let me know.

rjsparks (approved):

PR #4207, opened 2022-07-10
feat: Add opengraph information to document pages

larseggert:
@mnot I think you have some familiarity with opengraph; does this look reasonable at all?


ietf/templates/doc/opengraph.html
```diff
@@ -0,0 +1,16 @@
+{# Copyright The IETF Trust 2016-2020, All Rights Reserved #}
+{% load origin %}
+{% load static %}
+{% origin %}
+<meta property="og:title" content="{{ doc.title }} ({% if doc.get_state_slug == 'rfc' and not snapshot %}RFC {{ rfc_number }}{% else %}{{ name }}-{{ doc.rev }}{% endif %})">
```
mnot, line 5: Including draft name here is going to be very noisy.

```suggestion
<meta property="og:title" content="({% if doc.get_state_slug == 'rfc' and not snapshot %}RFC {{ rfc_number }}: {% endif %}){{ doc.title }}">
```

ietf/templates/doc/opengraph.html
```diff
@@ -0,0 +1,16 @@
...
+<meta property="og:url" content="{{ settings.IDTRACKER_BASE_URL }}{{ request.path }}">
+<meta property="og:image" content="{{ settings.IDTRACKER_BASE_URL }}{% static 'ietf/images/ietf-logo-nor-512.png' %}">
+<meta property="og:image:alt" content="Logo of the IETF">
```
mnot, line 8: What about when it's _not_ an IETF draft? I think it's important to distinguish other streams as well as drafts not on a stream.

ietf/templates/doc/opengraph.html
```diff
@@ -0,0 +1,16 @@
...
+<meta property="og:image:width" content="512">
+<meta property="og:image:height" content="512">
+<meta property="og:type" content="article">
+<meta property="og:site_name" content="IETF Datatracker">
+<meta property="og:article:section" content="IETF - Internet Engineering Task Force">
+<meta property="og:description" content="{% for author in doc.documentauthor_set.all %}{{ author.person.name}}{% if not forloop.last %} and {% endif %}{% endfor %}. {{ doc.title }}. {% if doc.get_state_slug == 'rfc' and not snapshot %}RFC {{ rfc_number }}, {{ published.time|date:'M Y' }}, {{doc.pages}} pages{% else %}{{ name }}-{{ doc.rev }}; work in progress{% endif %}.">
```
mnot, line 14: This is noisy. Social platforms (e.g., [twitter](https://cards-dev.twitter.com/validator)) have metadata previewers that let you see how it will be presented.

Can you take the same approach that [I took](https://github.com/ietf-tools/xml2rfc/pull/777)?

larseggert: Done, please take a look?

mnot:
Looks reasonable to me. Have you tested with the validator?

larseggert:
Yes, but uploading an HTML to eggert.org and then feeding that in. The Twitter validator shows no error but says it can't render. The Facebook one seems happier. If you want to try yourself: https://eggert.org/x.html

mnot:
Hmm. The URL is to 127.0.0.1 - I wonder if they're blocking it because of that?

mnot:
Also, last time I checked the twitter bot can't do TLS 1.3, so you need to allow 1.2.

larseggert:
I don't do TLS 1.3, only 1.2.

mnot:
What do your access logs say?

larseggert:
```
199.59.150.182  - - [13/Jul/2022:19:30:15 +0300] "GET /x.html HTTP/1.1" 200 21456 "-" "Twitterbot/1.0"
```
but still
Screen Shot 2022-07-13 at 9 31 08

larseggert:
Facebook seems to like it though: https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Feggert.org%2Fx.html

rjsparks:
It's not clear to me that this is ready to merge. Are you still looking into why there were failures at twitter?

larseggert:
There is no way to debug this unfortunately. I wonder if it would be safe to merge this and then fix things if they need fixing once everything is in place at the correct URLs. This shouldn't cause any failures, because Twitter would just fall back to the normal cards.

mnot:
I think the twitter validator tool is just broken -- the URLs on my personal blog do the same thing, but if I go to Twitter itself, compose a tweet, and paste in a URL from my blog, the preview shows up successfully.

Likewise, if I do that with Lars' test URL, it works -- although the logo doesn't show correctly.

Screen Shot 2022-07-17 at 9 58 03 am

So I think we just need to fix the logo (1:1 ratio, and maybe smaller?). I see you use a different URL in the PR, so maybe that's already addressed?

Also, I'll take this opportunity to ask again, explicitly: are we **sure** we want to put the IETF logo on drafts that aren't on any stream? That's going to be misleading for the vast majority of people who see it.

larseggert:
I am using the same logo logic now that you are proposing for xml2rfc, so that should avoid IETF logos on I do visual drafts.

I asked Greg to have the design company do a 1:1 icon already, so that should be coming. Or I can do a 1:1 version of the logo in the meantime.

rjsparks (approved):

rjsparks (approved):

PR #4211, opened 2022-07-12
refactor: replace datetime.now and datetime.today with timezone.now

jennifer-richards:
Prepare for switching to `USE_TZ=True` by globally replacing `datetime.datetime.now()` with `django.utils.timezone.now()` through the project. Calls to `datetime.datetime.today()` have also been replaced. Despite the misleading name, this method is a synonym for `datetime.datetime.now()` when the latter is used without an argument.

Adds migrations to update model fields whose default value was set using `datetime.now`. Two old migrations used `datetime.datetime.today()` to find sessions in the future. After initially updating these to use `timezone.now()`, I reverted the change because I think it makes sense to leave these as they were. Migrating back to that point without setting `USE_TZ=False` is going to break timestamps anyway. If we really want to support that we could add a migration step to interrupt migrations and prompt the user to modify `settings.USE_TZ` at the appropriate time. I don't think that complexity is warranted, however.


jennifer-richards:
Converting to draft while I check that it's still compatible with recent updates

ietf/utils/test_runner.py
```diff
@@ -808,8 +808,8 @@ def setup_test_environment(self, **kwargs):
         for offset in range(10):
             try:
                 # remember the value so ietf.utils.mail.send_smtp() will use the same
-                ietf.utils.mail.SMTP_ADDR['port'] = base + offset 
-                self.smtpd_driver = SMTPTestServerDriver((ietf.utils.mail.SMTP_ADDR['ip4'],ietf.utils.mail.SMTP_ADDR['port']),None) 
+                ietf.utils.mail.SMTP_ADDR['port'] = base + offset
```
rjsparks, line 811: well, ok, but it makes reviewing a really big PR even longer...

rjsparks (approved):
I reviewed this fairly quickly - I don't think there's much opportunity for simple errors in what the PR aims for, and we'll have plenty of time to check for subtle as we move to the next phase of this project.

PR #4212, opened 2022-07-12
fix(ui): Make the heading and textarea font sizes a bit smaller

larseggert:


ietf/static/css/ietf.scss
```diff
@@ -73,7 +80,8 @@ html {
 
 // Make textareas use a monospace font
 textarea {
-    font-family: var(--bs-font-monospace);
+    font-family: $font-family-code;
+    font-size: $code-font-size !important;
```
NGPixel, line 84: Usually not a good idea to force `!important` on a base tag... Makes overriding unnecessary difficult in the future.

I would suggest creating a dedicated class with the `!important` font-size if necessary, and apply that class where required.

NGPixel (changes requested):

NGPixel (approved):

PR #4217, opened 2022-07-14
fix: set vite base path during deploy build + allow google fonts

NGPixel:


dev/deploy/build.sh
```diff
@@ -3,5 +3,9 @@
 echo "Compiling native node packages..."
 yarn rebuild
 echo "Packaging static assets..."
-yarn build
+if [ "$SHOULD_DEPLOY" = true ]; then
```
rjsparks, line 6: I suggest you use this instead
```
if [ "${SHOULD_DEPLOY} == "true" ]; then
```

ietf/utils/test_runner.py
```diff
@@ -852,6 +852,10 @@ def setup_test_environment(self, **kwargs):
                     "script-type": "off",
                     # django-bootstrap5 seems to still generate 'checked="checked"', ignore:
                     "attribute-boolean-style": "off",
+                    # self-closing style tags are valid in HTML5. Both self-closing and non-self-closing tags are accepted. (vite generates self-closing link tags)
+                    "void-style": "off",
```
rjsparks, line 856: This is a bigger hammer than I think we want - I had hoped to scope the skip to the -neue template.  We'll want to have a discussion with Lars about what amounts to code style for these overall.

rjsparks (changes requested):

rjsparks (approved):

PR #4226, opened 2022-07-18
fix: Fix removetags

larseggert:
I don't think this ever worked.


rjsparks (approved):

rjsparks:
We really should start adding unit-tests for these kinds of functions

PR #4244, opened 2022-07-23
fix: send status change eval and lc to IANA. Fixes #3504.

### LINKED ISSUE #3504: Status change messages to IANA
`type_defect`   |    by cmorgan@amsl.com

___

Can these addresses please be added to the lists of recipients for status change notifications (like “Moving RFC 4491 to Historic”)?

For status change last calls: drafts-lastcall@icann.org

For status change evaluations: drafts-eval@icann.org

(IANA is already receiving the status change approval messages).

Thank you!

___
_Issue migrated from trac:3504 at 2022-03-04 09:23:54 +0000_

rjsparks:


ietf/doc/views_ballot.py
```diff
@@ -679,9 +679,7 @@ def ballot_writeupnotes(request, name):
                     send_mail_preformatted(request, msg, override=override)
 
                     addrs = gather_address_lists('ballot_issued_iana',doc=doc).as_strings()
-                    override={ "To": "IANA <%s>"%settings.IANA_EVAL_EMAIL, "Bcc": None , "Reply-To": []}
-                    if addrs.cc:
-                        override['CC'] = addrs.cc
+                    override={ "To": addrs.to, "Bcc": None , "Reply-To": [], "CC": addrs.cc or None }
```
jennifer-richards, line 682: This changes the treatment of `override['CC']` from `'CC' in override == False` to `override['CC'] = None`. I think that means it will actively strip an existing `CC` instead of simply not injecting one. Is that desired/tolerable?

jennifer-richards (changes requested):

ietf/doc/views_ballot.py
rjsparks: so, in this case, yes, that's what we want. What was there before was only working because the CC pulled out of `iesg_ballot_issued` _happened_ to be empty.

In a later pass, I will rework this to not use override, but build the message directly from the template as done for the other half of this PR. I didn't do it here yet because I want it to be part of a more global search and eradication of this pattern.

jennifer-richards (approved):

PR #4249, opened 2022-07-23
Fix: Add 'back' button instead of 'cancel' to session request page

### LINKED ISSUE #4007: "Cancel" button doesn't work as expected on Meeting Request page
### Describe the issue

When you press "Cancel" button at the end of a meeting request page for a particular WG, the result is that you are being scrolled to the top of the page. Nothing more happened. Currently the only way to escape from the meeting request page is to request a session (or press browser's "Back" button).

### Code of Conduct

meadmaker:
The 'Cancel' button that goes back to the same page was confusing, so change it out for a 'Back' button instead.

Fixes #4007


ietf/secr/templates/includes/sessions_request_form.html
```diff
@@ -118,7 +118,7 @@
     <div class="button-group">
         <ul>
             <li><button type="submit" name="submit" value="Save"{% if is_locked %} disabled{% endif %}>Save</button></li>
-            <li><button type="submit" name="submit" value="Cancel">Cancel</button></li>
+            <li><button type="button" onclick="window.location='{% url "ietf.secr.sreq.views.main" %}'">Back</button></li>
```
larseggert, line 121: This should simply be an `a` tag with an `href`, rather than needing js.

larseggert (changes requested):

ietf/secr/templates/includes/sessions_request_form.html
larseggert, line 121:
```suggestion
            <li><a type="button" href="{% url 'ietf.secr.sreq.views.main' %}">Back</a></li>
```

larseggert: Or does this template not use bs5?

meadmaker: Bootstrap is available, but not used on the overall template.  I could pull it in if desired.

If so, then I'll probably make the same changes to `ietf/secr/templates/sreq/view.html` as well, since I copied it out of there.

meadmaker: Oh, I just realized this is wrong anyway - I should have it go back through the history instead of specifying a specific target, because this form is included on two paths, which have different navigation.

meadmaker:
I've now made that change, and it's ready for re-review.

larseggert (approved):

rjsparks:
Do not put the effort into bs5 this template. We will be replacing /secr/sreq with a reimplementation in /meeting later instead.
If this has become complicated to address, just remove the Cancel button and leave it to the user to use their browser back button.

rjsparks (approved):

PR #4256, opened 2022-07-23
Fix4209

### LINKED ISSUE #4209: WG draft "Submit to IESG for publication" workflow is infuriating
### Describe the issue

As a WG chair and shepherd of a document, once the shepherd writeup is complete I want to do two things. First, upload the writeup to the datatracker. Second, move the document into the next state.

To do step one, I navigate to the document page (e.g. https://datatracker.ietf.org/doc/draft-ietf-quic-version-negotiation/) and in the `Stream` section is a `Shepherd write-up` field with an `Edit` button that I hit, which lets me upload the doc. That all works ok, brilliant.

To do step two, I navigate to the document page and in the `Stream` section is the field WG state, that tells me "WG Consensus: Waiting for Write-up". Since I just uploaded the write-up, I want to change this state. So I press `Edit` and I'm taken to the "Change IETF WG state" page. 

Screenshot from 2022-07-11 18-02-13

The page very helpfully puts up the text `Move document to the recommended next state:` and presents a button labelled `Submitted to the IESG for publication`. When I click the button, the state updates in the dropdown box. Then I scroll down and hit the button labelled `Submit`. And it fails!

Screenshot from 2022-07-11 17-55-18

The error that is reported is

> You may not set the Submitted to IESG for Publication state using this form. Use the "Submit to IESG for publication" button on the document's main page instead. If that button does not appear, the document may already have IESG state. Ask your Area Director or the Secretariat for help.

The message is clear and the suggested action works. But this transition phase catches me out everytime, probably because my memory is short and I trust the datatracker to suggest useful tips to me. Instead, the datatracker suggests something that it then immediately tells me was the wrong thing to do. I'd be ever so grateful if this workflow could be improved. I have two suggestsions

1) If a different button is needed to be clicked, and you know that, just put the button on this page.
2) If you can't put the submit button on this page (for technical or other reasons) then don't suggest I waste my time doing anything else on this page. I.e. disable all the controls and just present the message in red immediately.

### Code of Conduct

russhousley:
Fix: Provide a button to Submit to IESG on the Change IETF WG state when that is the recommended next state.
Fixes #4209.

### DIFF
--- a/ietf/doc/tests_draft.py
+++ b/ietf/doc/tests_draft.py
@@ -1767,7 +1767,7 @@ class ChangeStreamStateTests(TestCase):
     def test_pubreq_validation(self):
         role = RoleFactory(name_id='chair',group__acronym='mars',group__list_email='mars-wg@ietf.org',person__user__username='marschairman',person__name='WG Cháir Man')
         RoleFactory(name_id='delegate',group=role.group,person__user__email='marsdelegate@ietf.org')
-        draft = WgDraftFactory(group=role.group,stream_id='ietf')
+        draft = WgDraftFactory(group=role.group)
 
         url = urlreverse('ietf.doc.views_draft.change_stream_state', kwargs=dict(name=draft.name, state_type="draft-stream-ietf"))
         login_testing_unauthorized(self, "marschairman", url)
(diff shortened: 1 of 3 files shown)


larseggert:
I guess we both worked on this: #4254

russhousley:
I like what you did for the template better than my approach.  However, I think that the change in the view to remove sub-pub from the pull down is also needed.

larseggert:
Do you want to combine the two into your PR?

russhousley:
Yours is easier.  You just need to add one line to /ietf/doc/views_draft.py.

I just need to take line 1567 from https://github.com/russhousley/datatracker/blob/fix4209/ietf/doc/views_draft.py

The other changes in my PR can be ignored because you made a more robust fix in the template.

larseggert:
Done! Please check?

rjsparks:
Closing this as OBE by #4254

Issue #4242, opened 2022-07-23
Make AD work queues more accable


kivinen:
### Description

I.e. create page https://datatracker.ietf.org/doc/ad/ which lists all area directors and points to individual queue pages https://datatracker.ietf.org/doc/ad/martin.duke etc.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #4259, opened 2022-07-23
feat: Add list page as requested in #4242

kivinen:
Add new page  to list all area directors and their current workload. Include links to the specific dashboards
for each area director. This new page is in doc/ad/.


ietf/doc/views_search.py
```diff
@@ -307,6 +307,32 @@ def cached_redirect(cache_key, url):
...
+def ad_dashboard_group_type(doc):
+    # Return group type for document for dashboard.
+    # If doc is not defined return list of all possible
+    # group types
+    if not doc:
+        return ('I-D', 'RFC', 'Conflict Review', 'Status Change', 'Charter')
+    if doc.type.slug=='draft':
+        if doc.get_state_slug('draft') == 'rfc':
+            return 'RFC'
+        elif doc.get_state_slug('draft') == 'active' and doc.get_state_slug('draft-iesg') and doc.get_state('draft-iesg').name in ('RFC Ed Queue Internet-Draft'):
```
jennifer-richards, line 319: Should that be `...in ('RFC Ed Queue', 'Internet-Draft')`

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -779,3 +779,10 @@ def is_valid_url(url):
     except ValidationError:
         return False
     return True
+
+@register.filter
+def get_hash(h, key):
```
jennifer-richards, line 784: Suggest naming this `get_from_dict` or something that better matches the dict Python data type. (There's a similar filter `lookup` in `agenda_custom_tags.py` that could perhaps be moved and reused - this is a better place for it.)

ietf/doc/views_search.py
```diff
@@ -307,6 +307,32 @@ def cached_redirect(cache_key, url):
...
+            return 'RFC'
+        elif doc.get_state_slug('draft') == 'active' and doc.get_state_slug('draft-iesg') and doc.get_state('draft-iesg').name in ('Dead', 'I-D Exists', 'AD is watching'):
+             return None
+        elif doc.get_state('draft').name in ('Expired', 'Replaced'):
+            return None
+        else:
+            return 'I-D'
+    elif doc.type.slug=='conflrev':
```
jennifer-richards, line 327: Similar to @rjsparks's comment above would be better to make use of the `DocTypeName` data rather than repeating it here. A single `doc.type.slug in []` test could replace a bunch of these cases

ietf/doc/views_search.py
rjsparks: This looks like a real bug. Please address it, and consider the other feedback.

rjsparks (changes requested):
(See previous comments for requested change)

kivinen:
Jennifer Richards writes:
> +        elif doc.get_state_slug('draft') == 'active' and doc.get_state_slug('draft-iesg') and doc.get_state('draft-iesg').name in ('RFC Ed Queue Internet-Draft'):
> 
> Should that be ...in ('RFC Ed Queue', 'Internet-Draft')

Actually it should be 'RFC Ed Queue' without the Internet-Draft. It
worked as 'RFC Ed Queue Internet-Draft' did include RFC Ed Queue that
is in the draft-iesg name.

The ad_dashboard_group do return "RFC Ed Queue Internet-Draft", which
was why I had cut & pasted it here.
-- 
***@***.***

rjsparks:
There are other comments to consider, but I wanted to point out:
the change at [76df542](https://github.com/ietf-tools/datatracker/pull/4259/commits/76df542a8e53f978ad28708841d0786c31b6c416) should just be to use `==` instead of `in`.
You are getting lucky that a string in parenthesis is a string, not a tuple, and it leaves a trap for the maintainer:
```
>>> 'foo' in 'foobar'
True
>>> 'foo' in ('foobar')
True
>>> 'foo' in ('foobar',)
False
```

rjsparks:
Rather than push into Tero's branch, I'm working in https://github.com/rjsparks/datatracker/tree/ad-list-page. I'll eventually make a PR that supersedes this one.

rjsparks:
Replaced by #4315.

PR #4261, opened 2022-07-24
Fix: corrects the IESG evaluation record summary for approved docs

### LINKED ISSUE #3962: Don't show needed ballot counts on documents that have already been approved
Having a page that says
```
 Summary: Needs one more YES or NO OBJECTION position to pass.
```
on a document that has already been approved is confusing and incorrect.

Spectre17:
Fixes #3962.

This is done by extracting the IESG state (if applicable) and checking the state for approved (announced or not announced) documents. Documents in this state are made to works like documents in the "rfcqueue" or "pub" states, i.e., they list the status of the last ballot.

### DIFF
--- a/ietf/doc/views_doc.py
+++ b/ietf/doc/views_doc.py
@@ -1216,6 +1216,7 @@ def document_ballot_content(request, doc, ballot_id, editable=True):
         else:
             position_groups.append(g)
 
+    iesg_state = doc.get_state("draft-iesg").slug
     if (ballot.ballot_type.slug == "irsg-approve"):
         summary = irsg_needed_ballot_positions(doc, [p for p in positions if not p.is_old_pos])
     else:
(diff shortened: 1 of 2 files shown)


rjsparks (approved):

ietf/doc/views_doc.py
```diff
@@ -1216,6 +1216,8 @@ def document_ballot_content(request, doc, ballot_id, editable=True):
         else:
             position_groups.append(g)
 
+    iesg = doc.get_state("draft-iesg")
```
rjsparks, line 1219: There is a get_state_slug function that would have done the same as these two lines

ietf/templates/doc/document_ballot_content.html
```diff
@@ -39,7 +39,7 @@
                 </p>
             {% endif %}
         {% endif %}
-        {% if not ballot_open %}
+        {% if not ballot_open or iesg_state == "ann" or iesg_state == "approved" %}
```
rjsparks, line 42: I'm a little uncomfortable that this now doesn't tell the truth - While in 'approved", the ballot is, in reality, still open as far as the datatracker is concerned, but the IESG will not make more changes to positions.It is in the transient phase waiting for the secretariat to clear, hence close, it at the time of announcement.
But the subtlety isn't really useful to the consumer of this message, so we'll go with this as is.

rjsparks (approved):

PR #4262, opened 2022-07-24
fix: Update timezone on the interim create page

### LINKED ISSUE #3898: Interim meeting time entry
### What happened?

When creating an interim meeting, the default timezone is 'UTC'. However, if this isn't modified before the meeting is created, the meeting ends up being scheduled in the browser's local timezone. This results in a meeting request at the wrong time.

The 'Time' component does indicate 'local time', but it's not clear what that is in relation to -- is it 'local' to the meeting being scheduled, or 'local' to the user? The second interpretation isn't intuitive.

Safari

### Code of Conduct

meadmaker:
The interim creation screen will set the timezone to the timezone of
the browser.  Unfortunately, the UI element was never refreshed, so
it remained on its loaded default, thus telling the user the wrong
value.

This change adds a timezone widget refresh to the end of the function that
loads the browser's local timezone into the timezone value.

This change also adds code to update the 'Local timezone' help text
for the start time and end time input bozes to
call out the selected timezone specifically.

(Finally, it adds a newline at the end of the file, since none was present previously)

Fixes #3898

### DIFF
--- a/ietf/static/js/meeting-interim-request.js
+++ b/ietf/static/js/meeting-interim-request.js
@@ -233,7 +233,7 @@ const interimRequest = (function() {
         },
 
         timezoneChange: function () {
-            var tzname = $(this).val();
+            const tzname = $(this).val();
             $("input[name$='-time']")
                 .trigger('blur');
             $("input[name$='-end_time']")
--- a/ietf/static/js/meeting-interim-request.js
+++ b/ietf/static/js/meeting-interim-request.js
@@ -185,6 +185,8 @@ const interimRequest = (function() {
                     .length > 0) {
                     $('#id_time_zone')
                         .val(tzname);
+                    $('#id_time_zone')
+                        .trigger('change');
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


ietf/static/js/meeting-interim-request.js
```diff
@@ -231,10 +233,14 @@ const interimRequest = (function() {
         },
 
         timezoneChange: function () {
+            var tzname = $(this).val();
```
jennifer-richards, line 236: Should use `const tzname` - we're trying to rid the codebase of `var`

jennifer-richards (changes requested):
Looks good, just some minor lint.

It strikes me as a little awkward that we now have to hook up the change event on the timezone input before we call `initTimeZone()`. It might be worth adding a comment to discourage someone from "helpfully" switching that around.

NGPixel (approved):

jennifer-richards (approved):

PR #4263, opened 2022-07-24
Fix: Add a link to the interim request screen to the meeting menu

### LINKED ISSUE #3897: Scheduling an interim
### What happened?

When a chair is logged in, the 'meetings' menu has no option to schedule an interim meeting; instead, they need to open the 'documents' tab and select 'xxxx wg meetings'.

It'd make a lot more sense if the 'xxxx wg meetings' was under the 'meetings' menu, rather than the 'documents' menu.

Not Applicable

### Code of Conduct

meadmaker:
The subject pretty much says it all.

Fixes #3897

### DIFF
--- a/ietf/templates/base/menu.html
+++ b/ietf/templates/base/menu.html
@@ -272,6 +272,12 @@
     </li>
     <li>
         <a class="dropdown-item {% if flavor != 'top' %}text-wrap link-primary{% endif %}"
+           href="{% url 'ietf.meeting.views.interim_request' %}">
+            Request an interim
+        </a>
+    </li>
+    <li>
+        <a class="dropdown-item {% if flavor != 'top' %}text-wrap link-primary{% endif %}"
            href="{% url 'ietf.secr.sreq.views.main' %}">
             Request a session
         </a>


ietf/templates/base/menu.html
```diff
@@ -270,6 +270,12 @@
             Important dates
         </a>
     </li>
+    <li>
```
rjsparks, line 273: This, like request a session below it, really needs to be conditional.
If someone _cant_ request a session or an interim, it is wrong to add this link. 
Unfortunately, this is being done in the base menus, and unless we're already paying the price to calculate whether the current requester can do these things, the most we should do is switch them on "logged in" to avoid making page loads even slower.
In the long run, this is something we can rework with with async population of the menu, but we're just beginning to build the scaffolding for that kind of thing.

So, for now, I'm of mixed mind. Do we continue down the current path of providing menu entries that are ultimately useless for most users, or do we stop now and make this _slightly_ better.

rjsparks:
Mark made a second request in the ticket - to move the <group foo> meetings entries from the Group menu to the Meeting menu. I'm not sure this won't surprise people who are already used to looking under Group.

ietf/templates/base/menu.html
larseggert: Agreed. There are not many users with permissions to request interims. I think it is fair to expect that those are familiar enough with the DT to find the link on their group page.

rjsparks:
@meadmaker I'm going to close this one as wontmerge. I think we need to have a longer discussion about what to do in the original request from Mark Nottingham.

PR #4294, opened 2022-07-28
fix: Remove unicode encode from CSV row writer

### LINKED ISSUE #4292: cvs export of group documents is dumping b'' strings
From Tim Wicinski:
>FYI
>
>So I went and dumped the WG documents via the CSV link:
>
>https://datatracker.ietf.org/group/dnsop/documents/csv/
>
>Then I imported them into the googles sheets.
>The CSV file has all strings as "b'<string>'" which
>makes the import needing a bit of cleanup. 
>
>tim

cc @moonshiner

larseggert:
Fixes #4292


jennifer-richards (approved):
Looks good to me. I changed the PR title from 'CVS' to 'CSV'.

PR #4295, opened 2022-07-28
fix: Move Meetecho buttons to separate button group

### LINKED ISSUE #4288: The "Join Video/Meetecho" button on the agenda page should be larger/brighter than the other buttons
### Description

This would make it much easier to find the button that probably many people use most during the meeting.

### Code of Conduct

larseggert:
Fixes #4288


rjsparks:
@NGPixel if we make this change, a similar change to the new agenda pages should happen at the same time.

NGPixel:
The new agenda doesn't use a single button group. All buttons are separated. The buttons could be re-ordered to match this change though.

NGPixel (approved):

PR #4297, opened 2022-07-28
fix: Don't require a responsible AD to be identified for new status change

larseggert:
Fixes #4159

### DIFF
--- a/ietf/doc/views_status_change.py
+++ b/ietf/doc/views_status_change.py
@@ -465,7 +465,7 @@ class StartStatusChangeForm(forms.Form):
     document_name = forms.CharField(max_length=255, label="Document name", help_text="A descriptive name such as status-change-md2-to-historic is better than status-change-rfc1319.", required=True)
     title = forms.CharField(max_length=255, label="Title", required=True)
     ad = forms.ModelChoiceField(Person.objects.filter(role__name="ad", role__group__state="active",role__group__type='area').order_by('name'), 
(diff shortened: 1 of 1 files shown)


ietf/doc/views_status_change.py
```diff
@@ -465,7 +465,7 @@ class StartStatusChangeForm(forms.Form):
     document_name = forms.CharField(max_length=255, label="Document name", help_text="A descriptive name such as status-change-md2-to-historic is better than status-change-rfc1319.", required=True)
     title = forms.CharField(max_length=255, label="Title", required=True)
     ad = forms.ModelChoiceField(Person.objects.filter(role__name="ad", role__group__state="active",role__group__type='area').order_by('name'), 
-                                label="Shepherding AD", empty_label="(None)", required=True)
+                                label="Shepherding AD", empty_label="(None)", required=False)
```
rjsparks, line 468: This has to come with a migration. (The tests should really complain if migrations are missing).

rjsparks (changes requested):

ietf/doc/views_status_change.py
rjsparks: Nevermind - reading too fast and was thinking models rather than forms.

rjsparks (approved):

rjsparks:
Shouldn't have approved this - the tests still required that the forms fail if you don't supply an AD. Will provide a fix, but yet another place where we are letting failing tests slide and _real_ issues slide through with them.

PR #4305, opened 2022-08-01
chore: remove unneeded logging and code marked unreachable that has not been reached.

rjsparks:


ietf/doc/models.py
```diff
@@ -1238,8 +1227,6 @@ def __str__(self):
 
     def save(self, *args, **kwargs):
```
jennifer-richards, line 1228: Get rid of this now redundant method?

ietf/doc/views_doc.py
```diff
@@ -1972,7 +1972,7 @@ def rfcdiff_latest_json(request, name, rev=None):
             if doc.name != doc.canonical_name():
                 prev_rev = doc.rev
                 # not sure what to do if non-numeric values come back, so at least log it
-                log.assertion('doc.rfc_number().isdigit()')
+                log.assertion('doc.rfc_number().isdigit()') # .rfc_number() is expensive...
```
jennifer-richards, line 1975: Expensive, but at least cached. At the moment the value is used again in real code immediately after - could catch exceptions there instead of asserting ahead of time to reduce the risk that the assertions outlive the code they protect.

jennifer-richards (approved):
A couple small comments inline. Probably should remove the empty save() method but the other is just a thought.

ietf/doc/models.py
rjsparks: I don't think this is redundant? It's possible (and I think necessary in some places) to save a docevent without saving a change to the referenced Document, and we have a lot of construct that creates and saves docevents before calling save_with_history on Document. Do I misunderstand what you're pointing to as redundant?

rjsparks: Or is the point that we should just rely on models.Model.save()?

ietf/doc/views_doc.py
rjsparks: Well, as you note, it's not clear what to do if the expectations of those functions are violated, and I view the assertions as a short term safety net checking that we didn't have broken data. We could probably just pull the asserts, but since they aren't really adding load, leaving them for awhile longer won't hurt us.

PR #4345, opened 2022-08-18
fix: allow 90m session requests. Prompt for cannot meet at time reasoning.

rjsparks:


rjsparks:
This is a quick fix for new 115 requirements. In the future, we should make session duration options a configuration point for meetings.

larseggert:
So do you think this will be ready Monday? Or should I notify the IESG of potential delays?

rjsparks:
Depends on build and test.

jennifer-richards (approved):

PR #4356, opened 2022-08-23
Fix the interim meeting request approval form

### LINKED ISSUE #4355: Interim approval view does not work
A leftover issue from the bootstrap 5 facelift, I think (cc @larseggert):

The template at ietf/templates/meeting/interim_request_details.html does not create a form, but has some form-like buttons remaining on it.

The view expects the user to POST a form.

larseggert:
Fixes #4355

### DIFF
--- a/ietf/templates/meeting/interim_request_details.html
+++ b/ietf/templates/meeting/interim_request_details.html
@@ -103,54 +103,54 @@
             {% endif %}
         {% endfor %}
     </dl>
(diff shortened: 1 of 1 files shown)


rjsparks:
If the tests for this view used WebTest, the broken form would have been caught. We may want to consider a moving towards much more use of WebTest for views that have forms in them.

rjsparks (approved):

PR #4363, opened 2022-08-24
doc: update INSTALL to reflect needed roles for each step

rjsparks:


dev/INSTALL
```diff
@@ -77,7 +81,7 @@ General Instructions for Deployment of a New Release
 
  12. Start the datatracker service (it is no longer necessary to restart apache) ::
 
-      exit # or CTRL-D, back to root level shell
+      exit 
       sudo systemctl start datatracker.service datatracker.socket
 
  13. Start async task worker and message broker:
```
jennifer-richards, line 87: Did you want to call out the switch back to the docker-group user?

dev/INSTALL
```diff
@@ -138,7 +142,7 @@ The following process should be used:
  7. Stop the async task container (this may take a few minutes if tasks are in progress):
 
       cd /a/docker/datatracker
-      docker-compose stop celery
+      docker-compose down
```
jennifer-richards, line 145: good catch

jennifer-richards (approved):

dev/INSTALL
rjsparks: I'll leave it as is for now.

PR #4368, opened 2022-08-25
Fix typo on https://datatracker.ietf.org/help/state/draft/iesg where …

ghost:
…'publication' contains an 's' in the 'I-D Exists' state description.

### DIFF
--- a/ietf/doc/migrations/0007_idexists.py
+++ b/ietf/doc/migrations/0007_idexists.py
@@ -18,7 +18,7 @@ def forward(apps, schema_editor):
         slug = 'idexists',
         name = 'I-D Exists',
         used = True,
-        desc = 'The IESG has not started processing this draft, or has stopped processing it without publicastion.',
+        desc = 'The IESG has not started processing this draft, or has stopped processing it without publication.',
         order = 0,
     )
     idexists.next_states.set(State.objects.filter(type='draft-iesg',slug__in=['pub-req','watching']))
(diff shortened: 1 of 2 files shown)


ietf/doc/migrations/0007_idexists.py
```diff
@@ -18,7 +18,7 @@ def forward(apps, schema_editor):
         slug = 'idexists',
         name = 'I-D Exists',
         used = True,
-        desc = 'The IESG has not started processing this draft, or has stopped processing it without publicastion.',
+        desc = 'The IESG has not started processing this draft, or has stopped processing it without publication.',
```
rjsparks, line 21: I don't think this will cause trouble, but it functionally won't get used. The tests use a fixture rather than a from-scratch migration, and development instances start with a recent production database dump. I'm not confident that a from-empty migration construction of the database would produce anything useful at this point.

I bring this up to point out that in future work, touching old migrations is probably not the right thing to do.

rjsparks (approved):

ietf/doc/migrations/0007_idexists.py
rjsparks: Also to have an effect on production, a _new_ migration would be needed. But for simple string changes like these, we've been using the admin to make the adjustment.

rjsparks: Change made - rather than merge this, we'll pick up the change to the test fixture with the next commit that affects the tests.

jfinkhaeuser: You're welcome!

PR #4370, opened 2022-08-25
chore: migrate timestamps for use with USE_TZ=True

jennifer-richards:


ietf/meeting/models.py
```diff
@@ -355,19 +355,17 @@ def build_timeslices(self):
...
+            tzfn = os.path.join(settings.TZDATA_ICS_PATH, self.time_zone + ".ics")
+            if os.path.exists(tzfn):
+                with io.open(tzfn) as tzf:
+                    icstext = tzf.read()
+                vtimezone = re.search("(?sm)(\nBEGIN:VTIMEZONE.*\nEND:VTIMEZONE\n)", icstext).group(1).strip()
+                if vtimezone:
+                    vtimezone += "\n"
+                return vtimezone
+        except IOError:
+            pass
```
rjsparks, line 368: double-check: it's ok for this to return None? (It's what it did before, but we may never have encountered an IOError...

rjsparks (approved):

ietf/meeting/models.py
jennifer-richards: Looks like it's only used in one place and it's guarded by a check for truthiness, so it should be ok.

jennifer-richards: I'm going to make it explicit that it returns `None`, though. Falling off the end of a method body gives me the heebie-jeebies.

PR #4379, opened 2022-08-26
fix: allow shepherds for editorial stream documents. Fixes #4333

### LINKED ISSUE #4333: Add document shepherd support to the Editorial stream
The RSAB/RSWG would like to make use of document shepherds during the document approval process.

rjsparks:
Testing the render of the document_main view (to see if the shepherd line shows up in this case, and that if you have permissions to change it, the edit button shows up, etc). is egregious, the effort is better deferred until the view has its front-end replaced with Vue.

### DIFF
--- a/ietf/templates/doc/document_draft.html
+++ b/ietf/templates/doc/document_draft.html
@@ -529,7 +529,7 @@
                     </td>
                 </tr>
             {% endif %}
-            {% if doc.stream_id == 'ietf' or doc.stream_id == 'ise' or doc.stream_id == 'irtf' or doc.stream_id == 'editorial' %}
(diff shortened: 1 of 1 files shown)


rjsparks:
Fixes #4333

jennifer-richards (approved):
Looks good. I assume it's intended that this will (if I'm reading correctly) replace the RFC Editor note with the shepherd fields, as is done for other shepherded types.

rjsparks:
Good thing to ask about - there is an assumption that the ability to affect the RFC editor note for the document is off in the balloting scaffolding. 

We haven't done the work to add that scaffolding for the editorial stream yet, but we will before the RSAB would request publication of anything, so we can live with the disconnect for now.

PR #4380, opened 2022-08-26
feat: show active adhoc groups on group menu. Fixes #4339.

### LINKED ISSUE #4339: `hotrfc` not accessible from "Groups" menu
### Describe the issue

https://datatracker.ietf.org/group/hotrfc/about/ exists, but I can't figure out how to get to it from within the menu structure.

### Code of Conduct

rjsparks:
This is *gross*.

Please pull the PR and look at the result. I don't think it's what we really want.

### DIFF
--- a/ietf/doc/templatetags/active_groups_menu.py
+++ b/ietf/doc/templatetags/active_groups_menu.py
@@ -9,7 +9,7 @@ from ietf.name.models import GroupTypeName
 register = template.Library()
 
 parents = GroupTypeName.objects.filter(
-    slug__in=["ag", "area", "rag", "team", "dir", "program"]
+    slug__in=["ag", "area", "rag", "team", "dir", "program", "adhoc"]
 )
 
 others = []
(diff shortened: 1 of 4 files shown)


rjsparks:
image

rjsparks:
Fixes #4339

larseggert:
Why is the stuff in the screenshot gross? I think this is better than not showing these at all.

larseggert:
Would a better approach to be to give these groups different shapes/homes, e.g., directorates, teams, etc.?

larseggert:
All of these could become teams under GEN, IMO, and we could eliminate the concept of an "ad hoc group".

rjsparks:
Making them genarea teams would address my discomfort completely - would, e.g, the IEPG organizers have any issue with that?

larseggert:
It was proposed that they could be under OPS, which IMO makes sense.

rjsparks:
I've done this to the sandbox (effects will only last until it resets its database tomorrow).
If we go down this path, I will need to make it so that teams have full meeting materials support (which should just be groupfeatures configuration). Where/if these things appear in the group filters on the meeting agenda page will also change.

rjsparks:
Changes made to the production database - closing this as OBE

PR #4382, opened 2022-08-29
fix: Static paths don't need to be prefixed with the base URL

larseggert:


NGPixel:
Does removing the base path still produce an absolute URL? As even though relative URLs will _work_ on Facebook, the opengraph spec requires an absolute URL.

See https://stackoverflow.com/a/9858694

larseggert:
It does, at least in production. (I was putting a draft on Facebook and the image didn't load, so I checked the HTML source and the attribute was incorrect.)

NGPixel (approved):

rjsparks (approved):

rjsparks:
`{% static foo/bar %}` follows configuration to define how it expands. In our case, the relevant config is currently at https://github.com/ietf-tools/datatracker/blob/main/ietf/settings.py#L163, but note that we really need to bring this back to being served by datatracker instead of www now that datatracker is CDNed.

PR #4385, opened 2022-08-29
fix: show()/hide() timeslots/sessions on schedule editor

### LINKED ISSUE #4375: Filtering group types is not quite completely working
From @flynnliz

On Wednesday during lunch at 12:15 there’s a WG Chairs lunch that’s held in one of the regular session meeting rooms. This timeslot is marked as “Other” but I cannot get it to disappear from the grid. Even when I have selected that the Timeslot Types visible should only be Regular, this lunch session still appears. I have also tried marking that time slot as Registration or other random things just to get it to go away, but it still appears in the middle of the grid. It pushes out the rest of the day on Wednesday so the session time slots don’t line up — see screenshot. The regular sessions not aligning vertically makes it really difficult to re-schedule any sessions needed so we can accurately see conflicts. For example, if I want to check which sessions can go in that one empty spot, I also have to look at LPWAN to its right which is the session that conflicts with everything else in that column.

Also, that WG Chairs lunch session is actually from 12:15-13:15 but the header on its rectangle lists 13:30-14:30 as it’s trying to fit itself into the rest of that column, which just adds to the confusion of what times those lpwan and idr sessions are.

image

jennifer-richards:
Fixes #4375

### DIFF
--- a/ietf/static/js/edit-meeting-schedule.js
+++ b/ietf/static/js/edit-meeting-schedule.js
@@ -882,8 +882,12 @@ $(function () {
      * Responsible for final determination of whether a timeslot is visible, invisible, or hidden.
      */
     function updateTimeSlotVisibility() {
-        timeslots.not(classes_to_hide).removeClass('hidden');
-        timeslots.filter(classes_to_hide).addClass('hidden');
+        const tsToShow = timeslots.not(classes_to_hide);
+        tsToShow.removeClass('hidden');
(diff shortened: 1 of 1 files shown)


jennifer-richards:
(There's still room to clean up the handling of showing/hiding, but this is the low-hanging fix.)

rjsparks (approved):

PR #4389, opened 2022-08-30
fix: don't migrate dumpinfo. automate migrations and setting USE_TZ.

rjsparks:


jennifer-richards (approved):
+1 for the branch name

PR #4402, opened 2022-09-01
fix: fix failing tests and eliminate naive datetime warnings

jennifer-richards:
All tests now pass and no datetime warnings are emitted during the tests.

I've replaced the formerly unused `PRODUCTION_TIMEZONE` setting with the constant `DEADLINE_TZINFO` in `ietf/utils/timezone.py`. The constant is a `tzinfo` rather than a string. My intention was to use this when silently converting `date` instances to `datetime` instances where the current behavior assumes pacific time. In other cases, I've generally let things default to UTC.

That's the intention, but obviously, trust no one when time zones are involved.


rjsparks (approved):

rjsparks:
One thought: The DEADLINE_TZINFO idea is invoked when we have endpoint times, but may need to apply for start times as well.

jennifer-richards:
Good point. I meant to use it for any timestamp that represents either end of an interval, including where comparing "today" to endpoints.

PR #4415, opened 2022-09-05
feat: Add column sums to `/doc/ad` dashboard

larseggert:
Also add trends (shown only to ADs) and clean up the display a bit.


rjsparks:
You keep tweaking this - should it be marked draft?

larseggert:
Nope, I think this is ready for review!

rjsparks:
This is was overlycomplicated code when it came in, and it's getting even more baroque with these changes. It really needs to be redone to not be based on SearchForm, but to make the right direct queries through the ORM instead.

We can let this in, in the spirit of prototyping, but it would be better to wait to add any more significant features to it until we can get to that rewrite.

ietf/doc/views_search.py
```diff
@@ -425,106 +458,211 @@ def ad_dashboard_sort_key(doc):
 
     return "3%s" % seed
 
+
 def ad_workload(request):
+    delta = datetime.timedelta(days=30)
+    today = datetime.date.today()
```
jennifer-richards, line 464: For better timezone behavior (and compatibility with the timezone-aware work going on right now), it'd be better to use `right_now = timezone.now()` (imported with `from django.utils import timezone`) instead of computing `today()`.

(other changes will be needed, noted inline)

```suggestion
    right_now = timezone.now()
```

ietf/doc/views_search.py
```diff
@@ -425,106 +458,211 @@ def ad_dashboard_sort_key(doc):
...
+                ad.counts[group_type][groups[group_type][group]] += 1
+                ad.doc_now[group_type][groups[group_type][group]].add(doc)
+
+                try:
+                    state_date = (
+                        doc.docevent_set.filter(
+                            Q(type="started_iesg_process") | Q(type="changed_state")
+                        )
+                        .order_by("-time")[0]
+                        .time.date()
```
jennifer-richards, line 601:
```suggestion
                        .time
```

ietf/doc/views_search.py
```diff
@@ -425,106 +458,211 @@ def ad_dashboard_sort_key(doc):
...
+                    )
+
+                except IndexError:
+                    state_date = datetime.date(1990, 1, 1)
```
jennifer-richards, line 605:
```suggestion
                    state_date = datetime.datetime(1990, 1, 1, tzinfo=datetime.timezone.utc)
```

jennifer-richards (dismissed):
Adding `datetime.date.today()` will introduce something between a quirk and a bug when our current work to add timezone awareness to the datatracker is merged in. I've suggested an alternative that is better compatible.

My suggested code will change the behavior at the ends of the window - instead of cutting it off at midnight server time, it will look back at the last 30 days from the current time. If that's not tolerable, I can suggest alternative code but I recommend doing it this way if possible.

larseggert:
@rjsparks om happy to rework this, since it is t urgent or anything, but will need a bit of help with the database side of things.

rjsparks:
@larseggert if you don't see a problem with Jennifer's suggestions, we can ask her to make them.

larseggert:
She totally can, or I can next week.

rjsparks (approved):

PR #4417, opened 2022-09-06
ui: Add table dividers

larseggert:
And add some &shy; hyphens while we're here.


ietf/templates/nomcom/private_index.html
```diff
@@ -11,21 +11,21 @@ <h2 class="mt-3">Nomination status</h2>
     <table class="table table-sm table-striped table-hover tablesorter">
         <thead>
             <tr>
-                <th scope="col" data-sort="position">Position</th>
+                <th scope="col" data-sort="position">Po&shy;si&shy;tion</th>
```
rjsparks, line 14: Wow, this makes the template almost unreadable to a human.
Perhaps there's a bit of template middleware somewhere that keeps a dictionary of good splits that a block of such things could be fed to? Or a templatetag at least.

larseggert:
There is https://django-softhyphen.readthedocs.io/en/latest/, but that says 
> (Warning! Because of its overhead, the filter is not recommended in production if it needs to run each time the page loads.)

rjsparks:
how badly do you need those hyphens at the moment. Could you back out that part of the change for now and we can investigate just how bad the performance overhead of django-softhyphen really is?

larseggert:
I'll take them out. They do make the layout on narrower screens break, since the table flows past the leftmenu, but that is minor.

larseggert:
Note that https://github.com/datadesk/django-softhyphen is archived and hasn't seen a commit in seven years...

rjsparks:
Thanks.

If we can't find a library someone else is supporting I would rather set up our own dict in utils for the words we want to soft-hyphenate and a template filter that would transform words using that dict than to manually, explicitly, soft-hyphenate each word each time it occurs. The latter raises the readability concern I already pointed to, and would likely lead to things like making programs scanning for spelling errors literals break.

rjsparks (approved):

PR #4421, opened 2022-09-08
fix: record and interpret RFC pub dates in correct timezone

jennifer-richards:


rjsparks (approved):

rjsparks:
Still twitchy about whether we have been doing the right thing synthesizing a datetime based on "now" when re-reading the rfc-editor index. Storing publications as an RFC as a docevent was probably not the best choice. As discussed when we refactor to make RFCs a document type, its publication date can be part of its model directly, and we can scrub these docevents.

jennifer-richards:
> Still twitchy about whether we have been doing the right thing synthesizing a datetime based on "now" when re-reading the rfc-editor index. Storing publications as an RFC as a docevent was probably not the best choice. As discussed when we refactor to make RFCs a document type, its publication date can be part of its model directly, and we can scrub these docevents.

Agree 110%. It's definitely not doing the right thing - e.g., I stumbled across [RFC 9225](https://datatracker.ietf.org/doc/rfc9225/) while testing, which is an April 1 RFC. Its timestamp, however, is April 2, which I suspect is due to the rfc editor sync not running on the 1st for some reason.

PR #4441, opened 2022-09-13
ui: Hyphenate table headers, and optionally force-wrap them

larseggert:


rjsparks (approved):

rjsparks:
@larseggert: Nick points out that `ui: whatever` is invalid and will be ignored when the release message is prepared.
You wanted `fix(ui): whatever` instead.
Sorry that I missed this before the merge.

PR #4451, opened 2022-09-14
fix: remove stream editor email duplication

### LINKED ISSUE #3638: Conflict Review request form should make it clear that the ise is already going to be copied
The Conflict Review request form should tell users that the ISE (rfc-ise@rfc-editor.org) is already copied on announcements, and strip that address from the "Send Notices To" field if a user adds it, so the secretariat need not remove it manually. The maintainer agreed to make this generic to all stream chairs (irtf-chair@irtf.org, iesg@ietf.org, iab-chair@iab.org), not just the ISE. See #3638.

### Task
- [ ] Inform users that stream-chair addresses are auto-copied and strip them from Send Notices To

---
**Agent-assisted triage**

- Suggested size: M
- Suggested priority: medium
- Suggested lane: needs-decision
- Analysis confidence: medium
- Updated with agent assistance on 2026-08-18
---

<details><summary>Original report</summary>

The Conflict Review request form has a field that populates the documents "Send Notices To".

The secretariat is currently dealing with a transition of the ISE, but the Send Notices To has explicitly named Adrian, and the secretariat is having to manually remove the name from things like approval announcements.

The form should both let the user know that the ise will get copied already showing them the rfc-ise@rfc-editor.org address, and clean it out of the Send Notices To field if it is added.

cc: @akvezza, @cindymorgan
</details>

meadmaker:
These changes will prevent the stream editor email address from being included in the `Notice emails` field during the creation of a conflict review.  This only works for documents in the IRTF or ISE streams.
- If the stream editor email address is detected in the input, then the address is silently removed.  
- The address is removed even if it has been entered multiple times.
- The email address suggestion generator has been updated to remove the stream editors
- The help text for the `Notice emails` field has been updated to say that the stream editors are automatically notified.

While I was in the code, I refactored a bit:
- The SimpleStartReviewForm is now a superclass to the StartReviewForm.  This was intended to ease the creation of a single `clean_notify()` validation method, which eventually moved to the view code instead.  However, the refactoring was left.
- The `start_review_as_secretariat()` and `start_review_as_stream_owner()` methods are now merged.  Ironically, this resulted in the exact same number of lines of code, but at least there is only one method to update going forward.

Fixes ietf-tools/datatracker#3638.


ietf/doc/views_conflict_review.py
```diff
@@ -369,14 +371,20 @@ def approve_conflict_review(request, name):
...
                                    ))
 
+
 class SimpleStartReviewForm(forms.Form):
-    notify = forms.CharField(max_length=255, label="Notice emails", help_text="Separate email addresses with commas.", required=False)
+    notify = forms.CharField(
+        max_length=255,
+        label="Notice emails",
+        help_text="Separate email addresses with commas. " + 
+                  "ISE and IRTF stream editors are notified automatically for their streams.",
```
rjsparks, line 380: stream editor -> stream manager

ietf/doc/views_conflict_review.py
```diff
@@ -392,10 +400,52 @@ def __init__(self, *args, **kwargs):
...
+
+    Start the conflict review process, setting the initial 
+    shepherding AD, and possibly putting the review on a telechat.'''
+
+    def cleaned_notify(notify, doc):
+        if doc.stream.slug not in ['irtf', 'ise']:
+            return notify
+        stream_managers = [
+            re.sub('([^<]*<)?(.*)(>.*)', '\2', r)
+            for r in Recipient(slug='stream_managers').gather(streams=[doc.stream.slug])
```
rjsparks, line 413: twitch: 
this really reads like it counts on side-effect. 
It might be better to create a new function in person/utils that returns the map of stream to manager addresses and have this code and the mailtrigger Recipient code use that instead. (It will make the explanation of the mailtrigger recipient more opaque, I'm really unsettled about Recipient being used this way.

ietf/doc/views_conflict_review.py
```diff
@@ -392,10 +400,52 @@ def __init__(self, *args, **kwargs):
...
+        ]
+        notifications = notify.split(',')
+        for mgr in stream_managers:
+            goods = [n for n in notifications if mgr not in n]
+            notifications = goods
```
rjsparks, line 418: This has to be a bug? You are calculating and throwing away a value for all but the last iteration of a for loop, and then using the value of the last iteration?

ietf/doc/views_conflict_review.py
```diff
@@ -409,13 +459,20 @@ def start_review_sanity_check(request, name):
 
     return doc_to_review
 
+
 def build_notify_addresses(doc_to_review):
```
rjsparks, line 463: I'm thinking this would be more maintainable if it used the above utility suggestion and was implemented using set (still returning list)

rjsparks (changes requested):

ietf/doc/views_conflict_review.py
meadmaker: This code does set subtraction of the items in stream_managers from the items in notifications.  It will:
- Iterate over item `mgr` in `stream_managers`
  - Determine every item in `notifications` that does not match `mgr`, and store that in `goods`
  - Copy `goods` over to `notifications`

Each pass through 417-418 ensures that one item out of the `stream_managers` list is not present in the `notifications` list.

The `goods` variable only exists here because I'm uncertain how well the language handles iterating over a variable as it mutates.  Assigning it to a temporary variable removes that ambiguity.

ietf/doc/views_conflict_review.py
meadmaker: I'll pull that into its own function.

Jennifer and I had gone back and forth about whether the list of stream managers should be coded in Python, or whether it should be somehow retrieved from the database.  The best that we had come up with for retrieving it from the database was to find the group associated with the document stream, then find the chair for that group, and use that chair's email address.  Which approach would you like to see?

rjsparks (commented):
Ugh - I see a bunch of comments in "Pending" from _months_ ago. Commiting them now just for the sake of history, but I'm going to close this PR and rework the problem to be solved.

PR #4470, opened 2022-09-16
fix: return same bibxml for versioned and versionless references

### LINKED ISSUE #4384: bibxml reference files no longer show all the name information
### Describe the issue

https://github.com/ietf-tools/bibxml-service/issues/280

Only `fullname` seems to be filled in any more.  `surname` or `initials` seem to be missing now.
This creates incorrect guesses as to what the surname and what the initial-generating names are.

This seems to have happened around August 18 to 20

### Code of Conduct

rjsparks:
Fixes #4384.

Refactors bibxml production to remove repeated logic.

Abandons the half-implemented idea that returning information from the Submission object might be better than the Document or DocHistory objects.

### DIFF
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -17,6 +17,7 @@ from urllib.parse import quote
 from django.conf import settings
 from django.contrib import messages
 from django.forms import ValidationError
+from django.http import Http404
 from django.template.loader import render_to_string
 from django.utils.html import escape
 from django.urls import reverse as urlreverse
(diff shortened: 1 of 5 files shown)


ietf/doc/utils.py
```diff
@@ -1317,3 +1318,33 @@ def fuzzy_find_documents(name, rev=None):
 
     FoundDocuments = namedtuple('FoundDocuments', 'documents matched_name matched_rev')
     return FoundDocuments(docs, name, rev)
+
+def bibxml_for_draft(doc, rev=None):
+
+    if rev != None and rev != doc.rev:
```
rjsparks, line 1324: Since the refactor makes this non-obvious, the primary fix is on this line at `and rev != doc.rev`

ietf/doc/utils.py
```diff
@@ -1317,3 +1318,33 @@ def fuzzy_find_documents(name, rev=None):
...
+        
+    latest_revision_event = doc.latest_event(NewRevisionDocEvent, type="new_revision")
+    latest_revision_rev = latest_revision_event.rev if latest_revision_event else None
+    best_events = NewRevisionDocEvent.objects.filter(doc__name=doc.name, rev=(rev or latest_revision_rev))
+    if best_events.exists():
+        # There was a period where it was possible to get more than one NewRevisionDocEvent for a revision.
+        # A future data cleanup would allow this to be simplified
+        best_event = best_events.order_by('-time').last()
+        log.assertion('doc.rev == best_event.rev')
+        doc.date = best_event.time.date()
```
rjsparks, line 1345: This will have to be re-fixed when merged to feat/tzaware

ietf/doc/utils.py
```diff
@@ -1317,3 +1318,33 @@ def fuzzy_find_documents(name, rev=None):
...
+def bibxml_for_draft(doc, rev=None):
+
+    if rev != None and rev != doc.rev:
+        # find the entry in the history
+        for h in doc.history_set.order_by("-time"):
+            if rev == h.rev:
+                doc = h
+                break
+    if rev and rev != doc.rev:
+        raise Http404("Revision not found")
```
rjsparks, line 1331: I spent a few moments pondering raising this from a utility function, and landed on it being the right thing to do still.

jennifer-richards: Your comment was separated from the line by a Codecov comment so I didn't connect it until after I'd already commented more or less the same thing. (I think it's slightly yuck, but we import django everywhere anyway and it takes so much boilerplate to convert it to a 404 in the view that I agree it's preferable to do it this way.)

ietf/doc/utils.py
jennifer-richards: Yep - I actually poked the code this replaces over on tzaware a couple hours ago. Oh well.

ietf/doc/utils.py
```diff
@@ -1317,3 +1318,33 @@ def fuzzy_find_documents(name, rev=None):
...
+    # For documents that have relevent NewRevisionDocEvents, use the date of the event.
+    # Very old documents don't have NewRevisionDocEvents - just use the document time.
+        
+    latest_revision_event = doc.latest_event(NewRevisionDocEvent, type="new_revision")
+    latest_revision_rev = latest_revision_event.rev if latest_revision_event else None
+    best_events = NewRevisionDocEvent.objects.filter(doc__name=doc.name, rev=(rev or latest_revision_rev))
+    if best_events.exists():
+        # There was a period where it was possible to get more than one NewRevisionDocEvent for a revision.
+        # A future data cleanup would allow this to be simplified
+        best_event = best_events.order_by('-time').last()
```
jennifer-richards, line 1343: `order_by('time').first()` rather than the double negative?

jennifer-richards (approved):
Looks much better than it did! A couple comments inline.

PR #4475, opened 2022-09-19
feat: use hmac, sha256, and a better secret for published nomcom hashes.

rjsparks:


ietf/nomcom/utils.py
```diff
@@ -165,6 +166,8 @@ def delete_nomcom_templates(nomcom):
     nomcom_template_path = '/nomcom/' + nomcom.group.acronym
     DBTemplate.objects.filter(path__contains=nomcom_template_path).delete()
 
+def command_line_safe_secret(secret):
+    return base64.encodebytes(secret).decode('utf-8')[:-2]
```
jennifer-richards, line 170: The `[:-2]` is chopping off the base64 padding, is that desirable? If the secret length changes, this might chomp actual bytes (or leave stray padding characters).

jennifer-richards (changes requested):
One question about base64 encoding.

ietf/nomcom/utils.py
rjsparks: it should be a -1 I suppose - the intent was to get the \n.

jennifer-richards: Makes sense - maybe use rstrip() instead?

jennifer-richards (approved):

rjsparks:
Thanks for catching the goof!

russhousley (approved):
This is a nice improvement.

PR #4488, opened 2022-09-21
feat: apis for attaching chatlogs and polls to session materials

rjsparks:
Very early look into where this is going, mostly so @alexamirante can comment on what the apis want to ingest.

Still much to be done - tests, initial views of the documents once we have them, etc.


ietf/meeting/utils.py
```diff
@@ -710,3 +710,34 @@ def handle_upload_file(file, filename, meeting, subdir, request=None, encoding=N
         subprocess.call(['unzip', filename], cwd=path)
 
     return None
+
+def new_doc_for_session(type_id, session):
+    typename = DocTypeName.objects.get(slug=type_id)
+    ota = session.official_timeslotassignment()
+    sess_time = ota and ota.timeslot.time
```
jennifer-richards, line 717: We'll need to make this TZ aware - or we could do it now like this:

```suggestion
    sess_time = ota and ota.timeslot.local_start_time()
```

Also, you'll need to handle the case where `ota is None` a few more times since it'll lead to `sess_time` being `None` in the code that follows.

jennifer-richards (commented):
Leafed through this. One detailed comment inline. API looks ok to me but it'd be nice to document the expected structure.

We're starting to see a lot of these boilerplate-ful API endpoints. Would be really nice to find a library that we could lean on instead of reinventing this each time. It'd be nice if it let us return text/plain or application/json based on the request headers, too, so that we don't hard code in that we speak JSON one way and text/plain the other.

ietf/meeting/utils.py
rjsparks: hmm - it would be very odd for ota to be None since this is used to report on a session that _happened_ already (and was set up well enough to happen in MeetEcho), but I'll look into better resiliency.

jennifer-richards: It just struck me as odd that there's the guard in this line.

ietf/doc/views_doc.py
```diff
@@ -725,6 +724,26 @@ def document_main(request, name, rev=None):
                            assignments=assignments,
                       ))
 
+    if doc.type_id == "chatlog":
```
jennifer-richards, line 727: Is a "polls" clause coming, too?

jennifer-richards (commented):
Had another look. Things look reasonable, a couple little things I noticed/wondered.

ietf/name/fixtures/names.json
```diff
@@ -3345,7 +3444,7 @@
       "has_session_materials": false,
       "is_schedulable": false,
       "material_types": "[\n    \"slides\"\n]",
-      "matman_roles": "[]",
+      "matman_roles": "[\n    \"chair\"\n]",
```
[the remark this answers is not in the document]
rjsparks: yes, and it's ok (or at least mostly unavoidable) - that reflects a change that was made in production using the admin.

ietf/doc/views_doc.py
rjsparks: yes, thanks for the nudge.

ietf/doc/migrations/0045_docstates_chatlogs_polls.py
```diff
@@ -0,0 +1,34 @@
+# Copyright The IETF Trust 2022, All Rights Reserved
+from django.db import migrations
+
+def forward(apps, shema_editor):
```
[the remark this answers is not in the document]
rjsparks: will fix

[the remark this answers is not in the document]
rjsparks: done

rjsparks:
This is now only waiting for some better FE rendering of the chatlog and polls documents

NGPixel (approved):

jennifer-richards (approved):

PR #4493, opened 2022-09-22
fix(ui): Improve the display of information on /meeting/requests

larseggert:


rjsparks:
Note the change to the PR title -  a bare 'ui: ' won't match what the CC engine is looking for.

rjsparks (approved):

PR #4509, opened 2022-09-27
feat: Set DISPLAY env, so X11 apps show on the host

larseggert:


NGPixel:
Note that `host.docker.internal` is only defined on Docker for Desktop on Mac and Windows. On Linux this entry doesn't exist and need to be added manually when creating the container using `--add-host=host.docker.internal:host-gateway`.

Would this create issues for Linux users that don't have this hostname defined in their containers?

larseggert:
I doubt it, but maybe we can add that setting to the container settings?

jennifer-richards:
Tried this on Linux - even adding the host.dockerinternal:host-gateway, I can't open X connections. Am I leaving something out?

Added `"host.docker.internal:host-gateway"` via docker-compose and get
```
dev ➜ /workspace $ ping host.docker.internal 
PING host.docker.internal (172.17.0.1) 56(84) bytes of data.
64 bytes from host.docker.internal (172.17.0.1): icmp_seq=1 ttl=64 time=0.102 ms

dev ➜ /workspace $ xterm
xterm: Xt error: Can't open display: host.docker.internal:0
```
(I opened X on the host with `xhost +` but it's been a while since I've done much X11 work so I may have missed some setup?)

jennifer-richards:
Poked around, I think the problem is that Ubuntu does not accept TCP connections for X11. I haven't found the magic to change this. TCP connections from the container work for `nc`, though.

larseggert:
You probably need to set `xhost +` or something similar

jennifer-richards:
> You probably need to set `xhost +` or something similar

It's more than that - the Ubuntu config has its xserver started with TCP listeners disabled entirely. I chased things for a while but did not find what piece was setting that command line before giving up.

The changes here were not harmful, they just don't make X connections work. I think that's enough to accept this PR. There'll just be further work needed to make X work with a system like mine.

I think it'd be fine to add the `host.docker.internal` address.

rjsparks (approved):

PR #4511, opened 2022-09-27
fix: And more `feat/htmlize` fixes

larseggert:


ietf/templates/doc/document_html.html
```diff
@@ -1,11 +1,13 @@
 {# Copyright The IETF Trust 2016, All Rights Reserved #}
 <!DOCTYPE html>
+{% load analytical %}
```
rjsparks, line 3: Good call, but what nudged you to notice to do this?

rjsparks (approved):

ietf/templates/doc/document_html.html
larseggert: I happened to look at `base.html`.

PR #4519, opened 2022-09-28
feat: include checkins at the registration desk in nomcom eligibility calculations

rjsparks:
co-author: @jennifer-richards

### DIFF
--- a/ietf/nomcom/utils.py
+++ b/ietf/nomcom/utils.py
@@ -548,10 +548,10 @@ def list_eligible_8788(date, base_qs=None):
     return remove_disqualified(three_of_five_eligible_8713(previous_five=previous_five, queryset=base_qs))
 
 def get_8989_eligibility_querysets(date, base_qs):
-    return get_threerule_eligibility_querysets(date, base_qs, three_of_five=three_of_five_eligible_8713)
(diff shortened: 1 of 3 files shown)


ietf/nomcom/utils.py
```diff
@@ -536,20 +539,26 @@ def list_eligible_8713(date, base_qs=None):
...
-    return remove_disqualified(three_of_five_eligible(previous_five=previous_five, queryset=base_qs))
+    return remove_disqualified(three_of_five_eligible_8713(previous_five=previous_five, queryset=base_qs))
 
 def get_8989_eligibility_querysets(date, base_qs):
+    return get_threerule_eligibility_querysets(date, base_qs, three_of_five=three_of_five_eligible_8989)
+
+def get_8989bis_eligibility_querysets(date, base_qs):
+    return get_threerule_eligibility_querysets(date, base_qs, three_of_five=three_of_five_eligible_8989bis)
+
+def get_threerule_eligibility_querysets(date, base_qs, three_of_five):
```
jennifer-richards, line 556: A comment (or type annotation) at least for the `three_of_five` parameter would be helpful. With all the similarly named methods and variables here it took me a while to sort out what was going on.

jennifer-richards (approved):
Looks good to me. One suggestion inline to add some comments, but not going to hold it up for that.

PR #4522, opened 2022-09-29
ci: Add dependency-review.yml

larseggert:


rjsparks:
This is something I think we had planned to talk about at the codesprint but did not manage to do so.

rjsparks:
I haven't been able to find _why_ I was waiting until we could talk wrt this one, so I'll merge it and see if it drives the thought to resurface.

rjsparks (approved):

PR #4526, opened 2022-09-30
feat: stop setting MeetingRegistration.attended on MeetEcho login.

rjsparks:
We capture attendence through MeetEcho using meeting.Attended now.


rjsparks:
cc @rpcross @alexamirante

jennifer-richards (approved):

PR #4535, opened 2022-09-30
feat: add slides / session materials links to session details modal

jennifer-richards:
Addresses some of #3725


playwright/tests/meeting/agenda.spec.js
```diff
@@ -368,12 +368,18 @@ test.describe('past - desktop', () => {
...
+      slides: {
+        decks: _.times(5, idx => ({
+          id: 100000 + idx,
+          title: faker.commerce.productName(),
+          url: `/meeting/${meetingData.meeting.number}/materials/slides-${meetingData.meeting.number}-${event.acronym}-${faker.internet.domainWord()}`,
+          ext: ['pdf', 'html', 'md', 'txt', 'pptx'][idx]
+        })),
+        actions: [{
+          label: 'Propose slides',
+          url: '/meeting/${meetingData.meeting.number}/session/${event.sessionId}/propose_slides'
```
NGPixel, line 380: Should use backticks `, not single quotes, otherwise the meeting number and session id won't be replaced in the string.

client/agenda/AgendaDetailsModal.vue
```diff
@@ -95,19 +95,29 @@ n-modal(v-model:show='modalShown')
...
-              )
-              i.bi.me-2(:class='`bi-filetype-` + slide.ext')
-              span {{slide.title}}
+          n-card(
+            :bordered='false'
+            size='small'
+            )
+            .text-center(v-if='state.isLoading')
+              n-spin(description='Loading slides...')
+            .text-center.p-3(v-else-if='!state.materials || !state.materials.slides || !state.materials.slides.decks || state.materials.slides.decks?.length < 1')
```
NGPixel, line 104: optional chaining operator `?.` not necessary here as you already validated that `.decks` exists in the previous condition

NGPixel (changes requested):

playwright/tests/meeting/agenda.spec.js
jennifer-richards: Fixed

client/agenda/AgendaDetailsModal.vue
jennifer-richards: Oops! That was left over from a check whether I could get away without the explicit check that `.decks` is present.

Ukandu147 (commented):
authorization code

jennifer-richards:
@rjsparks, this uses a hard-coded URL on the javascript side. @NGPixel is going to create a separate PR with improvements to the data store that will let us concentrate the URL-reversal logic in a single place as we discussed on the dev call.

jennifer-richards:
fyi, this is the effect of this on the sessions modal:

rjsparks (approved):

NGPixel (approved):

PR #4536, opened 2022-09-30
fix: correct meeting attendance calculations

rjsparks:


rpcross (commented):
Consider that some people register (and attend) both onsite and remote. Some 40+ examples in 114. Usually people who registered one-day onsite. In this case total unique attendees would be less than onsite attendees + remote attendees.

ietf/meeting/models.py
```diff
@@ -244,18 +244,31 @@ def get_attendance(self):
...
-                meetingregistration__meeting=self,
-                meetingregistration__attended=True,
-                meetingregistration__reg_type__contains='in_person',
-            ).distinct().count(),
-            online=Person.objects.filter(
-                meetingregistration__meeting=self,
-                meetingregistration__attended=True,
-                meetingregistration__reg_type__contains='remote',
-            ).distinct().count(),
+            onsite=attended.filter(meetingregistration__reg_type='onsite').count(),
```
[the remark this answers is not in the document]
jennifer-richards: Yep - the test in 8549cd0 demonstrates this.

jennifer-richards:
> Consider that some people register (and attend) both onsite and remote. Some 40+ examples in 114. Usually people who registered one-day onsite. In this case total unique attendees would be less than onsite attendees + remote attendees.

What is the best way to report this? Is the complexity of adding a total unique attendees in addition to onsite/remote warranted or would it be adequate to, say, count someone as onsite if they were onsite at all?

ietf/meeting/models.py
```diff
@@ -244,18 +244,31 @@ def get_attendance(self):
...
+        )
+        attended_per_meeting_attended = (
+            Q(attended__session__meeting=self)
+            # Note that we are not filtering to plenary, wg, or rg sessions
+            # as we do for nomcom eligibility - if picking up a badge (see above)
+            # is good enough, just attending e.g. a training session is also good enough
+        )
+        attended = Person.objects.filter(
+            attended_per_meetingregistration | attended_per_meeting_attended
+        ).distinct()
```
jennifer-richards, line 267: There seem to be people who have attended sessions but have not registered. E.g.,
```
(Pdb) attended.filter(meetingregistration__meeting=self).count()
1290
(Pdb) attended.exclude(meetingregistration__meeting=self).count()
11
```

jennifer-richards (approved):

rjsparks:
Input from the IETF Chair and Secretariat: They would like anyone who attended in person to only be counted as an in-person attendee even if they also registered/attended remotely.

ietf/meeting/models.py
rjsparks: Yes, this is expected - Note the comment about attending e.g. a training session. Could also have been the hackathon.

```
    attended.exclude(meetingregistration__meeting=self).values_list("attended__session__group__acronym"): '<QuerySet [('hackathon',), ('can',)]>'
```

Should we explicitly exclude hackathon? If so, what else should be excluded. Do we revert to only counting wg, rg, and plenary sessions?

jennifer-richards: Ah, I see the significance of the comment now. As it stands, I think these records will be ignored - without a meetingregistration, they're ignored by the queries for onsite and remote registrations. Maybe that's ok?

rjsparks: Yes, the people who didn't register as onsite or remote won't be counted. To reconvince myself that made it all ok, I reconsidered the case that someone registered to be remote, but only ever showed up for the hackathon and questioned whether they should be counted. I landed on "yes" because someone could register onsite, pick up their badge, and only go to the hackathon and they would get counted.

ietf/meeting/models.py
```diff
@@ -266,9 +266,13 @@ def get_attendance(self):
             attended_per_meetingregistration | attended_per_meeting_attended
         ).distinct()
 
+        onsite=set(attended.filter(meetingregistration__meeting=self, meetingregistration__reg_type='onsite'))
+        remote=set(attended.filter(meetingregistration__meeting=self, meetingregistration__reg_type='remote'))
```
jennifer-richards, line 270: Would it be better to do the exclusion in the query and just pull the counts over to Python? I'm thinking something like:
```
  onsite = attended.filter(...type='onsite').count()
  remote = attended.filter(...type='remote').exclude(...type='onsite').count()
```
(I'm not sure if that's an improvement.)

jennifer-richards (approved):
Approving - one possible change to consider and take or leave.

ietf/meeting/models.py
rjsparks: I started there and it produced surprisingly wrong results. i have a dim memory that there is a bug in the stack that causes the sql query to fail - don't remember if the issue was in Django2 or in mysql. So I just avoided it for now, and I think the result is actually a little more readable.

rjsparks: It does trade off doing things in django's memory rather than the databases, but this isn't a performance crunchpoint.

PR #4540, opened 2022-10-03
fix: Don't show ADs as part of "bethere" contraints.

larseggert:
Also tweak some minor UI stuff while I'm here.


rjsparks:
Is the point of this (not showing ADs) just to reduce repetitive content on the display at that view? I worry that this will trigger bug reports of "but the AD _has_ to be there"...

larseggert:
Yes. I also think that this view is internal to the secretariat and the IESG, who will hopefully understand that.

rjsparks (approved):

PR #4551, opened 2022-10-07
feat: include proposed WGs on meeting requests page

jennifer-richards:
As requested by @flynnliz

### DIFF
--- a/ietf/templates/meeting/requests.html
+++ b/ietf/templates/meeting/requests.html
@@ -68,8 +68,8 @@
                                     <i class="bi bi-people-fill"></i> {{ session.attendees }}
                                 </div>
                             {% endif %}
-                            {% if session.group.is_bof %}
-                                <span class="badge bg-success float-end">BOF</span>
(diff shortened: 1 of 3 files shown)


ietf/templates/meeting/requests.html
```diff
@@ -68,8 +68,8 @@ <h2 class="mt-5" id="{% firstof area.grouper.acronym "other-groups" %}">
                                     <i class="bi bi-people-fill"></i> {{ session.attendees }}
                                 </div>
                             {% endif %}
-                            {% if session.group.is_bof %}
-                                <span class="badge bg-success float-end">BOF</span>
+                            {% if session.group.state.slug != "active" %}
+                                <br><span class="badge bg-success">{{ session.group.state.name }}</span>
```
rjsparks, line 72: Slightly concerned about bofs and proposed being the same color, but we can start here and improve later.

rjsparks (approved):

PR #4553, opened 2022-10-07
fix: work around dataTransfer.getData limitation in agenda editor

jennifer-richards:
The agenda editor was broken as of Chrome 106 without this patch. It's a bit of a hack, but I believe it respects the letter of the relevant specs.

### DIFF
--- a/ietf/static/js/edit-meeting-schedule.js
+++ b/ietf/static/js/edit-meeting-schedule.js
@@ -412,7 +412,6 @@ $(function () {
             let dropElement = jQuery(this);
 
             const sessionElement = getDraggedSession(event);
-            console.log(`Hiya: {sessionElement}`);
             if (!sessionElement) {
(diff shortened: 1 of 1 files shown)


ietf/static/js/edit-meeting-schedule.js
```diff
@@ -402,6 +412,7 @@ $(function () {
             let dropElement = jQuery(this);
 
             const sessionElement = getDraggedSession(event);
+            console.log(`Hiya: {sessionElement}`);
```
rjsparks, line 415: Leftover debug?

rjsparks (changes requested):

ietf/static/js/edit-meeting-schedule.js
jennifer-richards: Ugh, yes. Removed.

rjsparks (approved):

PR #4564, opened 2022-10-12
fix: revert photo-on-menubar enhancement

rjsparks:
Turns out the state of our thumbs is poor enough to make this a non-starter for now.

We need to revisit whether to clean up our image store or to shift to a third-party (gravatar, or clouflare's self-serve image manager for example).

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -227,13 +227,6 @@ th {
     }
 }
 
-// Style the navbar user photo
-.nav-link .user-photo {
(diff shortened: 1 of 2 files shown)


jennifer-richards (approved):
Leaves behind a now-unused [.nav-link .user-photo style](https://github.com/ietf-tools/datatracker/blob/08dfb82fa4102e996dac46bf8c112a636090d461/ietf/static/css/ietf.scss#L231) which should perhaps be reverted also. Up to you whether to bother.

PR #4587, opened 2022-10-14
fix: Fix incorrect "GDPR" features/terminology. Fixes #4521.

### LINKED ISSUE #4521: Fix incorrect "GDPR" features/terminology
### Describe the issue

Datatracker has features/terminology that claim to relate to GDPR but our legal advice is that these do not make any sense at all and should be altered.  That means the following:

1. Remove this page entirely https://datatracker.ietf.org/help/personal-information

3. The "Account Info" page should change as follows:
- The title of the page should change to "Your account"
- The subtitle "The information you provide below is used to generate your public datatracker profile page" removed.
- The bang inside a circle icon next to some fields needs to be removed from all fields
- The fields reordered as follows
   - Fields: User name, password, noncom eligible,
   - Then a separator subtitle "Published information" and explanatory text "All of the data below may be openly published in Datatracker, Meetecho and other IT systems".  If you do not want these data published then leave these fields blank."
   - Fields: Full name (Unicode), Plain name correction (Unicode), Full name (ASCII), Pronouns, Biography, External resources
   - Add the following info text under External resources (similar to the other info text): "
   - Then a separator subtitle "Email addresses" and explanatory text "You can optionally link more email addresses to your account, which will then ensure that all I-Ds you have published using different email addresses are all shown against your account."
   - Fields: Email addresses
   - Remove the tick box declaration entirely
   - Change the "Submit" button to say "Apply changes"
   - Remove the key of the bang in a circle and "Personal information requiring consent" and remove the two paragraphs that begin "Personal information"

### Code of Conduct

rjsparks:


JayDaley (approved):
Seems right to me though my DT knowledge is limited

jennifer-richards (approved):

PR #4595, opened 2022-10-17
fix: regenerate correct shepherd template. Fixes #4558

### LINKED ISSUE #4558: Reset to Template Text button on shepherd writeup form generates wrong template type
At doc/draft-ietf-stir-identity-header-errors-handling/edit/shepherdwriteup/, before a writeup was generated, the form presents the template appropriate for a wg document. Pressing the Reset to Template Text button replaces it with the template for an individual submission document.

rjsparks:
Finishes what #4316 should have done.

### DIFF
--- a/ietf/doc/views_draft.py
+++ b/ietf/doc/views_draft.py
@@ -989,14 +989,16 @@ def edit_shepherd_writeup(request, name):
                 return redirect("ietf.doc.views_doc.document_main", name=doc.name)
 
         elif "reset_text" in request.POST:
+            if not doc.group.type.slug or doc.group.type.slug != "wg":
+                generate_type = "individ"
+            else:
+                generate_type = "group"           
             init = {
                 "content": render_to_string(
                     "doc/shepherd_writeup.txt",
                     dict(
                         doc=doc,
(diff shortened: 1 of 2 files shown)


ietf/doc/views_draft.py
```diff
@@ -995,7 +995,7 @@ def edit_shepherd_writeup(request, name):
                     dict(
                         doc=doc,
                         type="individ"
-                        if not doc.group.type.slug or doc.group.type.slug != "ietf"
+                        if not doc.group.type.slug or doc.group.type.slug != "wg"
                         else "group",
                         stream=doc.stream.slug,
```
jennifer-richards, line 1000: "Just" a style thing, but I had to stare at lines 997-999 for way too long to decipher the syntax. If it won't fit on one line, I think it either should be parenthesized as a hint or computed into a temporary variable to make the logic readable.

jennifer-richards (changes requested):
I think the change is fine, but needs a syntax tweak for readability badly enough to change it while you're in the area. Comment is inline.

jennifer-richards (approved):

PR #4605, opened 2022-10-19
fix: Fix table sort functionality for tables with `colspan` attributes

### LINKED ISSUE #4555: nomcom private page tables have column-sort markers that do nothing
### Describe the issue

the "nominees" page has column-sort markers (up/down) for the first table, which is summary counts, and for the second table, which are specific entries.  The first table sorts by clicking on those markers. The second table does nothing.  

### Code of Conduct

larseggert:
Fixes #4555

### DIFF
--- a/ietf/static/js/list.js
+++ b/ietf/static/js/list.js
@@ -24,11 +24,9 @@ function replace_with_internal(table, internal_table, i) {
 }
 
 function field_magic(i, e, fields) {
-    if ($(e)
-        .attr("colspan") === undefined &&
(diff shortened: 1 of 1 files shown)


ietf/static/js/list.js
```diff
@@ -62,12 +60,21 @@ $(document)
                 // get field classes from first thead row
                 var fields = $(header_row)
```
rjsparks, line 61: I'm going to approve this, but this bit of new code is particularly hard to understand. I wonder if it could be written in a way that would promote better maintenance?

rjsparks (approved):

PR #4608, opened 2022-10-19
fix: Serve documents and photos from dev instance

### LINKED ISSUE #4507: TRACKER_BASE_URL should be overriden to the dev instance when in dev mode
Right now, links produced for meeting materials (or any link constructed with settings.TRACKER_BASE_URL) by a dev container point back to the production datatracker. They should instead point to the local instance.

---

larseggert:
Fixes #4507


rjsparks:
This isn't quite as simple as the issue title suggests. The links from the materials page generated with this proposed fix are not the right ones - they need to be to what would serve the content locally, not back to the datatracker page - it may have looked like it worked since by coincidence the links that were created didn't 404 (only because the structure of links under the two places is the same), but the content being returned is not what should be returned.

Further, we need to test that a reasonable thing happens if the docker container is exposing the running server somewhere other than port 8000 to the host.

larseggert:
Um. I'm not sure I grok all that. Maybe I just close the PR and someone else can pick up the ticker?

rjsparks:
wfm. Sorry my description wasn't clear.

PR #4609, opened 2022-10-19
fix: Don't say that IESG processing has started

### LINKED ISSUE #4286: When an AD puts a document in AD is watching, email emitted says the IESG is starting processing
(Reported by Ben Campbell)

That confuses chairs. The message should be more specific in this case.

cc @mskucherawy

larseggert:
Fixes #4286

### DIFF
--- a/ietf/doc/views_draft.py
+++ b/ietf/doc/views_draft.py
@@ -713,7 +713,7 @@ def edit_info(request, name):
                 e.by = by
                 e.doc = doc
                 e.rev = doc.rev
-                e.desc = "Document is now in state <b>%s</b>" % doc.get_state("draft-iesg").name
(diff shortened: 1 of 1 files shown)


ietf/doc/views_draft.py
```diff
@@ -592,7 +592,7 @@ def to_iesg(request,name):
                 e.by = by
                 e.doc = doc
                 e.rev = doc.rev
-                e.desc = "IESG process started in state <b>%s</b>" % target_state['iesg'].name
+                e.desc = "Document is now in state <b>%s</b>" % target_state['iesg'].name
```
rjsparks, line 595:
```suggestion
                e.desc = "Document is now in IESG state <b>%s</b>" % target_state['iesg'].name
```

rjsparks (approved):

PR #4615, opened 2022-10-20
fix: use UTC for materials/proceedings cutoff checks

### LINKED ISSUE #4436: enforce "important-dates" deadlines in UTC
### Describe the issue

Many items on the list at https://datatracker.ietf.org/meeting/important-dates/ indicate that their deadline is in UTC. While reworking the timezone handling, verify whether these are implemented to match and fix if not.

### Code of Conduct

jennifer-richards:
Fixes #4436

### DIFF
--- a/ietf/meeting/models.py
+++ b/ietf/meeting/models.py
@@ -1167,7 +1167,7 @@ class Session(models.Model):
         return can_manage_materials(user,self.group)
 
     def is_material_submission_cutoff(self):
-        return date_today(self.meeting.tz()) > self.meeting.get_submission_correction_date()
+        return date_today(datetime.timezone.utc) > self.meeting.get_submission_correction_date()
     
     def joint_with_groups_acronyms(self):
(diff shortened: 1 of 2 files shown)


ietf/meeting/models.py
```diff
@@ -1167,7 +1167,7 @@ def can_manage_materials(self, user):
         return can_manage_materials(user,self.group)
 
     def is_material_submission_cutoff(self):
-        return date_today(self.meeting.tz()) > self.meeting.get_submission_correction_date()
+        return date_today(datetime.timezone.utc) > self.meeting.get_submission_correction_date()
```
rjsparks, line 1170: good find. Makes no sense that this was tied to meeting timezone.

rjsparks (approved):

ietf/meeting/models.py
jennifer-richards: Thanks (though I'm probably responsible for tying it there to begin with....)

PR #4621, opened 2022-10-20
fix: stable and correct ordering of agenda items

rjsparks:


jennifer-richards (approved):

ietf/meeting/models.py
```diff
@@ -1105,8 +1105,18 @@ def all_meeting_sessions_for_group(self):
         from ietf.meeting.utils import add_event_info_to_session_qs
         if self.group.features.has_meetings:
             if not hasattr(self, "_all_meeting_sessions_for_group_cache"):
-                sessions = [s for s in add_event_info_to_session_qs(self.meeting.session_set.filter(group=self.group,type=self.type)) if s.official_timeslotassignment()]
-                self._all_meeting_sessions_for_group_cache = sorted(sessions, key = lambda x: x.official_timeslotassignment().timeslot.time)
+                sessions = [s for s in add_event_info_to_session_qs(self.meeting.session_set.filter(group=self.group)) if s.official_timeslotassignment()]
+                for s in sessions:
+                    s.ota = s.official_timeslotassignment()
```
jennifer-richards, line 1110: After approving, noticed that this means `official_timeslotassignment()` is called twice for each session. Is that cheap enough (or cached effectively by the ORM) not to worry about it? There are only a couple dozen sessions filtered out in the list comprehension so it'd probably be preferable to filter the list after caching `s.ota`.

PR #4622, opened 2022-10-21
fix: reduce agenda data calculation time

rjsparks:
This intentionally leaves a commented out debug statement. Against main without #4621, it will report differences with the secretariat group. It will report equal once #4621 is in place.

I had done the work to precalculate the order in the view, but while investigating what led to #4621, discovered that the calculations had _already been made and attached to the objects_.

It's not clear why it wasn't being used.

### DIFF
--- a/ietf/meeting/helpers.py
+++ b/ietf/meeting/helpers.py
@@ -95,6 +95,7 @@ def preprocess_assignments_for_agenda(assignments_queryset, meeting, extra_prefe
       a.session.historic_parent
       a.session.rescheduled_to (if rescheduled)
       a.session.prefetched_active_materials
+      a.session.order_number
     """
     assignments_queryset = assignments_queryset.prefetch_related(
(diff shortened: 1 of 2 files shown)


jennifer-richards (approved):
This looks fine.

It is possible for a session to have `order_number = None` if `session.group` is `None`. I don't think those should be showing up in the schedules, but it might cause bad links if it does. Could fall back to `0` if it comes up instead of propagating the `None`.

Other thought - it'd be good to update the comment on `preprocess_assignments_for_agenda()` to document that it's setting the `order_number`.

rjsparks:
> It is possible for a session to have `order_number = None` if `session.group` is `None`. I don't think those should be showing up in the schedules, but it might cause bad links if it does. Could fall back to `0` if it comes up instead of propagating the `None`.

Session is defined such that Django will not allow session.group to be None.

PR #4650, opened 2022-10-26
fix: Handle RFC names that are not all lowercase or have spaces

### LINKED ISSUE #4576: htmlizer entry point does wrong thing with cased inputs
From @agmalis

> I note that the URL
> 
> https://datatracker.ietf.org/doc/html/rfc7209
> 
> correctly displays RFC 7209.
> 
> However, the URL
> 
> https://datatracker.ietf.org/doc/html/RFC7209
> 
> actually displays draft-ietf-l2vpn-evpn-req-07.txt .
> 
> Cheers,
> Andy

larseggert:
Fixes #4576


ietf/doc/utils.py
```diff
@@ -1293,6 +1293,7 @@ def fuzzy_find_documents(name, rev=None):
     document.
     """
     # Handle special case name formats
+    name = re.sub(r"\s+", "", name.lower())
```
rjsparks, line 1296: Can we restrict this to only happen if name.lower() startswith `rfc`?

ietf/doc/urls.py
```diff
@@ -82,14 +82,14 @@
 
 # End of block that should be an idealized docs.ietf.org service instead
 
-    url(r'^html/(?P<name>[Rr][Ff][Cc] [0-9]+?)(\.txt|\.html)?/?$', views_doc.document_html),
+    url(r'^html/(?P<name>[Rr][Ff][Cc]\s*[0-9]+?)(\.txt|\.html)?/?$', views_doc.document_html),
```
rjsparks, line 85: What is the problem that led to wanting to do this? It changes the contract for what we'll provide, and we would be expected to keep this working in perpetuity. 
If it's a linkifier figuring out what to do, why not make it the linkifiers problem to normalize the produced URL instead?

rjsparks:
> Fixes https://github.com/ietf-tools/datatracker/issues/4576
This is a surprising fix for that bug. I think it is probably just avoiding a bug we have elsewhere in the code that should be addressed. Did you determine _why_ this ended up on the last draft instead of the RFC? I would have expected there to be code that would be touching that decision directly.

larseggert:
The issue is that the `if re.match("^[Rr][Ff][Cc] [0-9]+$",name):` case in the original code doesn't hit, because of the space in the pattern. And even if it did, it would rewrite `name` incorrectly, because the replacement assumes a space exists.

rjsparks (approved):

PR #4651, opened 2022-10-26
chore: Fix spelling/wording via `codespell`

larseggert:


ietf/mailinglists/migrations/0001_initial.py
```diff
@@ -40,15 +40,15 @@ class Migration(migrations.Migration):
             },
         ),
         migrations.CreateModel(
-            name='Whitelisted',
+            name='Allowlisted',
```
rjsparks, line 43: Making this change in the 0001_initial migration will have no effect on the production database. This file should be left alone to allow the migration system to work properly.

What needs to happen instead is a change to the model (and it's associated admin and resource files) and a _new_ migration that will make the change to the data.

rjsparks (changes requested):
This really should be split into two PRs.

It is not a chore just fixing spelling as advertised - it is a change to the models and the api.

Please create one PR that changes the mailinglists models (and the related changes that requires, including the new migration it needs). This one should be labeled a feat: (Even though it's a breaking change for anyone depending on the v1 api, we've already decided we won't treat that as something that bumps our major version number).

Please use a separate PR for the changes that do not affect the database and api. This one can be a chore.

Consider breaking the changes that affect settings, code and text shown to users apart from changes that only affect comments if that's not too hard to do. It would make looking for anything outside the repository that needs to be changed to follow along simpler.

larseggert:
Closing in favor of #4794 and #4797.

PR #4654, opened 2022-10-26
chore: update create_dummy_meeting to specify session purposes

jennifer-richards:
Also broken on the main branch, but fixing on feat/tzaware because no one is likely to run the command on main until after the tzaware switch.


rjsparks (approved):

rjsparks:
Not a priority, but this looks like a place where using factories would be _way_ better than bare ORM create calls.
Could ditch the manual effort of looking up/encoding group pks for example...

jennifer-richards:
Created #4657 to capture the factory refactoring

PR #4663, opened 2022-10-27
feat(agenda): fallback link to plaintext version

### LINKED ISSUE #4655: Add link to txt agenda to top menu
### Describe the issue

As the current agenda does not work anymore for all browsers, it would be nice to have link to the txt version of the agenda so I could use that to see what is going on in the IETF.

I.e. add link to https://datatracker.ietf.org/meeting/agenda.txt so it can be found from the top menu.

### Code of Conduct

NGPixel:
Adds fallback links to loading screen and error screen:
image

image

Fixes #4655

### DIFF
--- a/client/App.vue
+++ b/client/App.vue
@@ -4,6 +4,8 @@ n-theme
     .app-error(v-if='siteStore.criticalError')
       i.bi.bi-x-octagon-fill.me-2
       span {{siteStore.criticalError}}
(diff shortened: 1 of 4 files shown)


rjsparks (approved):

larseggert:
Nice fix. I just wish we wouldn't continue to build a ton of custom CSS instead of using b5. It will be a pain to move to newer bs versions.

PR #4676, opened 2022-10-28
feat: allow secretariat to edit/cancel sreqs when the tool is locked

### LINKED ISSUE #4272: Secretariat could not cancel TDD
The UI didn't give Liz the ability to cancel TDD at 114.
I added a SchedulingEvent manually through the admin.

jennifer-richards:
Related to #4272

### DIFF
--- a/ietf/secr/sreq/views.py
+++ b/ietf/secr/sreq/views.py
@@ -849,7 +849,6 @@ def view(request, acronym, num = None):
     session = get_initial_session(sessions)
 
     return render(request, 'sreq/view.html', {
-        'is_locked': is_locked,
         'can_edit': (not is_locked) or has_role(request.user, 'Secretariat'),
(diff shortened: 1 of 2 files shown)


ietf/secr/sreq/views.py
```diff
@@ -850,6 +850,8 @@ def view(request, acronym, num = None):
 
     return render(request, 'sreq/view.html', {
         'is_locked': is_locked,
```
rjsparks, line 852: The template doesn't use this now, I think? If I'm right, we shouldn't pass it.

rjsparks (changes requested):

ietf/secr/sreq/views.py
jennifer-richards: Good catch, I remembered it appearing in other places but that must have been other templates. Will remove.

rjsparks (approved):

Issue #3591, opened 2022-03-08
Automatically return to correct schedule in schedule editor


jennifer-richards:
The schedule editor links to several related views - edit session details, edit timeslots, etc. These all need a convenient way to return to the editor.

Currently, generic links to the agenda list for the current meeting are provided. This requires an extra step to select the schedule that was being edited (and often additional clicks after that, see #3590).

It'd be more usable if the user could be directly returned to where they left the schedule editor.

Edit: Some already do (the edit session details does this, I believe by passing a hidden parameter in the template)

PR #4682, opened 2022-11-01
feat: add 'cancel session' button to meeting schedule editor

### LINKED ISSUE #4272: Secretariat could not cancel TDD

jennifer-richards:
Fixes #4272, correctly this time I hope.

Also begins to address #3591. To do this, passes the pk of the schedule being edited as a query parameter in the URL when visiting the cancel_session view.

### DIFF
--- a/ietf/meeting/forms.py
+++ b/ietf/meeting/forms.py
@@ -773,6 +773,13 @@ class SessionEditForm(SessionDetailsForm):
         super().__init__(instance=instance, group=instance.group, *args, **kwargs)
 
 
+class SessionCancelForm(forms.Form):
+    confirmed = forms.BooleanField(
+        label='Cancel session?',
+        help_text='Confirm that you want to cancel this session.',
+    )
+
+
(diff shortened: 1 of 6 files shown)


jennifer-richards:
The two failing tests are unrelated to this PR - they slipped into feat/tzaware (oops). I will fix those with another PR.

rjsparks (approved):
What does the workflow look like if someone decides they need to un-cancel a session?
(Asking for admin-level help is ok for this edge case)

jennifer-richards:
> What does the workflow look like if someone decides they need to un-cancel a session? (Asking for admin-level help is ok for this edge case)

Requires the admin to fix it. I *think* that's how it was before.

I contemplated adding an "uncancel" option but a) did not see an obvious example of how to do this in the existing workflows and b) it will require changes to the "readonly" session handling. At the moment, once a session is canceled it becomes untouchable on the schedule editor.

PR #4693, opened 2022-11-05
Add note to devguide re: Snap incompatibility

aph3rson:
Plugin doesn't work w/ Snap installation, add note.
See note [here](https://code.visualstudio.com/docs/devcontainers/containers#_system-requirements:~:text=snap%20package%20is%20not%20supported).


aph3rson:
needs a rebase, my fork is missing the right branches. :facepalm:

PR #4696, opened 2022-11-05
feat: Add requested link to the light blue box on group meetings page

### LINKED ISSUE #4689: link directly to page of proceedings
### Description

"This page shows meetings within the last four years. For earlier meetings, please see the proceedings."

It would be great if this could actually link to the spot on ietf.org, where the proceedings *are*
I can't find them on the new site in the menus.

### Code of Conduct

russhousley:
Add requested link to the light blue box.

### DIFF
--- a/ietf/templates/group/meetings.html
+++ b/ietf/templates/group/meetings.html
@@ -1,5 +1,5 @@
 {% extends "group/group_base.html" %}
-{# Copyright The IETF Trust 2015, All Rights Reserved #}
+{# Copyright The IETF Trust 2015-2022, All Rights Reserved #}
 {% load origin %}
 {% block title %}
     Meetings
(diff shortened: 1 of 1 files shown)


ietf/templates/group/meetings.html
```diff
@@ -85,6 +85,7 @@ <h2 class="mt-3" id="pastmeets">Past Meetings</h2>
         </table>
     {% endif %}
     <p class="alert alert-info my-3">
-        This page shows meetings within the last four years. For earlier meetings, please see the proceedings.
+        This page shows meetings within the last four years. For earlier meetings, please see the
+        <a href="https://www.ietf.org/how/meetings/past/">proceedings</a>.
```
larseggert, line 89: You could use `settings.IETF_HOST_URL` instead of hardcoding, but we're not very consistent with it.

russhousley: I looked at the pulldown for "Meeting proceedings", and it is hard coded there, so I did the same.

rjsparks: we should be more consistent with it, but lets get that with a concentrated pass through the code.

rjsparks (approved):

PR #4698, opened 2022-11-05
fix: Eliminate use of django.utils.datetime_safe

IurmanJ:
<s>Use standard library instead. Hard to test since the code part using `datetime_safe` does not seem to be called from anywhere. Another alternative to this patch attempt would be to simply remove that code part.</s>

Instead of replacing `datetime_safe` by the standard api, just remove that part of the code (because it is unused).

Fixes #4513

### DIFF
--- a/ietf/meeting/models.py
+++ b/ietf/meeting/models.py
@@ -60,16 +60,6 @@ for name in pytz.common_timezones:
 timezones.sort()
 
 
-# this is used in models to format dates, as the built-in json serializer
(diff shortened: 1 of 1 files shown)


larseggert:
I'd be fine with removing the function. @rjsparks?

jennifer-richards:
Seems fine to me, I don't see any use of it.

IurmanJ:
Rebased commit to remove the mentioned code instead of replacing it

rjsparks (approved):

PR #4700, opened 2022-11-05
fix: Don't send I-D expiration warning if it will happen within 12 hours.

### LINKED ISSUE #4635: draft expiration warning with zero time left is not useful.
cc @russhousley, @bacampbe ,@mskucherawy

Recently draft-ietf-stir-rfc4916-update expired. The chairs, ad, and authors were sent two emails, nearly simultaneously. The second noted that the draft had expired. The first said:
```
The following draft will expire soon:

Name:     draft-ietf-stir-rfc4916-update
Title:    Connected Identity for STIR
State:    I-D Exists
Expires:  2022-10-23 (in 0 minutes)
```

That was a waste of resources at best, and a source of confusion at worst.

The expiration warning management command should instead have a configured buffer for which to not warn if the clock is close to running out. I suggest not sending the warning message if the remaining time is less than 12 hours.

russhousley:
Don't send I-D expiration warning if it will happen within 12 hours.


ietf/doc/expire.py
```diff
@@ -85,6 +85,12 @@ def send_expire_warning_for_draft(doc):
         return # don't warn about dead or inactive documents
 
     expiration = doc.expires.date()
+    now_plus_12hours = datetime.datetime.now() + datetime.timedelta(hours=12)
```
jennifer-richards, line 88: Should be `timezone.now()` instead of `datetime.datetime.now()`

jennifer-richards (changes requested):
This is targeting main - would be nice to aim it at feat/tzaware instead. For either branch, replacing `datetime.datetime.now()` with `timezone.now()` as noted inline is preferable.

rjsparks:
I'll retarget this and make the changes to use timezone.

jennifer-richards (approved):

PR #4704, opened 2022-11-05
Fix 3760 more visible subject email on review assignment

evyncke:
Fix issue 3760 by a new subject line, e.g.,
`Subject: For Benno Overeinder, Intdir Last Call review assigned: draft-ietf-add-dnr`

Also, send the email to all secretaries even to the requestor (so they can 'see' what is automatically generated).


client/agenda/AgendaMobileBar.vue
```diff
@@ -31,7 +31,7 @@ import {
 } from 'naive-ui'
 
 import { useAgendaStore } from './store'
-import { useSiteStore } from '../shared/store';
+import { useSiteStore } from '../shared/store'
```
larseggert, line 34: Superfluous change?

rjsparks: There are many files in here that shouldn't be - I suspect Eric was based in the wrong place. I'll work with him to pull out the parts of this that were meant to change.

evyncke: Indeed, I tried to start from feat/tzaware rather than main but I failed.

evyncke: Will redo the changes in a fresh PR

evyncke:
This is the plan for this week, so good move

Issue #4711, opened 2022-11-05
Top right document search does not allow opening document in new tabs.


kivinen:
### Describe the issue

When I do search on the top right Document search text box, I quite often want to open multiple of those quick links as separate tabs (i.e., I might not be sure which of the drafts is the one I want to look at). If I try to use browsers menu (right mouse button) and select the  "Open in new tab" or "Open in new window" I cannot because the datatracker immediately loads the page when I press right mouse button on the list item, i.e., it binds both left and right mouse buttons to open the page.

Also using Control-left-mouse-click (open in new tab), Control-shift-left-mouse-click (oipen in new tab and select that tab) works as datatracker always opens the draft in current tab/window.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #4716, opened 2022-11-06
Search UI

kivinen:
Implements #4711 i.e., allows using control-click or context menu to open items in the top right search list to open multiple documents in different tabs.

### DIFF
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -3466,8 +3466,6 @@ def upcoming(request):
 
     # Get ietf meetings starting 7 days ago, and interim meetings starting today
     ietf_meetings = Meeting.objects.filter(type_id='ietf', date__gte=today-datetime.timedelta(days=7))
-    for m in ietf_meetings:
-        m.end = m.date + datetime.timedelta(days=m.days-1)  # subtract 1 to avoid counting an extra day
 
     interim_sessions = add_event_info_to_session_qs(
         Session.objects.filter(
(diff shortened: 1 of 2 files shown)


larseggert:
There are some meeting-related changes in this PR that seem unrelated?

kivinen:
Lars Eggert writes:
> There are some meeting-related changes in this PR that seem unrelated?

Most likely they are from my previous change
https://github.com/ietf-tools/datatracker/pull/4701 that was already
merged in. I did create new branch for these changes, but guess I
should have done something else also to get rid of those changes.
-- 
***@***.***

larseggert:
You need to create your new branch off tzaware, not of another branch that has changes.

rjsparks:
At this point, it would be better to recreate this PR from current main.
I took a look at detangling this myself, and I could do so in a pinch, but I think it will be easier for Tero when he has a moment.

kivinen:
Robert Sparks writes:
> At this point, it would be better to recreate this PR from current main.
> I took a look at detangling this myself, and I could do so in a pinch, but I
> think it will be easier for Tero when he has a moment.

If you tell me how to do that I can do it, but every time I try to do
something with git/github it takes about 2 hours, and couple of
times, when I rm -rf everything and start over... 
-- 
***@***.***

rjsparks:
This is moved to #4747.

rjsparks:
(I tried to do a soft-reset juggle, but ended up with exactly the opposite of what I wanted. I built the correct thing on a fresh checkout of main instead.)

PR #4740, opened 2022-11-09
test: validate HTML in own test case and report all results

jennifer-richards:
This fixes two shortcomings with the HTML validation:
1. Validation results are now reported in a single test case at the end instead of being tacked on to often unrelated tests
2. All validation results are reported - previously, the last batch (up to 30 of each `kind`) were silently dropped during the test runner teardown

Using `memory_profiler`, it appears that these changes add ~50 MB to the ~600 MB peak ram used by running the full test suite. The test run is a bit faster than before (consistent with old code comments estimating about 10% speed overhead).

A side effect of point 2 was that, when running only a few tests, HTML validations were run but their results discarded. That was the immediate impetus for this pull request.

### DIFF
--- a/ietf/utils/test_runner.py
+++ b/ietf/utils/test_runner.py
@@ -325,10 +325,6 @@ class ValidatingTemplate(Template):
         settings.validate_html.batches[kind].append(
             (self.origin.name, content, fingerprint)
         )
-        # FWIW, a batch size of 30 seems to result in less than 10% runtime overhead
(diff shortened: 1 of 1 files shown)


jennifer-richards:
Note that the failing test is expected. That's fixed by the next pull request.

larseggert (approved):

rjsparks (approved):

PR #4744, opened 2022-11-10
feat: move to postgresql

rjsparks:


jennifer-richards (approved):

docker/scripts/db-include-fix.py
```diff
@@ -0,0 +1,28 @@
...
+        'ENGINE': 'django.db.backends.mysql',
+        'USER': 'django',
+        'PASSWORD': 'RkTkDPFnKpko',
+        'OPTIONS': {
+            'sql_mode': 'STRICT_TRANS_TABLES',
+            'init_command': 'SET storage_engine=InnoDB; SET names "utf8"',
+        },
+    },
+}
+newcontent = content.replace("""DATABASES = {
```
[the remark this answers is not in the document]
rjsparks: that is bizarre - how did this even seem to work? Maybe it got clobbered between test and commit. Fix inbound.

docker/scripts/db-include-fix.py
```diff
@@ -0,0 +1,28 @@
+with open('/workspace/ietf/settings_local.py','r') as contentfile:
+    content = contentfile.readlines()
+newcontent = content.replace("""DATABASES = {
```
jennifer-richards, line 3:
```suggestion
from pathlib import Path
content = Path('/workspace/ietf/settings_local.py').read_text()
newcontent = content.replace(
    """DATABASES = {
```

docker/scripts/db-include-fix.py
```diff
@@ -0,0 +1,28 @@
...
+        'ENGINE': 'django.db.backends.mysql',
+        'USER': 'django',
+        'PASSWORD': 'RkTkDPFnKpko',
+        'OPTIONS': {
+            'sql_mode': 'STRICT_TRANS_TABLES',
+            'init_command': 'SET storage_engine=InnoDB; SET names "utf8"',
+        },
+    },
+}""", "from ietf.settings_mysqldb import DATABASES")
+newcontent = content.replace("""DATABASES = {
```
jennifer-richards, line 17:
```suggestion
}""",
    "from ietf.settings_mysqldb import DATABASES",
).replace(
    """DATABASES = {
```

docker/scripts/db-include-fix.py
```diff
@@ -0,0 +1,28 @@
...
+    'default': {
+        'HOST': 'pgdb',
+        'PORT': 5432,
+        'NAME': 'ietf',
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'USER': 'django',
+        'PASSWORD': 'RkTkDPFnKpko',
+    },
+}""", "from ietf.settings_postgresqldb import DATABASES")
+with open('/workspace/ietf/settings_local.py','w') as replacementfile:
```
jennifer-richards, line 27:
```suggestion
}""",
    "from ietf.settings_postgresqldb import DATABASES",
)
with Path('/workspace/ietf/settings_local.py').open('w') as replacementfile:
```

jennifer-richards (changes requested):
Suggested a tweak to the `db-include-fix.py` that gets it working on my system. Might also be worth adding some debug output to indicate whether that script makes changes or not - it's a little fragile. OTOH may not have a long enough lifespan to warrant it.

The pause migration is still not working as intended, haven't sorted out why.

ietf/utils/migrations/0004_pause_to_change_database_engines.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.28 on 2022-11-10 09:27
+
+from django.conf import settings
+from django.db import migrations
+
+def forward(apps, schema_editor):
+    assert(settings.DATABASES['default']['ENGINE'] == 'django.db.backends.postgresql_psycopg2', 'Please change to use postgres before continuing')
```
jennifer-richards, line 7:
```suggestion
    assert settings.DATABASES['default']['ENGINE'] == 'django.db.backends.postgresql_psycopg2', 'Please change to use postgres before continuing'
```
Brace yourself for a groan - the `assert` keyword cannot be used with parentheses around its arguments. It's not a method call, so the parentheses mean you're asserting the tuple `(False, 'Please change...')` which is truthy regardless of whether its first element is `True` or `False`.

ietf/utils/migrations/0004_pause_to_change_database_engines.py
```diff
@@ -0,0 +1,20 @@
...
+
+def reverse(apps, schema_editor):
+    assert(settings.DATABASES['default']['ENGINE'] == 'django.db.backends.mysql', 'Please change to use mariadb before continuing')
```
jennifer-richards, line 10:
```suggestion
    assert settings.DATABASES['default']['ENGINE'] == 'django.db.backends.mysql', 'Please change to use mariadb before continuing'
```

jennifer-richards (changes requested):

PR #4759, opened 2022-11-15
fix: use tzaware today while expiring drafts

rjsparks:


rjsparks:
This needs #4755's fixes or in_draft_expire_freeze fails (at get_second_cut_off).

jennifer-richards (approved):

Issue #4758, opened 2022-11-15
Unexpected results from model_to_timeline_data() after midnight UTC


jennifer-richards:
### Describe the issue

A couple of tests failed when run in the window where the UTC date has rolled over but PST has not yet done so:
```
======================================================================
FAIL: test_personal_chart (ietf.doc.tests.ChartTests.test_personal_chart)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/datatracker/datatracker/ietf/doc/tests.py", line 2510, in test_personal_chart
    self.assertEqual(len(d), 1)
AssertionError: 2 != 1

======================================================================
FAIL: test_search_chart_data (ietf.doc.tests.ChartTests.test_search_chart_data)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/datatracker/datatracker/ietf/doc/tests.py", line 2478, in test_search_chart_data
    self.assertEqual(len(d), 1)
AssertionError: 2 != 1
```
These are failing because the `model_to_timeline_data()` method is returning two columns when the test expects only one. Printing the value returned during a run of the `test_personal_chart` test gives
```
    d: '[[1668470400000, 1], [1668384000000, 0]]'
```
I think this means the `model_to_timeline_data()` and the tests disagree about what time zone is used to define days in the time series. Need to sort out what it should be doing and update the code and / or tests to match.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #4763, opened 2022-11-15
fix: use UTC for model_to_timeline_data()

jennifer-richards:
This changes the method to operate in UTC instead of server time. The existing code was mixing UTC and server time, though, giving timestamps purporting to be UTC but computed from the date in the server timezone. To the extent that it matters, I think this change is preferable to restoring the original behavior.


jennifer-richards:
I expect this will fix the failure in #4758 but have not yet verified.

rjsparks (approved):

PR #4767, opened 2022-11-16
refactor: clarify handling what groups looked like at past session times

rjsparks:
This preserves the current functionality, which on review is a bit coarse - it would be more correct to model the group at the time the session met, not just when the meeting that the session was in started (groups get closed during IETF meetings). But that's chasing a small enough bit of diminishing return that I don't think it's work pursuing.

This also (tries to) preserves the current optimization of gathering the GroupHistory objects out of the database in a small number of queries, which is the motivation for some of the spaghetti that had crept into the codebase here. It's not clear that that optimization is the best one, but that should be pursued as part of a more general optimization task.

### DIFF
--- a/ietf/group/factories.py
+++ b/ietf/group/factories.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2015-2020, All Rights Reserved
+# Copyright The IETF Trust 2015-2022, All Rights Reserved
 import datetime
 import debug # pyflakes:ignore
 import factory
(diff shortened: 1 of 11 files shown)


jennifer-richards (changes requested):
This looks like a nice improvement to me. One concern is whether there will be situations where `group_at_the_time()` fails. I flagged a spot where some code has a fall-back to handle the `None` case, but looking at the implementation of `group_at_the_time()`, I think that situation would cause a `KeyError`. Looking deeper, I don't think that can happen because `find_history_replacements_active_at()` will always include a value for every object. 

So I think the change I'm requesting is actually to drop the `or self.session.group` in meeting/models.py, but consider whether you agree with my analysis. (If I'm mistaken and it is possible to have no "at-the-time" group, that should probably be fixed as it's a bit nonsensical.)

jennifer-richards (approved):

Issue #4706, opened 2022-11-05
IETF profile should include the reviews done


evyncke:
### Description

The IETF profile page, e.g., https://datatracker.ietf.org/person/rjsparks@nostrum.com, should also include the directorate reviews done in the last year.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

evyncke:
Just a wild guess... I used your own history of reviews and it is really long... Perhaps only displaying the last year but showing the count of all reviews ?

PR #4768, opened 2022-11-17
feat: Display last year reviews in the profile

evyncke:
#4706 : display a limited amount of directorate reviews.

Goal is to give some recognitions (or even incentives) to active directorate reviewers.

As discussed with @cjbc @jgscudder @AndrewAlston and a few others

### DIFF
--- a/ietf/person/tests.py
+++ b/ietf/person/tests.py
@@ -112,6 +112,14 @@ class PersonTests(TestCase):
         r = self.client.get(url)
         self.assertContains(r, person.name, status_code=200)
 
+    def test_case_insensitive(self):
+        # Case insensitive seach
+        person = PersonFactory(name="Test Person")
+        url = urlreverse("ietf.person.views.profile", kwargs={ "email_or_name": "test person"})
+        r = self.client.get(url)
+        self.assertContains(r, person.name, status_code=200)
+        self.assertNotIn('More than one person', r.content.decode())
+
     def test_person_profile_duplicates(self):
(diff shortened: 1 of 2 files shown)


rjsparks:
I'm moving this to draft - there's more that needs to be done, and some things to talk about.
- [ ] It needs tests
- [ ] It should _probably_ reuse code and html from the personal profile page
- [ ] It needs (as the comments indicate) to filter on time
- [ ] The primary queryset could be adjusted to make better use of the underlying Review models rather than the Document model.

rjsparks:
It's a bit frustrating that ReviewAssignment objects are linked to Email objects rather than Person objects. We should at some point reconsider how we support letting people use different email addresses for different activities. But given that this is how the models are constructed at the moment, consider:
```
ReviewAssignment.objects.filter(reviewer__person=person, state_id='completed', completed_on__gte=timezone.now()-datetime.timedelta(days=365))
```

You can then easily get to things like:
```
>>> Counter(ReviewAssignment.objects.filter(reviewer__person=rjs, state_id='completed', completed_on__gte=timezone.now()-datetime.timedelta(days=365)).values_list('result_id',flat=True))
Counter({'ready': 5, 'ready-issues': 4, 'ready-nits': 2, 'issues': 1, 'almost-ready': 1})
```

But, again, this is already in the logic used to show reviews at pages like /accounts/review/ and we should look at reusing what's there.

evyncke:
Happy to keep learning ;-)

PR #4774, opened 2022-11-18
feat: add row of button-links to group meetings and meeting session views

rjsparks:
This should be evaluated as if (or when) #4767 has been merged. In particular, review of the refactor commit should be done in that PR and not this one. For this, focus on ed29b24

This brings ietf/templates/meeting/session_buttons_include.html back into the project, based on a older version - fixup was non-trivial, so this file deserves careful scrutiny, particularly from @larseggert to make sure the framing remains as intended.

This removed the separate `Edit materials` button on the group meetings page rows and replaced it with a pencil icon on the `Materials` button. (Both previously pointed to the same URL).

Remaining Nits: Now the calendar icon is in a different place and styled differently than the other buttons. It should probably be moved in with them.

### DIFF
--- a/ietf/templates/group/meetings-row.html
+++ b/ietf/templates/group/meetings-row.html
@@ -70,9 +70,7 @@
                     {% if s.meeting.type.slug == 'interim' %}
                         {% include "meeting/interim_session_buttons.html" with show_agenda=False show_empty=False session=s meeting=s.meeting use_notes=s.meeting.use_notes %}
                     {% else %}
-                        {% with schedule=s.meeting.schedule %}
-                            {% include "meeting/session_buttons_include.html" with show_agenda=False item=s.official_timeslotassignment session=s meeting=s.meeting use_notes=s.meeting.use_notes %}
-                        {% endwith %}
+                        {% include "meeting/session_buttons_include.html" with show_agenda=False item=s.official_timeslotassignment session=s meeting=s.meeting use_notes=s.meeting.use_notes %}
                     {% endif %}
                 </div>
             {% endif %}
--- a/ietf/templates/group/meetings-row.html
+++ b/ietf/templates/group/meetings-row.html
@@ -57,14 +57,24 @@
                 Minutes
             </a>
             <a class="btn btn-primary btn-sm"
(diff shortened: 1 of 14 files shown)


ietf/templates/group/meetings-row.html
```diff
@@ -57,14 +57,24 @@
...
-                   href="{% url 'ietf.meeting.views.session_details' num=s.meeting.number acronym=s.group.acronym %}">
-                    Edit Materials
-                </a>
+            {% if not s.cancelled %}
+                <div class="regular float-end">
+                    {# see note in the included templates re: show_agenda parameter and required JS import #}
+                    {% if s.meeting.type.slug == 'interim' %}
+                        {% include "meeting/interim_session_buttons.html" with show_agenda=False show_empty=False session=s meeting=s.meeting use_notes=s.meeting.use_notes %}
+                    {% else %}
+                        {% with schedule=s.meeting.schedule %}
```
jennifer-richards, line 73: I don't see anything using this `schedule` variable? Drop? (Or make a parameter to the `{% include %}` instead, if somehow it is being used?)

ietf/templates/meeting/session_details_panel.html
```diff
@@ -21,6 +21,10 @@ <h3 class="mt-4" id="session_{{ session.pk }}">
                 {# see note in the included templates re: show_agenda parameter and required JS import #}
                 {% if meeting.type.slug == 'interim' %}
                     {% include "meeting/interim_session_buttons.html" with show_agenda=False show_empty=False %}
+                {% else %}
+                    {% with schedule=s.meeting.schedule %}
```
jennifer-richards, line 25: Same - is something using `schedule`?

ietf/templates/meeting/session_buttons_include.html
```diff
@@ -0,0 +1,365 @@
+{# Copyright The IETF Trust 2015, All Rights Reserved #}
```
jennifer-richards, line 1: update copyright year?

jennifer-richards (changes requested):
Just a couple minor things to check / fix.

jennifer-richards (approved):

PR #4785, opened 2022-11-23
fix: Don't show reorder UI fixtures unless user can reorder

### LINKED ISSUE #4773: Dragdrop handles appear on session detail page even when not logged in
At views like https://datatracker.ietf.org/meeting/115/session/stir, if the user doesn't have permission to change the slide order, the handles and tooltips for the handles should not be in the html.

larseggert:
Fixes #4773

### DIFF
--- a/ietf/templates/meeting/session_details_panel.html
+++ b/ietf/templates/meeting/session_details_panel.html
@@ -66,21 +66,19 @@
                             <td>
                                 <a href="{{ pres.document.get_href }}">{{ pres.document.title }}</a>
                                 <a href="{{ url }}">({{ pres.document.name }})</a>
-                            </td>
-                            <td class="text-end">
                                 {% if user|has_role:"Secretariat" or can_manage_materials %}
+                                    {% if pres.document.type.slug != 'bluesheets' or user|has_role:"Secretariat" or meeting.type.slug == 'interim' and can_manage_materials %}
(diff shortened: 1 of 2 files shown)


ietf/templates/meeting/session_details_panel.html
```diff
@@ -66,22 +66,20 @@ <h3 class="mt-4">Agenda, Minutes, and Bluesheets</h3>
                             <td>
                                 <a href="{{ pres.document.get_href }}">{{ pres.document.title }}</a>
                                 <a href="{{ url }}">({{ pres.document.name }})</a>
-                            </td>
-                            <td class="text-end">
                                 {% if user|has_role:"Secretariat" or can_manage_materials %}
+                                    {% if pres.document.type.slug != 'bluesheets' or user|has_role:"Secretariat" or meeting.type.slug == 'interim' and can_manage_materials %}
```
rjsparks, line 70: Why did this block move?

rjsparks (approved):

ietf/templates/meeting/session_details_panel.html
larseggert: So the float-end looks better.

rjsparks: I don't see, at this point, how the float-end could have been affected?

PR #4786, opened 2022-11-23
fix: Ignore `readonly` on `select`

larseggert:
This was caused by #4708


rjsparks (approved):

jennifer-richards (commented):
Hm, I'd overlooked that a `select` really is not supposed to have a `readonly` attribute. I think this change is fine, pragmatically, but it'd perhaps be better to replace this hack [edit: meaning the one in #4708] with a better fix. Standard suggestion is to `disable` the `select` and add a hidden input to pass the value back to the form.

That'll take some nontrivial work, though, so I'd be fine with committing this and creating a ticket to clean things up.

Unless HTML 5 has changed the landscape - have not looked deeply into it.

PR #4792, opened 2022-11-24
fix: Allow more email addresses in notification lists

### LINKED ISSUE #4736: Maximum number of email addresses on send notices to
### Describe the issue

I cannot add 16 email addresses to the send notices.  Is it possible to increase the limit?  Is the editing limit due to characters or number of email addresses. 

Background.  Draft was limited to 2 editors from a combination of authors + editors.   The editors want to automatically notify the original authors and editors. 

page to reference draft-ietf-idr-bgp-car-00.txt 

### Code of Conduct

larseggert:
Fixes #4736


rjsparks (changes requested):
This is insufficient - a matching change to the document model is also needed, or the extra bits won't be stored in the database. See  https://github.com/ietf-tools/datatracker/blob/main/ietf/doc/models.py#L111. 

It's not immediately clear whether leaving this as a `CharField` in the model and making it significantly larger is the right thing to do. I suspect it probably needs to be converted to a `TextField`. Either way, the change will need to be accompanied by a migration effecting the change in the database.

(You build that with something like `ietf/maange.py makemigrations doc` after changing the model.)

ietf/doc/migrations/send_notices.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.28 on 2022-12-05 14:37
```
rjsparks, line 1: This needs to be renamed to be 0048_send_notices.py (Though i'd suggest 0048_longer_notify_field.py)

rjsparks (changes requested):

rjsparks (approved):

jennifer-richards (approved):
Only comment is that I'd probably have named the migration something more descriptive (`0048_lengthen_notification_lists` or something), but that's a small nit. Looks good otherwise.

uzkjzjjz:
Notes_221114_232000_f65

PR #4794, opened 2022-11-24
chore: Use `codespell` to fix typos in comments.

larseggert:
First part of replacement of #4651


ietf/ietfauth/tests.py
```diff
@@ -197,7 +197,7 @@ def register_and_verify(self, email):
     def test_create_whitelisted_account(self):
         email = "new-account@example.com"
 
-        # add whitelist entry
```
rjsparks, line 200: This change goes a little beyond a spelling fix, but its not a blocking point.

rjsparks (approved):

PR #4797, opened 2022-11-25
chore: Use codespell to fix typos in code.

larseggert:
Second part of replacement of #4651

@rjsparks, I probably need to revert some things here, and I also still need to add that new migration - how do I do that?


rjsparks:
After you change the model, run `ietf/manage.py makemigrations`. If it gives the migration an inscrutable name, rename it (or just pass --name into the makemigrations invocation in the first place).

larseggert:
I just tried `ietf/manage.py makemigrations` but it says `No changes detected`. Did I not change the model in the current PR?

rjsparks:
Because you changed the names in the _already existing migrations_. 

Changing already applied migrations is never the right thing to do (unless you plan to run the reverse and forward again on the production database). Since you changed ietf/mailinglists/migrations/0001_initial.py, _Django_ thinks that the class has _always_ been named Allowlisted, and doesn't know that it needs to make changes to the database.

larseggert:
Thanks! I will try to fix next week.

ietf/utils/management/commands/coverage_changes.py
```diff
@@ -235,7 +235,7 @@ def handle(self, *args, **options):
             # verbosity = int(options.get('verbosity'))
             if not filenames:
                 filenames = [
-                    getattr(settings, 'TEST_COVERAGE_MASTER_FILE'),
+                    getattr(settings, 'TEST_COVERAGE_main_FILE'),
```
rjsparks, line 238: This isn't what is intended, and I suspect it might break the interface with coverage.

ietf/utils/management/commands/coverage_changes.py
```diff
@@ -34,7 +34,7 @@ class Command(BaseCommand):
             "        $ manage.py {name} --absolute --sections=url | grep False\n"
             "\n".format(**locals())
         )
-    args = "[[master_json] latest_json]"
+    args = "[[main_json] latest_json]"
```
rjsparks, line 37: Where is main_json (master_json) otherwise defined? Does it belong to coverage instead of this code?

rjsparks (changes requested):

ietf/static/js/document_relations.js
```diff
@@ -315,7 +315,7 @@ function draw_graph(data, group) {
         })
     ];
 
-    // // See https://github.com/d3/d3-force/blob/master/README.md#simulation_tick
+    // // See https://github.com/d3/d3-force/blob/main/README.md#simulation_tick
```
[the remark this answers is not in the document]
larseggert: They made the change.

ietf/utils/management/commands/coverage_changes.py
larseggert: This is just the help text, I believe.

ietf/person/management/commands/deactivate_email_addresses.py
```diff
@@ -84,7 +84,7 @@ def handle(self, *args, **options):
                     email.origin = email.person.user.username if email.person.user_id else ('script: %s deactivation' % options['reason'])
                     email.save()
                     PersonEvent.objects.create(person=email.person, type='email_address_deactivated',
-                        desc="Deactivated the email addres <%s>. Reason: %s" % (email.address, options['reason']) )
+                        desc="Deactivated the email address <%s>. Reason: %s" % (email.address, options['reason']) )
```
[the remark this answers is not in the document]
rjsparks: Several steps.
* `ietf/manage.py makemigrations --empty --name=personevent_spelling person
* Add a forward and reverse function to the empty migration it creates and wire them into `migrations.RunPython` (search for `def forward(` to find lots of examples.
* update the dependencies list to the tip of any apps from which you end up pulling in the model. (You'll at least pull in PersonEvent from Person, using `apps.get_model`, not an import, but you'll get the right person dependency automatically given that you made the migration on the person app. Note also that what you get is a facade with a working object manager, not the actual class - member functions of the class will not be available).
* the reverse should just `pass`. For the forward, you'll want to figure out how to iterate over the objects that have the spelling errors and update them as efficiently as you can (since however long this runs is downtime as the release it contains is deployed).

More to read at https://docs.djangoproject.com/en/2.2/topics/migrations/

ietf/dbtemplate/fixtures/nomcom_templates.xml
```diff
@@ -166,7 +166,7 @@ Thank you,</field>
 Thank you for accepting your nomination for the position of $position.
 
 Please remember to complete and return the questionnaire for this position at your earliest opportunity.
-The questionaire is repeated below for your convenience.
+The questionnaire is repeated below for your convenience.
```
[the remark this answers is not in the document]
rjsparks: Please wait to resolve these until the checkbox gets ticked - I haven't gotten to do the work there yet.

ietf/name/fixtures/names.json
```diff
@@ -385,7 +385,7 @@
   },
```
rjsparks, line 385: I've made the changes from this file to the production database. It would be better if this PR either didn't change this file at all, or update it to match production by fetching an updated database (after tonight's dump) and then running `ietf/manage.py generate_name_fixture --stdout | jq --sort-keys 'sort_by(.model, .pk)' >
ietf/name/fixtures/names.json`

rjsparks: meh - nevermind - I'll take the action to update main after this is merged - I suspect other names need to be updated from other PRs anyhow.

rjsparks (approved):

rjsparks (changes requested):
whoops - still need the migration for older PersonEvent objects (unless that's too much to ask - we can take this without it and go work that as a separate task.)

ietf/mailinglists/migrations/0003_allowlisted.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.28 on 2022-12-05 14:26
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('person', '0029_use_timezone_now_for_person_models'),
```
jennifer-richards, line 9: Probably doesn't strictly need this dependency, but harmless.

jennifer-richards (changes requested):
Missed renaming a method after fixing the same word in its code. Should fix that while in the area.

A couple other minor suggestions inline that can be taken or left as you prefer.

ietf/person/management/commands/deactivate_email_addresses.py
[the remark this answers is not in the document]
larseggert: I think this is beyond my paygrade...

jennifer-richards (approved):

rjsparks (approved):

PR #4799, opened 2022-11-28
fix: Fix spurious author extraction errors

larseggert:
This fixes edge cases including the following names:
* Joris van Kuijc van Malsen
* दिल
* शिला के.सी.
* Lady Garcia
* Carolina de la O
* உயிரோவியம் தங்கராஐ;
* Alexandre Koch II
* இந்திராணி ஆதர்ஷ்,ஆதேஷ்


larseggert:
@rjsparks I'm still running the tests in a loop to generate more failures, but I think this is ready for review in the meantime.

ietf/utils/draft.py
```diff
@@ -820,8 +825,9 @@ def dotexp(s):
                     first, last = author.rsplit(".", 1)
                     first += "."
                 else:
-                    author = "[A-Z].+ " + author
-                    first, last = author.rsplit(" ", 1)
+                    # handle single-word names
```
jennifer-richards, line 828: Was the `[A-Z].+` doing anything functional before?

jennifer-richards (approved):
One question, mostly out of curiosity, but this looks ok to me.

larseggert:
Found another case, where "இந்திராணி ஆதர்ஷ்,ஆதேஷ்" renders as "Intiraanni Aatrss,aateess" and the comma is confusing the current regex.

ietf/utils/draft.py
larseggert: I don't think so. If we end up in that part of the conditional, `author` does not contain a space, so it is a single word. So we'd never match a first name.

larseggert:
No more failures in the last few hours.

ietf/utils/draft.py
rjsparks: It looks like a bug to me. At the very least, the old code was a strange construct and I suspect it is an edit-fail from long ago.

jennifer-richards: Thanks, that's what it looked like but the code is a bit tricky (to say the least).

rjsparks (approved):

PR #4823, opened 2022-12-02
fix: don't use rfc names in bibxml-id elements

rjsparks:
fixes #4822


cesarapascual (commented):
ietf/doc/utils.py

cesarapascual (commented):
gh pr checkout 4823

jennifer-richards (approved):

PR #4845, opened 2022-12-07
fix: ignore badly coded user-provided html of older documents

### LINKED ISSUE #4842: htmlization page for rfc441 is 500ing
Unicode decode error that does not occur in the dev machines given the same input.

rjsparks:
Fixes #4842


uzkjzjjz:
Notes_221114_232000_f65

PR #4851, opened 2022-12-08
feat: Add links to tools, so people know where to file issues.

larseggert:


rjsparks (approved):

rjsparks:
Slightly twitchy about these URLs being directly in templates rather than in config, but we can move them to config if we ever have a second place that uses them.

PR #4853, opened 2022-12-08
fix: handle groups_at_the_time cache misses

### LINKED ISSUE #4852: Server error loading agenda for IETF-87
### Describe the issue

Attempting to load either the main or iCalendar agendas for meeting 87 results in a server error.

This happens because the `Meeting.groups_at_the_time()` method looks up and caches group meeting-time historical data for the groups and group parents, then refers to the cache to render the response. This process does not account for the possibility that a group has changed parents since the meeting. When that happens, the group's *current* parent is in the cache, but possibly not its *historical* parent.

### Code of Conduct

jennifer-richards:
Fixes #4852

### DIFF
--- a/ietf/meeting/models.py
+++ b/ietf/meeting/models.py
@@ -409,28 +409,34 @@ class Meeting(models.Model):
     def uses_notes(self):
         return self.date>=datetime.date(2020,7,6)
 
-    def groups_at_the_time(self):
+    def meeting_start(self):
+        """Meeting-local midnight at the start of the meeting date"""
+        return self.tz().localize(datetime.datetime.combine(self.date, datetime.time()))
+
+    def _groups_at_the_time(self):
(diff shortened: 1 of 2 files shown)


jennifer-richards:
For IETF-87, this currently adds 15 group history lookups (~2 queries each, I think) to the main agenda view to look up the one parent group (pk=1683) that is not in the cache. The 15 is because we apparently have at least 15 meeting instances hanging around during the rendering process, each with their own cache.

For meetings not affected by group parent changes (i.e., most of them), this code does not add any db calls.

To put the additional 30ish queries in context, I believe there are 150-200 bulk lookups of the group histories during main agenda rendering, with a few queries each, so this might be adding 5% to the queries involved in group history retrieval.

rjsparks (approved):

PR #4856, opened 2022-12-08
feat: validate the document notify field

rjsparks:
Allow for sloppier entry given that the widget is now a Textfield.

Rename the most recent migration.


jennifer-richards (approved):
Looks ok. You might consider using `email.utils.parseaddr` instead of rolling your own parser. It doesn't enforce the particular rule of having a closing `>` though (and can be quirky).

rjsparks:
There's also ietf.utils.mail.parseaddr, but I wanted to go with something that would be over-accepting initially.

rjsparks:
looks like the implementation right now doesn't preserve order and the tests don't  consistently pass.

ietf/doc/forms.py
```diff
@@ -87,24 +87,39 @@ class NotifyForm(forms.Form):
...
+                addrspec = nameaddr[nameaddr.find("<")+1:-1]
             else:
                 addrspec = nameaddr
             try:
                 validate_email(addrspec)
             except ValidationError:
                 bad_nameaddrs.append(nameaddr)
+        error_messages = []
+        if len(duplicate_nameaddrs) != 0:
+            error_messages.append(f'Duplicate addresses: {", ".join(duplicate_nameaddrs)}')
```
jennifer-richards, line 117: Is there a need to make the user manually remove duplicates rather than simply dropping one? I guess maybe that isn't desirable if the order is important.

Maybe more of a feature request than a review comment, but I wonder if it'd be more helpful to filter for duplicates on `addrspec` rather than `nameaddr`. Having both `John Smith <js@example.com>` and `Smith, J <js@example.com>` is a pretty common dup error for mailing lists. If we're going to insist the user fix duplicates, would this be more helpful to report?

jennifer-richards (changes requested):
The dup handling seems a little awkward - I feel like it should automatically remove exact duplicates and/or prompt the user to fix up cases of duplicate addresses that can't be automatically resolved. That's growing the scope a bit, though, so I don't insist.

I think there's enough here that a test case is needed unless it's disproportionately awkward to set up.

ietf/doc/forms.py
rjsparks: I originally automatically just removed duplicates, but in manually testing it found that it was surprising to put a new address on the end (not seeing that it was already there) and having it not show up on the end after the form supposedly validated and saved. Since we're letting these strings get really long, calling out duplicates rather than quietly burying them seemed the right thing to do.

rjsparks: You are right - the duplication concern is really at addrspec - will adjust.

jennifer-richards (approved):
Using a `dict` as an ordered set was both clever and prescient!

PR #4861, opened 2022-12-09
fix: For RFCs, correctly diff it against final I-D version

### LINKED ISSUE #4857: Compare links for RFCs compare drafts
### Describe the issue

On /doc/html/rfcXXXX the "Compare Versions" feature doesn't really produce what is expected for an RFC.  The comparison is between the last draft before publication and its immediate predecessor.

p.s., The new view on the datatracker for drafts and RFCs is excellent.

### Code of Conduct

larseggert:
Fixes #4857


larseggert:
OBE, see #4863 instead.

PR #4863, opened 2022-12-09
feat: Diff arbitrary versions from new HTMLization page

### LINKED ISSUE #4859: Diff any version for /doc/html
### Description

The new interface for /doc/html is great, but it inspired me to suggest a tweak.

Currently, the "Compare Versions" buttons offer options to compare in different ways (inline vs. side-by-side), but there is no easy way to choose which versions are compared.  I often find that I need to compare across multiple revisions, particularly when a document has gone through a lot of churn around IESG review time.

It might be relatively easy to add a dropdown/select form element that is populated with all of the versions of the document and have the button submit the form (using "GET") in a way that allows better access to comparisons.  This dropdown/select could be populated with all other versions (excluding the current), but default to the immediately preceding version.

### Code of Conduct

larseggert:
Fixes #4859 and #4857


rjsparks:
This is starting to repeat something that's implemented elsewhere much more completely (the diff form on the history page). For example - as is, it won't let you compare a wg-00 with the individual document that it replaces.

Could you just _use_ the form code from the history view instead? (Refactoring if necessary to support it being used in more than one place).

larseggert:
I'll try.

rjsparks:
With a fresh build of containers built on this PR (and putting the right drafts into /assets), I get the following:
image
which I don't think is intended?

ietf/doc/views_doc.py
```diff
@@ -533,6 +533,7 @@ def document_main(request, name, rev=None, document_html=False):
                                        review_assignments=review_assignments,
                                        no_review_from_teams=no_review_from_teams,
                                        due_date=due_date,
+                                       diff_revisions=get_diff_revisions(request, name, orig_doc) if document_html else None
```
rjsparks, line 536: There's a pattern that occurs quite often elsewhere in the code to deal with what this is achieving. Consider using it here since 145 and 536 are so far apart:

`doc if isinstance(doc,Document) else doc.doc`

(At some future refactor we may want to make this a method on DocumentInfo)

ietf/templates/doc/document_history_form.html
```diff
@@ -0,0 +1,97 @@
+{# Copyright The IETF Trust 2015, All Rights Reserved #}
```
rjsparks, line 1: 2014-2022

ietf/templates/doc/document_info.html
```diff
@@ -79,14 +79,13 @@
                 {% include "doc/revisions_list.html" with document_html=document_html %}
             </td>
         </tr>
-        {% if doc.rev != "00" %}
+        {% if doc.rev != "00" and diff_revisions|length > 1 %}
```
rjsparks, line 82: This prevents me from diffing a new wg-00 with whatever nonwg draft it replaces.

rjsparks (changes requested):
This is getting closer. It really needs tests.
There's enough going on here that I'll also ask for another review.

ietf/doc/views_doc.py
```diff
@@ -901,17 +902,28 @@ def document_email(request,name):
...
-def document_history(request, name):
-    doc = get_object_or_404(Document, docalias__name=name)
-    top = render_document_top(request, doc, "history", name)
-
+def get_diff_revisions(request, name, doc):
     # pick up revisions from events
     diff_revisions = []
 
-    diffable = [ name.startswith(prefix) for prefix in ["rfc", "draft", "charter", "conflict-review", "status-change", ]]
+    diffable = [
```
jennifer-richards, line 909: Trivial, but would be clearer to me if `diffable` were a Boolean - move the `any()` here?

ietf/templates/doc/document_history.html
```diff
@@ -18,85 +18,7 @@
...
-                               name="difftype"
-                               value="--hwdiff"
-                               id="hwdiff">
-                        <label for="hwdiff" class="btn btn-outline-primary">Wdiff</label>
-                    </div>
-                </div>
-            </div>
-            <button type="submit" class="btn btn-primary mb-3">Submit</button>
-        </form>
+        {% include "doc/document_history_form.html" %}
```
jennifer-richards, line 21: I'm not sure how Robert feels, but when including templates likely to be used in different views I think we ought to be passing the parameters as part of the `{% include %} instead of sharing the rendering context. I.e.,
```
{% include "<template path>" action_url=rfcdiff_base_url doc=doc blah=blah foo=bar only %}
```
That makes it much easier to sort out how to use the template in a new place.

jennifer-richards (commented):
Looks ok to me. I very much like separating out the `get_diff_revisions()` method and agree that some tests are warranted.

I have a couple suggestions inline.

ietf/templates/doc/document_history.html
rjsparks: I like the idea

rjsparks (approved):
I think the factored out get_diff_revisions deserves a much richer unit test, but I can work on that separately. 

There are assumptions baked in for the corner cases that tests can better document and catch violations around - things like:
* when to stop going down D->replaces->C->replaces->B->replaces->A
* What to do with 
```
C->replaces->B
            \->A
```
etc.

jennifer-richards (approved):

PR #4880, opened 2022-12-13
fix: Don't pass an empty doc rev into ietf.doc.views_doc.document_bibtex

### LINKED ISSUE #4878: htmlized doc view failing to reverse bibtex URL for some rfcs
### Describe the issue

Retrieving `/doc/html/fyi5` or `/doc/html/fyi27` from production or the sandbox fails with a reverse lookup failure. The former fails with
```
Reverse for 'ietf.doc.views_doc.document_bibtex' with keyword arguments '{'name': 'rfc1178', 'rev': ''}' not found. 1 pattern(s) tried: ['doc/(?P<name>[A-Za-z0-9._+-]+?)(?:/(?P<rev>[0-9]{1,2}(-[0-9]{2})?))?/bibtex/$']
```
The problem is apparently that it is asking for a blank rev- if I leave out the rev parameter entirely, the reverse goes through.

### Code of Conduct

larseggert:
Fixes #4878


jennifer-richards (approved):

rjsparks:
OBE by #4887

PR #4882, opened 2022-12-13
fix: Set font size from cookie before DOM is ready, to avoid flickering

larseggert:


ietf/static/js/document_html.js
```diff
@@ -11,22 +11,23 @@ import { populate_nav } from "./nav.js";
...
-document.addEventListener("DOMContentLoaded", function (event) {
-    // handle point size slider
-    const cookie = "doc-ptsize-max";
+// set initial point size from cookie before DOM is ready, to avoid flickering
+const ptsize_cookie = "doc-ptsize-max";
+
+function change_ptsize(ptsize) {
+    document.documentElement.style.setProperty(`--${ptsize_cookie}`,
+        `${ptsize}pt`);
+    cookies.set(ptsize_cookie, ptsize);
```
NGPixel, line 20: Use [window.localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#examples) instead of cookies. Otherwise you're needlessly sending that value on every request to the server (including assets).

NGPixel (changes requested):

NGPixel (approved):

PR #4886, opened 2022-12-13
feat: Add preference to make citation links go straight to cited docs

larseggert:


ietf/static/js/document_html.js
```diff
@@ -68,4 +68,36 @@ document.addEventListener("DOMContentLoaded", function (event) {
     };
     defpane.show();
     document.activeElement.blur();
+
+    if (cookies.get("reflinks") != "refsection") {
```
rjsparks, line 72: Is this another place localStorage should be used?

rjsparks (approved):

PR #4890, opened 2022-12-14
chore: use prebuilt pg db image in dev

NGPixel:
This PR completely replaces MariaDB with PostgreSQL in dev, using the prebuilt image.


docker/scripts/app-init.sh
```diff
@@ -117,29 +115,6 @@ echo "Running initial checks..."
 
 /usr/local/bin/python $WORKSPACEDIR/ietf/manage.py migrate --settings=settings_local
 
-
```
rjsparks, line 120: We still need one call to migrate in this block - a user might have new migrations in their fork, not yet committed to feat/posgtres that need to be run.

rjsparks (changes requested):

docker/scripts/app-init.sh
NGPixel: Unlikely as its using a new db image?

rjsparks: Doh - read right past new L116

rjsparks (approved):

PR #4896, opened 2022-12-14
refactor: add guards/type hints to build_file_urls()

jennifer-richards:


ietf/doc/utils.py
```diff
@@ -995,8 +996,11 @@ def get_search_cache_key(params):
     key = "doc:document:search:" + hashlib.sha512(json.dumps(kwargs, sort_keys=True).encode('utf-8')).hexdigest()
     return key
     
-def build_file_urls(doc):
-    if isinstance(doc,Document) and doc.get_state_slug() == "rfc":
+def build_file_urls(doc: Union[Document, DocHistory]):
+    if doc.type != 'draft':
```
rjsparks, line 1000: doc.type_id I think?

rjsparks (changes requested):

ietf/doc/utils.py
jennifer-richards: Yes, thank you

jennifer-richards:
Added a test in addition to the correction you caught.

rjsparks (approved):

PR #4897, opened 2022-12-14
fix: handle names with endings that might parse as versions better

### LINKED ISSUE #4855: OAuth 2.1 links show OAuth 2.0 (Problem with drafts with version numbers in the name)
### Describe the issue

The links to OAuth 2.1 are now returning the OAuth 2.0 spec. It looks like the `-1` in the URL is being completely ignored. This seems to have changed relatively recently.

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-07

image

Previously this was showing version -07 of the OAuth 2.1 draft. https://web.archive.org/web/20221109125705/https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-07

### Code of Conduct

rjsparks:
fixes #4855


rjsparks:
@lars - this is a proposed replacement for #4862.
Either still need tests (and probably a light refactor of the method to make it more testable).

rjsparks:
Either should also try to fix #4898 at the same time.

larseggert:
Let's use this one as a base, will close #4862.

rjsparks:
OK - I've looked more closely and there's not a quick refactor that would let us test the small bit of logic that's in this functions dispatch, there are already tests of the fuzzy matcher that cover what caused the pain that led to this PR, and #4898 is a symptom of the much more general issue with viewing the last version of a draft before it became RFC. Lets solve the last one after the RFC-as-type-of-document major refactor. Thus, I think this simple fix should go forward so the oauth wg's issue gets resolved.

jennifer-richards (approved):

PR #4899, opened 2022-12-15
More fixes for HTMLization and PDFization

### LINKED ISSUE #4885: RFCs without XML are not displayed properly.  RFC6761
### Describe the issue

The `width: 72ch`  of `.rfchtml pre` elements cuts out a few columns of text.  I attach a screenshot of Firefox 102.5.0esr.

Removing the property (using the *explore* feature) renders the text properly.

screenshot

### Code of Conduct

larseggert:
This merges in some changes from @martinthomson and some other assorted fixes.

Fixes #4885.


ietf/doc/models.py
```diff
@@ -617,7 +617,7 @@ def pdfized(self):
         if text:
             stylesheets.append(finders.find("ietf/css/document_html_txt.css"))
         else:
-            text = self.htmlized()
+            text = f'<div class="rfcmarkup">{self.htmlized()}</div>'
```
rjsparks, line 620: maintainability question: why didn't this go into the htmlized function? If this is really only pdf specific, then maybe the class should reflect that so it doesn't get applied to other things by future people who didn't understand the intent?

larseggert: It could go there. I'll move it.

rjsparks (approved):

PR #4923, opened 2022-12-21
feat: Render fresh HTML and TXT for display when we have the XML

larseggert:
For comments.


ietf/doc/models.py
```diff
@@ -523,9 +524,81 @@ def replaces_canonical_name(self):
     def replaced_by(self):
         return set([ r.document for r in self.related_that("replaces") ])
 
+    def render_txt_and_html(self):
```
rjsparks, line 527: This is going to take too long to run - users will get timeouts from the cdn,similar to what they get now sometimes posting a draft via the UI instead of the async API. What needs to happen here instead is to dispatch work to the celery containers we have now instead, doing something reasonable for _this_ attempt to view (Probably best to show whatever we _have_ rather than return a "Check back in a few minutes" or the like). But we have to build that scaffolding first.

rjsparks:
For housekeeping - could we convert this to an issue (we need a different solution) and close the PR?

larseggert:
Of course.

cabo:
I don't understand this.  Rendering XML at time of retrieval does not work for I-Ds, which could have xi:include to drafts without revision indicators.  You need to capture which actual I-D revision that was at the time of submission.

rjsparks:
> I don't understand this. Rendering XML at time of retrieval does not work for I-Ds, which could have xi:include to drafts without revision indicators. You need to capture which actual I-D revision that was at the time of submission.

Yes, that's a separate issue that we're already tracking. Before this will make sense, that has to have been done.

I think the point of this PR (and we should have the discussion in the issue this moves to) is to get fresh renderings due to xml2rfc fixes. Whether it's right to do so automatically is something I'm not sure about. 

@larseggert am I missing the point? Was there something else that rerendering here was hoping to achieve?

larseggert:
What triggered this is that we have quite a bit of old HTML that was generated by old xml2rfc versions that doesn't render correctly with the ASCIIfying CSS styles. And we never re-render that HTML, making it very hard to maintain consistent styling.

Moazzem-Hossain-pixel:
> For comments.

Could you help me to issue Entra ID card? I'm a non-experienced new user.
Might be token on etherdelta? #199

rjsparks:
closing this as it's asleep. It can be pulled out of the closed well later if we need it.

PR #4949, opened 2023-01-04
fix: test on postgres. post and check intended pks.

rjsparks:


jennifer-richards (approved):

jennifer-richards:
One minor question - should the description be "pass and check" instead of "post and check"?

rjsparks:
> One minor question - should the description be "pass and check" instead of "post and check"?

I really meant "put the right PKs in POST messages". What would make that more clear?

jennifer-richards:
Ah, maybe just capitalizing POST? The mix of punctuation and all lowercase broke my parser.

PR #4968, opened 2023-01-11
Limit the width of the document selectors to the width of the parent.

larseggert:
It's unfortunately not responsive (select2 limitation), but better than being too long.

### DIFF
--- a/ietf/templates/doc/document_history_form.html
+++ b/ietf/templates/doc/document_history_form.html
@@ -13,6 +13,7 @@
 {% endif %}
             <select class="form-select{% if document_html %} form-select-sm mb-1{% endif %} select2-field"
                     data-max-entries="1"
+                    data-width="resolve"
                     data-allow-clear="false"
                     data-minimum-input-length="0"
                     {% if not document_html %}id="url1"{% else %}aria-label="From revision"{% endif %}
@@ -40,6 +41,7 @@
 {% endif %}
             <select class="form-select{% if document_html %} form-select-sm mb-1{% endif %} select2-field"
                     data-max-entries="1"
+                    data-width="resolve"
                     data-allow-clear="false"
                     data-minimum-input-length="0"
                     {% if not document_html %}id="url2"{% else %}aria-label="To revision"{% endif %}
--- a/ietf/templates/doc/document_html.html
+++ b/ietf/templates/doc/document_html.html
@@ -41,6 +41,14 @@
         {% include "base/icons.html" %}
         {% include "doc/opengraph.html" %}
          {% analytical_head_bottom %}
+        <style>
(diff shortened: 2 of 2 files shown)


jennifer-richards:
What do you mean by "it's not responsive"? (I understand the words, but not sure what you're getting at.)

If I understand what it's doing, it should be computing the style seen by the `select` element and applying that to select2's magic. Does applying responsive style to the input not pass through correctly?

jennifer-richards:
Interesting. I toyed with it on Firefox here and it seemed to behave reasonably when the window was narrow. If I put a %-based style on the `select`, that caused it to track the window width (without that, it started narrower than the window allowed).

In any case, it looks fine to me.

larseggert:
What I mean that it looks fine with the change, but when you then narrow or widen the window, the select doesn't grow or shrink.

jennifer-richards:
I see what you mean now. Not sure why I didn't notice before - I must have mixed up what I was looking at.

If I put an explicit `style="width: 100%` on the field then it expands/shrinks with the width and never exceeds the page width. With that, it seems to behave the same with or without the `data-width="resolve"`. Maybe add this to the CSS as a workaround instead or in addition to the change in this PR?

larseggert:
@jennifer-richards I can't `style="width: 100%;"` to work? It just stays too large.

jennifer-richards:
Maybe I'm misunderstanding the intention and wandering into the woods. For me, the width tracks the `<div class="col-sm-10">`, so 100% is indeed quite wide, but not wider than the div.

larseggert:
This is what I see now:
Screenshot 2023-01-11 at 21 08 19
With the PR, it becomes:
Screenshot 2023-01-11 at 21 09 15
But when you widen the window, it stays at that width:
Screenshot 2023-01-11 at 21 09 50

jennifer-richards:
Ah, ok. I was looking at a different view (that template is also used on the 'History' tab). It looks to me like there are other width problems on the sidebar. Note the up-arrow button at the top in this screenshot - it's running off the viewport just like the select boxes. I think they're accurately filling a parent that's too wide. (In the screenshot, the top box has both `style="width:100%;"` and `data-width="resolve'`, while the bottom has neither. They both seem to behave the same.)
Screenshot from 2023-01-11 15-24-08

larseggert:
Yes, the button is in the top right of the sidebar and the sidebar gets stretched too wide by the select.

jennifer-richards (approved):

jennifer-richards:
That looks much better.

On a lark, I tried throwing in a window resize handler that destroys and re-initializes the select2 fields. It worked great when increasing the width of the window but still happily stretched off the viewport when decreasing it. It *really* doesn't want to do the right thing...

PR #4970, opened 2023-01-11
Add a datatracker button to the htmlized side pane

### LINKED ISSUE #4944: side bar datatracker package pretends to lead to versioned drafts but does not
### Describe the issue

For example, both

https://datatracker.ietf.org/doc/html/draft-davies-int-historic-04

and

https://datatracker.ietf.org/doc/html/draft-davies-int-historic-05

have a link in the top right side bar claiming to be versioned, but instead both go to the same main document page (which is good, it should just not mislead people with the version in the link name )

### Code of Conduct

larseggert:
Fixes #4944

### DIFF
--- a/ietf/templates/doc/document_html.html
+++ b/ietf/templates/doc/document_html.html
@@ -131,8 +131,8 @@
                             <i class="bi bi-arrow-bar-up"></i>
                     </button>
                     <div class="pt-2 pt-lg-3 px-md-2 px-lg-3">
-                        <p>
-                            <a href="{% url 'ietf.doc.views_doc.document_main' name=doc.canonical_name %}">
+                        <a class="btn btn-primary btn-sm" href="{% url 'ietf.doc.views_doc.document_main' name=doc.canonical_name %}">Datatracker</a>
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):

ietf/templates/doc/document_html.html
```diff
@@ -131,8 +131,8 @@
                             <i class="bi bi-arrow-bar-up"></i>
                     </button>
                     <div class="pt-2 pt-lg-3 px-md-2 px-lg-3">
-                        <p>
-                            <a href="{% url 'ietf.doc.views_doc.document_main' name=doc.canonical_name %}">
+                        <a class="btn btn-primary btn-sm" href="{% url 'ietf.doc.views_doc.document_main' name=doc.canonical_name %}">Datatracker</a>
```
rjsparks, line 134: Should this be changed to point to the same version of the metadata page that you are looking at on the /html/ page? Or are you leaving it such that it always points to the current version for a reason?

larseggert: I have no preference.

PR #4975, opened 2023-01-12
fix: rfc2html creates "a" tags without href for references, handle that

larseggert:
Fixes the issue raised by @becarpenter in https://mailarchive.ietf.org/arch/msg/tools-discuss/oz-Fl58l0cffW-NENTAAn598dqA

CC @cabo

### DIFF
--- a/ietf/utils/text.py
+++ b/ietf/utils/text.py
@@ -51,7 +51,8 @@ validate_url = URLValidator()
 
 def check_url_validity(attrs, new=False):
     if (None, "href") not in attrs:
-        return None
+        # rfc2html creates a tags without href
+        return attrs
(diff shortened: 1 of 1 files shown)


Mwilson7583 (commented):
This is for MrC, my husband, james. Finally got the proof of you getting sneaky and dumb. Acting like you have no knowledge of how to redirect or block or prevent me from seeing the messages between you nd these women. So as i said would happen.. and now that there is no dispute as to that you're behind the almost 3 years of lying and cheating while i support and get ur "medicine" and stress over bills?  I now  know that the plan to get ur disability and run out n shack up one or multiple women

rjsparks (approved):

PR #5017, opened 2023-01-23
feat: remove the format element from the bibxml-id source of truth

### LINKED ISSUE #5018: Remove the format element from the bibxml produced by the datatracker for bibxml-ids

rjsparks:
fixes #5018


rjsparks:
cc: @alicerusso, @ajeanmahoney, @cabo, @reschke

Note that this does not add a `target` attribute. We can discuss that in a separate issue.

jennifer-richards (approved):

cabo:
> Note that this does not add a `target` attribute. We can discuss that in a separate issue.

This is pretty much what kramdown-rfc now does for RFCs and I-Ds in default mode (it also deletes any target attributes on RFCs and I-Ds, so you don't need to bother putting them in).
xml2rfc does a great job computing the URIs (this is not so REST, but works great).

PR #5026, opened 2023-01-24
refactor: Simplify tests and comply with StandardJS style

jennifer-richards:
Use === instead of ==, except where it's used to match both null and undefined. (I think just null is enough because both `cookies.get()` and `Storage.getItem()` return `null` when they have no data, but no harm in being liberal.)

@larseggert may want to look to be sure I didn't break this, but I think it's equivalent.

### DIFF
--- a/ietf/static/js/document_html.js
+++ b/ietf/static/js/document_html.js
@@ -62,8 +62,8 @@ document.addEventListener("DOMContentLoaded", function (event) {
 
             const val = in_localStorage.includes(btn.name) ?
                 localStorage.getItem(btn.name) : cookies.get(btn.name);
-            if (val == id || ((val == undefined || val == null) &&
-                    btn_pref[btn.name] == id)) {
+            if (val === id ||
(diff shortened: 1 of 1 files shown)


ietf/static/js/document_html.js
```diff
@@ -62,8 +62,8 @@ document.addEventListener("DOMContentLoaded", function (event) {
 
             const val = in_localStorage.includes(btn.name) ?
                 localStorage.getItem(btn.name) : cookies.get(btn.name);
-            if (val == id || ((val == undefined || val == null) &&
-                    btn_pref[btn.name] == id)) {
+            if (val === id ||
+              ((val == null) && btn_pref[btn.name] === id)) {
```
NGPixel, line 66: Should use `===` here, as it will never be `undefined`. The extra parentheses around `val === null` could also be removed. And might be more readable to put everything on the same line.

NGPixel (changes requested):

NGPixel (approved):

rjsparks (approved):

PR #5036, opened 2023-01-25
fix: only show irtf stream ballot and pubreq buttons to the IRTF chair

### LINKED ISSUE #5009: rg chairs get the "Issue IRSG Ballot", "Begin IESG Conflict Review" and "Request publication" buttons
But when they press it, they get "permission denied".

The button should only show up for the IRTF chair.

rjsparks:
fixes #5009

Tests haven't been added to verify non-IRTF chair IRSG members don't get buttons and can't access views. Those will come as part of the refactor of the views and tests in #5011


jennifer-richards (approved):
Looks good. I'd be inclined to refactor the check added to `request_publication()` to use a single, common, `permission_denied()` call, but no objection to it as it is.

PR #5037, opened 2023-01-26
fix: Also extract document names from XML seriesInfo attributes and XInclude URLs

### LINKED ISSUE #3871: referencedby link shows partial result only
### What happened?

https://datatracker.ietf.org/doc/draft-ietf-ace-aif/referencedby/

only shows draft-ietf-ace-key-groupcomm

It should be showing:

draft-ietf-ace-key-groupcomm
draft-ietf-ace-key-groupcomm-oscore
draft-ietf-ace-mqtt-tls-profile
draft-ietf-ace-oscore-gm-admin

I haven't checked with other documents.

Safari

### Code of Conduct

larseggert:
The old code only looked in the anchor string for the document names of references, which doesn't work if the anchor uses a mnemonic. This caused lots of missed references for many documents.

The new code looks into `xi:include` URLs and `seriesInfo`, too.

Fixes #3871


ietf/doc/utils.py
```diff
@@ -771,7 +771,11 @@ def rebuild_reference_relations(doc, filenames):
     unfound = set()
     for ( ref, refType ) in refs.items():
         # As of Dec 2021, DocAlias has a unique constraint on the name field, so count > 1 should not occur
-        refdoc = DocAlias.objects.filter( name=ref )
+        refdoc = (
+            DocAlias.objects.filter(name=ref[:-3])
+            if re.match(r"^draft-.*-\d{2}$", ref)
```
rjsparks, line 776: This falls into trouble if the draft being referenced has a name that actually ends with -nn, such as `draft-ietf-tsvwg-ieee-802-11`. The pattern we've had before is to look to see if there's a matching name with the -nn first and then check to see if there's a match without it only if the first doesn't exist.

ietf/doc/models.py
```diff
@@ -642,9 +642,23 @@ def pdfized(self):
         return pdf
 
     def references(self):
+        from ietf.doc.utils import rebuild_reference_relations
+        files = {}
+        for type in ["xml", "txt"]:
+            name = self.get_file_name().replace(".txt", f".{type}")
+            if os.path.isfile(name):
+                files[type] = name
+        rebuild_reference_relations(self, files)
```
rjsparks: This adds a non-trivial penalty on _every access_ to the function. I was anticipating we would use a management command to update the database a single time (or at least on-demand) with a process other than the webserver.

I had also intended to update rebuild_reference_relations fully expand xincludes before extracting references from the XML (rather than the proposal to guess based on what the xinclude url contains).

larseggert: Argh. That was not supposed to be part of the commit - I just did this foe testing. Will remove.

ietf/utils/xmldraft.py
```diff
@@ -60,17 +60,43 @@ def cleanup():  # called when context exited, even if there's an exception
...
-        label = anchor.rstrip('0123456789')  # remove trailing digits
-        if label in ['rfc', 'bcp', 'fyi', 'std']:
-            number = int(anchor[len(label):])
-            return f'{label}{number}'
-        return anchor
+    def _document_name(self, ref):
+        """Get document name from reference."""
+        series = ["rfc", "bcp", "fyi", "std"]
+        # handle xinclude first
+        # FIXME: this assumes the xinclude is a bibxml href
```
rjsparks, line 67: What's to fix here? If it's not a bibxml href, the code won't crash. Worst case it fails and reports a reference that's not really a reference, yes? Given that this is only called with blobs taken out of the references section, I'm having a hard time constructing a likely scenario that this could happen.

rjsparks: (I am still a little itchy about looking at the href url form rather than expanding the xinclude and parsing the resulting xml instead, but also dont really want to add fetching external things at this point in the code.)

ietf/utils/xmldraft.py
```diff
@@ -60,17 +60,43 @@ def cleanup():  # called when context exited, even if there's an exception
...
+        if ref.tag.endswith("}include"):
+            name = re.search(
+                rf"reference\.({'|'.join(series).upper()})\.(\d{{4}})\.xml",
+                ref.attrib["href"],
+            )
+            if name:
+                return f"{name.group(1)}{int(name.group(2))}".lower()
+            name = re.search(r"reference\.I-D\.(.*)\.xml", ref.attrib["href"])
```
rjsparks, line 75: What if the reference was to a specific version already, e.g. `https://bib.ietf.org/public/rfc/bibxml3/reference.I-D.draft-ietf-sipcore-multiple-reasons-00.xml`?
Wouldn't this then return "draft-draft-ietf..."?

rjsparks (changes requested):

ietf/utils/xmldraft.py
larseggert: The FIXME was exactly on that, i.e., that the proper wat of doing it would be to expamnd the xinclude rather than relying on a certain format in the URL. So there could still be false negatives.

ietf/utils/xmldraft.py
larseggert: Wouldn't the URL be `https://bib.ietf.org/public/rfc/bibxml3/reference.I-D.ietf-sipcore-multiple-reasons-00.xml`?

rjsparks: no. In fact, until we broke what Tony had in mind, the URL you post would not have worked. His design (which was what was running at tools) used I-D.draft-<whatever> for versioned things and ID.<whatever> for unversioned references. We made it so that both forms work at the datatracker. There are lots of documents that use the older forms.

larseggert: Ah - fixed!

krisRam432 (commented):
@meadmaker #5086 

>

larseggert:
@rjsparks I think we're good to go?

rjsparks:
This is queued for another careful readthrough - likely midweek.

rjsparks (approved):
We can clean the older names out of the tests with another PR if the aren't really needed.

I'm still a little itchy about inferring semantics from a pointer's handle rather than de-referencing the pointer and inspecting the results, but we'll address that concern a different way.

PR #5064, opened 2023-02-01
fix: Break long words in cards to prevent layout jumps

larseggert:


ietf/templates/doc/document_bofreq.html
```diff
@@ -154,7 +154,7 @@
         <div class="card-header">
             {{ doc.name }}-{{ doc.rev }}
         </div>
-        <div class="card-body">
+        <div class="card-body text-break">
```
rjsparks, line 157: Slight fret that this will interact badly with the assumptions markdown makes when it renders html, but we can try this and watch for trouble.

rjsparks (approved):

PR #5066, opened 2023-02-01
fix: Need to linkify during Markdown rendering

larseggert:


ietf/utils/markdown.py
```diff
@@ -27,23 +37,14 @@ def markdown(text):
                 python_markdown.markdown(
                     text,
                     extensions=[
+                        # TODO: discuss which extensions we want to enable, see
```
rjsparks: We did discuss this at the time this went in. I'm ok refreshing the discussion, but what's here isn't whimsical.

rjsparks:
Is there a test that exercises this config change to make sure it does what we want, and would let us know if something later breaks it?

ietf/utils/markdown.py
larseggert: Want me to remove the comment?

larseggert:
Let me think about that.

larseggert:
@rjsparks added a test.

rjsparks (approved):

PR #5080, opened 2023-02-03
fix: Fix document display on very narrow screens

larseggert:
Issue by @martinthomson


rjsparks:
I don't object to this, but doesn't it introduce one of the places that will be painful to port forward should bootstrap significantly change under us again?

larseggert:
That's why I am adding to the existing bootstrap variable, rather than overwriting it.

NGPixel (approved):

PR #5103, opened 2023-02-08
fix: `add_links_in_new_revision_events` wasn't doing anything, remove

larseggert:
Whatever `add_links_in_new_revision_events` was supposed to do, it didn't do it. The linkification of things has been happening via the `format_history_text` template tag.

### DIFF
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -13,7 +13,6 @@ import textwrap
 
 from collections import defaultdict, namedtuple, Counter
 from typing import Union
-from urllib.parse import quote
 from zoneinfo import ZoneInfo
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


rjsparks:
format_history_text likely _broke_ add_links_in_new_revision_docevents.

What this was _supposed_ to do was get rid of the embedded `<b` tags, link directly to the .txt file in the archives, and add a "diff from previous" link to each new_revision (and submission) history entry. You can see it (from a _long_ time ago) at places like: https://web.archive.org/web/20151011043636/https://datatracker.ietf.org/doc/draft-ietf-stir-certificates/history/

This has been broken for a long time without complaint. We can remove this code - it has some unhappy things in it anyhow (hardcoded URLs). 

But do we want to bring back the the diff link? If so, add another issue to do so.

We are already tracking an issue to remove the html that's been embedded into DocHistory.desc fields over time.

rjsparks (approved):

PR #5107, opened 2023-02-08
fix: Don't show stale area URLs

### LINKED ISSUE #4840: Update links to Area wikis and bug trackers at https://datatracker.ietf.org/wg/
### Describe the issue

Especially with impending the retirement of trac.ietf.org, the links for each area at: https://datatracker.ietf.org/wg/ should be updated or removed. (Even now, most of the links are either broken or obsolete.)

### Code of Conduct

larseggert:
Also uppercase some acronyms while I'm here.

Fixes #4840

### DIFF
--- a/ietf/templates/group/active_wgs.html
+++ b/ietf/templates/group/active_wgs.html
@@ -31,18 +31,6 @@
                 {% endfor %}
             </ul>
         {% endif %}
-        {% comment %}
-        {% if area.urls %}
-            <h3 class="mt-3 navskip">{{ area.acronym|upper }} Area-specific web page{{ area.urls|pluralize }}</h3>
-            <ul class="list-unstyled">
-                {% for url in area.urls %}
-                    <li>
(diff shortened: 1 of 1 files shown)


ietf/templates/group/active_wgs.html
```diff
@@ -31,8 +31,9 @@ <h3 class="mt-3 navskip">
                 {% endfor %}
             </ul>
         {% endif %}
+        {% comment %}
```
rjsparks, line 34: Why comment instead of delete?

larseggert: I thought maybe we bring them back in the future?

rjsparks: Unless you mean "in the next week or so", I'd rather delete it - recreating something like it when/if we want it back won't be that hard.

rjsparks (approved):

PR #5108, opened 2023-02-08
fix: Remove knobs for changing consensus bit IETF-stream docs

### LINKED ISSUE #4809: Remove knobs for changing the consensus bit for docs in the IETF stream
See RFC8789. Until what's stated there is changed by another RFC, the consensus bit can and should be automatically handled by the datatracker - having UI to change it only introduces confusion and unnecessary work for ADs.

larseggert:
Fixes #4809


rjsparks:
This should also set the consensus bit to yes when the document enters the IETF stream I think.

rjsparks:
Even you misunderstand what the bit _means_.
It is so horribly named.
It does NOT MEAN "this document has consensus". It only means "Use consensus boilerplate if you publish this as an RFC".

rjsparks:
We have no use in the datatracker for a separate "This document has consensus" statement other than what we already have with IESG approval.

rjsparks:
I'll go re-review the code, but at this point, unless the IESG changes its mind and decides it's ok to publish something that doesn't have IETF consensus on the IETF stream again, then I propose we remove the bit _altogether_ (I don't think it makes any difference on any other stream). If I'm right, and some future IESG wants to actually _use_ the non-consensus boilerplate on the IETF stream again, we can introduce something at that time that causes less confusion.

larseggert:
Am fully ok with ripping the bit out.

rjsparks:
@csperkins : Does IRTF still request publication with both types of boilerplate (do we need to keep this bit for the IRTF?)

csperkins:
IRTF needs this (e.g., RFC 9064 is IRTF Stream and non-consensus).

larseggert:
@rjsparks does this still need changes?

rjsparks:
I'm still working that out. It will need a block of uninterrupted time - it's not going to be part of this weeks release - but I'll have actionable feedback (or will have merged it) by next.

rjsparks:
So, yes, more definitely needs to be done, and some of what this does will make what remains harder. I'm currently thinking I'll start a new PR that cherrypicks some of this one. But the way the code treats the ConsensusDocEvents and the way it protects the views related to them is all out of sync with how we're actually using the concept - it will be a slightly bigger task to reconcile those things.

larseggert:
Closing in favor of #5184

PR #5126, opened 2023-02-10
feat: Add group stats sunburst plots to active WG page

larseggert:


rjsparks:
Are you sure you want the charts directly on the /wg page vs on their own, one-click away?
I'm wondering if they would inconvenience people using the page to navigate.
If we think the only use of the page is discovery/browsing, then having them on page makes sense.

larseggert:
In which way would the inconvenience? Because the layout jumps? 

Am also wondering if we should put one of them onto the main landing page and make the sunburst sectors links to the WG pages...

rjsparks:
I think you've been staring at these so long as part of the IESG conversation that you're giving them too much focus.

It's a lot to ask of a newcomer to understand what the sunbursts are trying to say and to guess at how the metrics they are showing should be important to them.

It's a big cognitive shift to have to wrap past a bunch of stats to get to a list of things. We cram _way_ to much on many pages.

If the main page should be stats or something that starts to explain what's going on, then sure - we should take away the search there and move search to a different page. But the result shouldn't focus on _just_ working groups. RGs, AGs, Programs, etc...

Consider the group about page - what you're proposing here is analagous to moving the document relationship diagram to being _on the about page_, not one click away from it.

rjsparks:
> Thanks, that's a good suggestion. See https://github.com/ietf-tools/datatracker/issues/5137.
Well, it wasn't what I was suggesting, but ok. What I was trying to point to was how we already treat the document dependency graphs as a "one click away" thing:
image

larseggert:
Ah! I can totally make it modals, esp. on the area page.

ietf/group/views.py
```diff
@@ -1354,6 +1354,85 @@ def group_menu_data(request):
     return JsonResponse(groups_by_parent)
 
 
+@cache_control(public=True, max_age=30 * 60)
+@cache_page(30 * 60)
+def group_stats_data(request, years="3", only_active="1"):
```
rjsparks, line 1359: This signature, and the url dispatch to it, need some tweaking.

```
  <title>OverflowError
          at /group/groupstats.json/3333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333/1</title>
```

Using an int for a bool like this (and leaving the other values unguarded is probably not a good idea.
If we really want these parameters in the URL, the signature should accept a bool, and the URL dispatcher be configured to send True or False based on something like (ignoring years for the moment):
```
/group/groupstats.json/ ends up with True being passed
/group/groupstats.json/active/ ends up with True being passed.
/group/groupstats.json/all/ ends up with False being passed.
```
But shouldn't this be modeled more as a json API where you send years and only_active in a POST?

rjsparks: Alternatively, do these _really_ need to be variants, or could they just be constants in the code now that you've explored the output some?

larseggert: I couldn't figure out how to pass arguments to template tags via URLs, maybe there is a better way?

rjsparks, line 1359: Can we pull these out of the signature for now? If not, could we at least make only_active a Boolean?

rjsparks (changes requested):

rjsparks (approved):

ietf/static/js/highcharts.js
```diff
@@ -3,11 +3,124 @@ import Highcharts from "highcharts";
...
+                                return `There ${this.value == 1 ? "is" : "are"} ${this.value} ${dataset == "docs" ? "documents" : "pages"} in ${this.name}.`;
+                            }
+                        },
+                        dataLabels: {
+                            formatter() {
+                                return this
+                                    .point.active ?
+                                    this.point
+                                    .name :
+                                    `(${this.point.name})`;
```
NGPixel, line 65: Many strange line breaks in this file. Can we put this on a single line?

ietf/static/js/highcharts.js
```diff
@@ -3,11 +3,124 @@ import Highcharts from "highcharts";
...
 window.Highcharts = Highcharts;
+
+window.group_stats = function (url, chart_selector) {
+    $.getJSON(url, function (data) {
+        $(chart_selector)
+            .each(function (i, e) {
+                const dataset = e.dataset.dataset;
+                if (!dataset) {
+                    console.log(
+                        "dataset data attribute not set");
```
NGPixel, line 40: Put on a single line.

ietf/static/js/highcharts.js
```diff
@@ -3,11 +3,124 @@ import Highcharts from "highcharts";
...
+                            }
+                        ]
+                    }],
+                });
+
+                // limit data to area if set and (for now) drop docs
+                const slice = data.filter(d => (area ==
+                            "ietf" && d.grandparent == area
+                        ) || d.parent == area || d.id ==
+                        area)
```
NGPixel, line 94: Put on a single line.

ietf/templates/group/active_areas.html
```diff
@@ -53,5 +69,16 @@ <h2 class="mt-3" id="id-{{ area.acronym|slugify }}">
...
+        {% include "group/group_stats_modal.html" with group=area only %}
     {% endfor %}
+{% endblock %}
+{% block js %}
+    <script src="{% static 'ietf/js/highcharts.js' %}"></script>
+    <script>
+        $(function () {
+            group_stats(
+                "{% url 'ietf.group.views.group_stats_data' %}",
+                ".chart");
```
NGPixel, line 81: Put on a single line.

ietf/templates/group/group_about.html
```diff
@@ -393,4 +405,12 @@ <h2 class="mt-3">
 {% block js %}
 <script src="{% static 'ietf/js/d3.js' %}"></script>
 <script src="{% static 'ietf/js/document_relations.js' %}"></script>
+    <script src="{% static 'ietf/js/highcharts.js' %}"></script>
+    <script>
+        $(function () {
+            group_stats(
+                "{% url 'ietf.group.views.group_stats_data' %}",
+                ".chart");
```
NGPixel, line 413: Put on a single line.

NGPixel (changes requested):

NGPixel (approved):

PR #5130, opened 2023-02-10
feat: Improve links to meeting tools from group pages

### LINKED ISSUE #4738: Join meetecho tool from meeting pages
### Description

Would it be possible to get the link to join the in-room meeting tool from a WG's meeting page?

Currently it's faster for me to visit the WG page and navigate to meetings, rather than loading the ITF115 entire agenda and search for a particular group.

So when I'm looking at a page like https://datatracker.ietf.org/wg/oauth/meetings/ or even https://datatracker.ietf.org/meeting/115/session/oauth it would be great to have a link to join the meetecho tool.

This is particularly relevant when I'd like to join the meetecho tool from my computer while in a meeting room, since I don't have an easy way to scan the QR code on the screen into my computer, and there is no alternate link for the QR code visible on the screen to manually enter either.

### Code of Conduct

jennifer-richards:
Fixes #4738. Fixes both views mentioned in the ticket.

On the group meeting page (`/wg/oauth/meetings`), the session buttons were included for each entry, but the `session_buttons_include.html` template required that `now` be set in the rendering context. To fix this, I've created a template tag that allows a template to get `timezone.now()` with `{% get_now as now %}` instead of relying on the view to pass it in.

For the meeting session page (`/meeting/115/session/oauth`), I've added a block for each session that shows the same links generated by the `session_buttons_include.html`, excluding a few that are redundant against what's already present. I've also updated this to use the `get_now` tag.


ietf/utils/templatetags/datetime_filters.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class GetNowNode(Node):
+    """Node that stores timezone.now() in a template variable"""
+    def __init__(self, var_name):
+        self.var_name = var_name
+
+    def render(self, context):
+        context[self.var_name] = timezone.now()
+        return ''  # render nothing
```
rjsparks, line 38: This works by side-effect of render. Is this common in the Django tree already?

I wonder if a context processor (see ietf.contextprocessors.py) injected via middleware might be better?

rjsparks: (Or if there's maybe _already_ something like "when the request arrived" in the context?)

jennifer-richards: That is the method in the documentation for setting context variables (https://docs.djangoproject.com/en/2.2/howto/custom-template-tags/#setting-a-variable-in-the-context). The render method is the hook that's available to do things when rendering the template tag.

I'll see if there's something in the `Request` or otherwise in the context. I have mixed feelings about injecting things into the context - wariness of global variables. It would have the advantage here of preventing skew, for what that's worth.

jennifer-richards: As far as I can tell there's no way to get it from the request. I think it'll work to inject it as `timezone_now` in the context. I don't think we're likely to have conflicts with that.

rjsparks: Given the doc you point to, lets leave it as is.

rjsparks (approved):

PR #5155, opened 2023-02-15
fix: Move rfcdiff-latest-json API call from /doc to /api

kesara:


ietf/doc/urls.py
```diff
@@ -85,11 +85,6 @@
     url(r'^html/(?P<name>[Rr][Ff][Cc] [0-9]+?)(\.txt|\.html)?/?$', views_doc.document_html),
     url(r'^idnits2-rfcs-obsoleted/?$', views_doc.idnits2_rfcs_obsoleted),
     url(r'^idnits2-rfc-status/?$', views_doc.idnits2_rfc_status),
-# These two are proof-of-concept of a service that would redirect to the latest version
-#    url(r'^rfcdiff-latest/%(name)s(?:-%(rev)s)?(\.txt|\.html)?/?$' % settings.URL_REGEXPS, views_doc.rfcdiff_latest),
-#    url(r'^rfcdiff-latest/(?P<name>[Rr][Ff][Cc] [0-9]+?)(\.txt|\.html)?/?$', views_doc.rfcdiff_latest),
-    url(r'^rfcdiff-latest-json/%(name)s(?:-%(rev)s)?(\.txt|\.html)?/?$' % settings.URL_REGEXPS, views_doc.rfcdiff_latest_json),
```
rjsparks, line 91: Why not leave the old urls in place (but dispatch them to the new views) for one release and remove them later so that you don't have to tightly coordinate changes at author-tools with the release of this PR?

kesara: Good idea. I have added back these URLs for the migration period.

rjsparks: Ignore it.

ietf/api/urls.py
```diff
@@ -56,6 +57,9 @@
     url(r'^version/?$', api_views.version),
     # Application authentication API key
     url(r'^appauth/[authortools|bibxml]', api_views.app_auth),
+    # latest versions
+    url(r'^rfcdiff-latest-json/%(name)s(?:-%(rev)s)?(\.txt|\.html)?/?$' % settings.URL_REGEXPS, api_views.rfcdiff_latest_json),
```
rjsparks, line 61: Sorry to ask for multiple changes, but it occurs to me that this would be better as a permanent redirect in case there are other users of the endpoint we're not yet aware of.

rjsparks (changes requested):

ietf/api/urls.py
kesara: No worries, I have updated the `ietf/doc/urls.py` URLs as permanent redirects.

rjsparks (approved):

PR #5162, opened 2023-02-16
fix: Don't show replaced, expired or otherwise abandoned docs for AD

### LINKED ISSUE #5153: How do I kill a chartering effort?
### Describe the issue

See https://datatracker.ietf.org/doc/charter-ietf-term/

I set the charter document state to "Not currently under review" and the WG state to "Abandoned", but it is still showing up on https://datatracker.ietf.org/doc/ad/lars.eggert

### Code of Conduct

larseggert:
Fixes #5153 and #2889.


larseggert:
@rjsparks better?

ietf/doc/views_search.py
```diff
@@ -679,7 +679,16 @@ def docs_for_ad(request, name):
...
-    #
+
+    # filter out some results
+    results = [
+        r
+        for r in results
+        if (r.type.slug != "charter" or r.group.state.slug != "abandon")
+        and (r.type.slug != "draft" or r.get_state("draft-iesg").slug != "dead")
+        and (r.get_state().slug not in ["repl", "replaced"])
+    ]
```
rjsparks, line 690:
```suggestion
    results = [
        r
        for r in results
        if not (
            r.type_id == "charter"
            and (
                r.group.state_id == "abandon"
                or r.get_state_slug("charter") == "replaced"
            )
        )
        and not (
            r.type_id == "draft"
            and (
                r.get_state_slug("draft-iesg") == "dead"
                or r.get_state_slug("draft") == "repl"
            )
        )
    ]
```

rjsparks (commented):
Yes, this is what you want. I do think it would be better to unwrap the logic puzzle into something that requires less thinking when the next coder gets here. I've suggested an alternative.

rjsparks (approved):

PR #5165, opened 2023-02-16
feat: treat django auth username as case insensitive throughout the datatracker

rjsparks:


ietf/secr/rolodex/views.py
```diff
@@ -91,10 +90,10 @@ def add_proceed(request):
                              )
 
             # in theory a user record could exist which wasn't associated with a Person
-            try:
+            
+            user = User.objects.filter(username__iexact=email).first()
+            if not user:
```
jennifer-richards, line 95: Restrict username to lowercase like we do in the main `create_account()` view?

jennifer-richards (changes requested):

ietf/secr/rolodex/views.py
rpcross: Probably a good idea. I've reached out to staff to see if this feature is still used. If not it should be removed.

ietf/api/views.py
```diff
@@ -192,7 +192,7 @@ def err(code, text):
             except ValueError as e:
                 return err(400, "Unexpected POST data: %s" % e)
             response = "Accepted, New registration" if created else "Accepted, Updated registration"
-            if User.objects.filter(username=email).exists() or Email.objects.filter(address=email).exists():
+            if User.objects.filter(username__iexact=email).exists() or Email.objects.filter(address=email).exists():
```
rpcross, line 195: With the existing implementation, the email address being submitted actually originates from Datatracker so the clause is not really needed. Unfortunately the the send_account_creation_email is generic, so there is no way to confirm this based on messages sent.

ietf/secr/rolodex/forms.py
```diff
@@ -85,7 +85,7 @@ def clean_user(self):
         if user:
             # if Django User object exists return it, otherwise create one
             try:
-                user_obj = User.objects.get(username=user)
+                user_obj = User.objects.get(username__iexact=user)
             except User.DoesNotExist:
                 user_obj = User.objects.create_user(user,user)
```
rpcross, line 90: May want to force lower() as with views

ietf/secr/rolodex/views.py
rjsparks: Not sure if maybe you meant to point to L96?. The code you point to is looking to see if a User already exists, and _that_ needs to use `__iexact`. I could see at 96 forcing an email.lower() for the username argument, which should be made explicit instead of positional.

Is that what you had in mind?

ietf/api/views.py
rjsparks: Better to leave it this way in case other things get out of sync.

ietf/secr/rolodex/forms.py
rjsparks: ok, again the arguments here should be explicit, not positional.

ietf/secr/rolodex/views.py
jennifer-richards: Yes, was pointing to the block rather than the specific line (github sometimes is picky about which lines you can comment on).

I guess since we already have some mixed case usernames it's not essential, but I think we should be consistent about whether we're forcing lowercase usernames or not.

rjsparks: I'll solve this in a slightly different way - this block gets `email` from NewPersonForm, and that form should include downcasing email while cleaning it.

jennifer-richards (approved):
Changes look good to me. (But yuck, it's pretty sketchy that the `clean_user()` converts a `CharField` into a `User` and sketchier still that it sometimes creates a `User` in the process. Ok to leave that be until the secr app gets retired, though, and just not let it persist past that.)

PR #5166, opened 2023-02-16
feat: Use case-insensitive matching for document searches

jennifer-richards:


rjsparks:
To inform whether to take the performance penalties of case-insensitive search over document names:

* We do not allow the creation of document names with lower case letters anywhere now.

In the past, we didn't clean as carefully, so we have this finite set of names matching `[A-Z]`:
```
draft-fricc-brim-BackboneRouting
draft-ietf-snmp-OsiSnmp
draft-ietf-appleip-MacIP
draft-chapin-clnp-ISO8473
draft-kunzinger-idrp-ISO10747
draft-ietf-dnsind-dynDNS-arch
draft-ietf-dnsind-dynDNS-impl
draft-ietf-sdr-IPv6-pack
draft-ietf-dnsind-dynDNS
draft-ietf-atommib-atm2TC
draft-various-snmpv2-adminv2C-syn
draft-ietf-atommib-perfhistTC
draft-okanoue-mobileip-R+A
draft-ietf-drums-MHRegistry
draft-ietf-ipsec-isakmp-SA-revised
draft-ietf-avt-X11-new
draft-kchapman-perfhist-sup-TC
draft-ietf-fax-feature-T30-mapping
draft-ietf-conneg-W3C-ccpp
draft-henry-DHCP-opt61-UUID-type
draft-sanchez-CSIP-v0r0
draft-sanchez-garcia-SSTP-v0r2
draft-coene-ss7-IP-addr
draft-sanchez-garcia-SSTP-v1r0
draft-rajagopalan-CR-overlay
draft-hss-megaco-R2package
draft-kumar-dna-rqmnt-IPv4
draft-nielsen-v6ops-3GPP-zeroconf-goals
slides-edu-intro-AAA
```

In the DocAlias table, the only name that is not also a Document name is
"Response to BBF liaison statement regarding Packet and DWDM/Optical integration"  pointing to liaison-2013-04-04-broadband-forum-packet-dwdm-optical-integration-attachment-1

That is probably an edge that was missed at the last cleanup of liaison attachments.

So I think the right thing to do is to make search case insensitive as the PR is currently suggesting (but to measure the performance hit, and revisit if it's severe.

We also need to check each point that can mint a _new_ Document object and verify that it checks that there isn't a case-varied collision first, similar to what we're doing for User.username.

Don't spend time exploring the option of creating forced downcased versions as DocAliases or similar - I fully expect to _remove_ DocAlias when we refactor to make RFCs first class document objects - it is a source of performance pain and code complexity that is not worth the benefit it was supposed to bring.

rjsparks:
We should address access to these documents through urls (such as the /doc/ tree) in a separate PR, but since I've called them out here, I'll leave a note:

Those views should just do a straight equality get or filter over name instead of forcing a case-insensitive lookup. For this finite set, a request for the legacy name with mixed case will match. We can add redirects from the all-lower-case to the legacy name. Any other case variants will simply be not found.

jennifer-richards:
> We should address access to these documents through urls (such as the /doc/ tree) in a separate PR, but since I've called them out here, I'll leave a note:
> 
Note that with the changes here, visiting `/doc/draft-fricc-brim-backbonerouting` (lowercase, no trailing `/`) hits the `search_by_name()` view. This then redirects to `/doc/draft-fricc-brim-BackboneRouting/` (mixed case, trailing `/`) which uses the `document_main()` view. Trying to retrieve `/doc/draft-fricc-brim-backbonerouting/` (lowercase, trailing `/`) gets a 404 until we deal with case folding there somehow.

So the `/doc/` tree is a bit mixed after this.

jennifer-richards:
I think this does the right things to doc searches and other uses of the `SearchForm`. There are a couple other form classes sharing that name that I also checked and cleaned up (the ipr was, I think, the only one that needed to be tweaked).

I've updated the test cases to try to check that the searches handle mixed up case. I'm not sure whether those are a lot of protection as long as we're using sqlite for the tests. It seems that the `__startswith` and perhapse `__contains` are case-insensitive in sqlite, so e.g., the `test_search_for_name` tests pass even without my changes.

rjsparks:
You can run the tests on postgres in your dev instance using settings_postgrestest. We might want to see if we can get gha to run tests for merges into feat/postgres using those settings. @NGPixel how much trouble would that be?

Issue #3890, opened 2022-04-26
Don't allow ADs to be added as bethere constraints for WGs


rjsparks:
Chairs have developed a practice of adding ADs to the bethere constraints in a session request. This is unnecessary, and even harmful, as it adds noise to the scheduling process which already knows that the AD for a group should be at its sessions.

Change the form to tell chairs not to add the group's AD, and block adding that AD (in rare cases, there may be more than one) from this field.

rjsparks:
Right now, the form autopopulates ADs and Chairs. We also have the issue of the "fetch from previous meeting" propagating the behavior.

Spectre17:
Per Robert Sparks, I am interpreting this issue to include Chairs along with ADs.

rjsparks:
@spectre17 - unless you have something ready to commit, we'll pull this back into the tools team

jennifer-richards:
Fixed by #5169

PR #5169, opened 2023-02-17
fix: Prevent ADs and chairs from being added to `bethere` constraint

### LINKED ISSUE #5163: Wording change on /secr/sreq
### Description

Requested by @flynnliz, please change, "You should not include the Area Directors; they will be added automatically," to something like, "You should not include your Area Director; the tool accounts for their conflicts separately," to avoid giving the impression that the AD's name will actually appear in the field.

People seem to add the AD anyway. Would be nice to prevent this because even the rephrasing may not discourage it enough.

### Code of Conduct

larseggert:
Fixes #5163 and #3890.

Also fix a js error while I'm here.


rjsparks (approved):

ietf/secr/sreq/forms.py
```diff
@@ -235,6 +236,25 @@ def clean_joint_with_groups(self):
...
+            )
+            if extra:
+                extras = ", ".join(e.name for e in extra)
+                raise forms.ValidationError(
+                    (
+                        f"Please remove the following person{pluralize(len(extra))}, the system "
+                        f"tracks their availability due to their role{pluralize(len(extra))}: {extras}."
+                    )
+                )
+        return True
```
rjsparks, line 256:
```suggestion
        return bethere
```

rjsparks (changes requested):

rjsparks (approved):

PR #5172, opened 2023-02-20
feat: Include previous draft URL in /api/rfcdiff-latest-json

kesara:


kesara:
Python test error might be a flaky datetime-related test.
Unrelated to the proposed changes.

jennifer-richards:
> Python test error might be a flaky datetime-related test. Unrelated to the proposed changes.

It's flaky datetime-related, but not related to the recent timezone work. The test fixtures in `settings.py` include `ietf.utils.test_data.make_immutable_test_data()` for every test. That fixture sets up `TelechatDate` instances that includes "tomorrow". If you run it very near midnight in `settings.TIME_ZONE`, then that might be "today" when you actually run your test. The test then creates its own `TelechatDate` for "today," which leads to there being two instead of one.

I'll create a ticket for this - it'd be easy to fix the tests only to create a `TelechatDate` for today if there is not one already, but we probably also ought to put a unique constraint on `TelechatDate.date` if we mean for that to be unique.

ietf/api/tests.py
```diff
@@ -758,6 +766,7 @@ def do_rfc_with_broken_history_test(self, draft_name):
                 content_url=rfc.get_href(),
                 name=rfc.canonical_name(),
                 previous=f'{draft.name}-10',
+                previous_url= f'https://www.ietf.org/archive/id/{draft.name}-10.txt',
```
rjsparks, line 769: It will save hassle later to pull that path out of settings instead of hardcoding it, even in the tests.

rjsparks (changes requested):

ietf/api/tests.py
kesara: Updated.

rjsparks (approved):

PR #5180, opened 2023-02-22
feat: move IETF Activity reports from external text based to HTML pages

rpcross:


ietf/iesg/views.py
```diff
@@ -528,4 +531,48 @@ def photos(request):
...
+    edate = first_day_this_month - relativedelta.relativedelta(days=1)
+    sdate = edate + relativedelta.relativedelta(day=1)
+    if request.method == 'GET':
+        form = ActivityForm(request.GET)
+        if form.is_valid():
+            month = form.cleaned_data['month']
+            year = form.cleaned_data['year']
+            sdate = datetime.date(int(year), int(month), 1)
+            next_month = sdate + relativedelta.relativedelta(months=1)
+            edate = next_month - relativedelta.relativedelta(days=1)
```
jennifer-richards, line 571: I think you have a fencepost error on your end date here and in the earlier default case. The `get_activity_stats()` method selects items with `time__gte=<midnight on sdate>` and `time__lt=<midnight on edate>`, so I think you'd want to pass 2023-02-01 and 2023-03-01 to get all of Feb 2023.

jennifer-richards (changes requested):
Comments are inline. I think there's a mustfix fencepost error on the end date handling plus some smaller stuff.

jennifer-richards (approved):

ietf/iesg/tests.py
```diff
@@ -96,6 +96,12 @@ def test_photos(self):
         ads = Role.objects.filter(group__type='area', group__state='active', name_id='ad')
         self.assertEqual(len(q('.photo')), ads.count())
         
+    def test_ietf_activity(self):
```
rjsparks, line 99: I'll take this as is, but please start making the tests more meaningful than "the page doesn't crash".

ietf/meeting/tests_views.py
```diff
@@ -7636,14 +7636,14 @@ def test_proceedings_overview(self, mock):
         response = self.client.get(url)
         self.assertContains(response, 'The Internet Engineering Task Force')
 
-    def test_proceedings_progress_report(self):
+    def test_proceedings_activity_report(self):
         make_meeting_test_data()
```
rjsparks, line 7640: Get out of the habit of using this, and build the data you need with Factories instead.

ietf/meeting/tests_views.py
```diff
@@ -7636,14 +7636,14 @@ def test_proceedings_overview(self, mock):
...
         MeetingFactory(type_id='ietf', date=datetime.date(2016,4,3), number="96")
         MeetingFactory(type_id='ietf', date=datetime.date(2016,7,14), number="97")
 
-        url = urlreverse('ietf.meeting.views.proceedings_progress_report',kwargs={'num':97})
+        url = urlreverse('ietf.meeting.views.proceedings_activity_report',kwargs={'num':97})
         response = self.client.get(url)
-        self.assertContains(response, 'Progress Report')
+        self.assertContains(response, 'Activity Report')
```
rjsparks, line 7646: Again, this test is just barely above "Page doesn't crash". We need to strive, instead, for tests that break when other code changes affect the purpose of of the views output in unexpected ways.

ietf/templates/meeting/activity_report.html
```diff
@@ -1,32 +1,22 @@
...
-        <li>
-            <h3 class="mt-3">In the final 4 weeks before meeting</h3>
-        </li>
-        <li>
-            {{ ffw_new_count|stringformat:"3s" }}    New I-Ds were received - {{ ffw_new_percent }} of total newbies since last meeting
-        </li>
-        <li>
-            {{ ffw_update_count|stringformat:"3s" }}    I-Ds were updated - {{ ffw_update_percent }} of total updated since last meeting
-        </li>
+        {% if not is_monthly_report %}
```
rjsparks, line 9: This raises the question of whether "is_monthly_report" accurately describes the condition being passed in. Reading this, I suspect what we're really trying to tell the template is "This is a meeting report". How is the parameter otherwise used?

ietf/templates/meeting/proceedings_activity_report.html
```diff
@@ -0,0 +1,18 @@
+{% extends "base.html" %}
+{% load ams_filters ietf_filters cache %}
+{% block title %}IETF {{ meeting.number }} Proceedings - Activity Report{% endblock %}
+{% block content %}
+    {% cache 3600 proceedings_activity_report meeting.number %}
```
rjsparks, line 5: I suspect this is wrong enough to need a change.
Instead of varying on meeting.number, you probably need to vary cach on sdate and edate.
Please check other places that the templates or views use the cache.

rjsparks (changes requested):

ietf/templates/meeting/activity_report.html
rpcross: Right now the distinction is binary, this template fragment is either used in a monthly activity report or a meeting activity report (spanning all time between two meetings). So "not is_monthly_report" is equivalent to "is_meeting_report". I added is_monthly_report because that was the new view for this PR. I can instead add "is_monthly_report" which is potentially more specific. Alternatively we could simply check for the existence of the meeting template variable, which isn't passed from the monthly activity report view. But if it were added later, the side effect would be unwanted.

ietf/templates/meeting/proceedings_activity_report.html
rpcross: For this template, proceedings_activity_report, meeting.number should be correct. For the monthly report, templates/iesg/ietf_activity_report.html, caching isn't installed, but it could be added, and should use sdate.

ietf/templates/meeting/activity_report.html
rpcross: Switched to use is_meeting_report in commit 9d762819d13b996da3059a05ea5ccfbe026e3947

rjsparks (approved):

PR #5184, opened 2023-02-22
feat: revise how ConsensusDocEvents are used

### LINKED ISSUE #4809: Remove knobs for changing the consensus bit for docs in the IETF stream

rjsparks:
This is intended to replace #5108.

It fixes #4809.

This leaves the ConsensusDocEvent model in place for now, but removes the ability to create new ones by editing the field in the UI. It makes it so that the IRTF chair can indicate whether consensus boilerplate can be used. For all other streams, the value is forced when a publication request is made.


larseggert (approved):

ietf/doc/views_draft.py
```diff
@@ -1250,6 +1209,46 @@ def edit_doc_extresources(request, name):
     title = "Additional document resources"
     return render(request, 'doc/edit_field.html',dict(doc=doc, form=form, title=title, info=info) )
 
+class ConsensusForm(forms.Form):
+    consensus = forms.BooleanField(required=True, label="When published as an RFC, should the consensus boilerplate be included?")
```
jennifer-richards, line 1213: With `required=True`, this field _must_ be checked or validation will fail. I don't think that's what was intended?

jennifer-richards (changes requested):
Unless it was intended that the consensus must be "True" on the edit form, need to make its field not required. Would be good to add a test to check that as well.

ietf/doc/views_draft.py
rjsparks: Gross. I had been counting on tests - points out we need a test that exercises setting the state both ways. Howeverr, walking through the experience manually says I need to rework this _again_. I'll bring it back when refactored.

rjsparks:
Closing this - the idea will need to be reworked again

PR #5192, opened 2023-02-23
fix: Use more robust filename/rev matching for meeting materials

### LINKED ISSUE #4731: slide names ending with numbers are breaking
Right now, https://datatracker.ietf.org/meeting/115/materials/slides-115-dnsop-the-alt-special-use-top-level-domain-draft-ietf-dnsop-alt-tld-18 404s. It shouldn't. 

https://datatracker.ietf.org/doc/slides-115-dnsop-the-alt-special-use-top-level-domain-draft-ietf-dnsop-alt-tld-18/ works fine, and from there, you can get to https://datatracker.ietf.org/meeting/115/materials/slides-115-dnsop-the-alt-special-use-top-level-domain-draft-ietf-dnsop-alt-tld-18-00.pdf

The issue is likely the document name ending in something that looks like a version.

jennifer-richards:
Fixes #4731.

Note that this does away with a series of moderately complex regex checks in the revision matching. I don't know the motivation behind those checks, but it seems to me that the `get_related_meeting()` check should be adequate to distinguish whether a document is suitable to be returned by this method.

### DIFF
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -202,32 +202,36 @@ def current_materials(request):
     else:
         raise Http404('No such meeting')
 
+
+def _get_materials_doc(meeting, name):
+    """Get meeting materials document named by name
+
+    Raises Document.DoesNotExist if a match cannot be found.
+    """
+    # try an exact match first
+    doc = Document.objects.filter(name=name).first()
+    if doc is not None and doc.get_related_meeting() == meeting:
+        return doc, None
+    # try parsing a rev number
+    name, rev = name.rsplit('-', 1)
+    doc = Document.objects.filter(name=name).first()
+    if doc is not None and doc.get_related_meeting() == meeting:
+        return doc, rev
+    # give up
+    raise Document.DoesNotExist
+
+
 @cache_page(1 * 60)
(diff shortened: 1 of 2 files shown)


ietf/meeting/views.py
```diff
@@ -202,32 +202,35 @@ def current_materials(request):
...
+    """Get meeting materials document named by name
+
+    Raises Document.DoesNotExist if a match cannot be found.
+    """
+    # try an exact match first
+    doc = Document.objects.filter(name=name).first()
+    if doc is not None and doc.get_related_meeting() == meeting:
+        return doc, None
+    # try parsing a rev number
+    name, rev = name.rsplit('-', 1)
```
rjsparks, line 216: This isn't sufficient. There are drafts out there with names that are prefixes of other names. You need to make sure the thing after the last - looks like two digits before trying to look up what came before.

rjsparks (changes requested):
There's also been a discussion about _disallowing_ new document names that end with -nn, and I'd be completely fine forcing that change. (Though it would irritate chairs that like to name their slides after a particular version of a draft, but it won't take long for them to adjust).

rjsparks (approved):

PR #5206, opened 2023-02-24
fix: point menu item for editoral stream to rswg documents

### LINKED ISSUE #5203: editorial refactor broke the main menu
The main menu link for the Editorial stream (/stream/editorial) now 404s.

rjsparks:
fixes #5203

### DIFF
--- a/ietf/group/tests.py
+++ b/ietf/group/tests.py
@@ -43,7 +43,7 @@ class StreamTests(TestCase):
 
         EditorialDraftFactory() # Quick way to ensure RSWG exists.
         r = self.client.get(urlreverse("ietf.group.views.stream_documents", kwargs=dict(acronym="editorial")))
-        self.assertRedirects(r, expected_url=urlreverse('ietf.group.views.group_home',kwargs={"acronym":"rswg"}),fetch_redirect_response=False)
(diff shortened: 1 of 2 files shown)


ietf/group/tests.py
```diff
@@ -41,6 +41,11 @@ def test_stream_documents(self):
         self.assertEqual(r.status_code, 200)
         self.assertContains(r, draft.name)
 
+        EditorialDraftFactory() # Quick way to ensure RSWG exists.
```
jennifer-richards, line 44: Is this needed? Since you have `fetch_redirect_response=False`, I don't think you'll actually hit the view that checks whether the group exists.

jennifer-richards (changes requested):
Not certain, but I think there's an unused factory call - suggest dropping that or, better, testing that the forwarding actually hits a valid view.

ietf/group/tests.py
rjsparks: I was fighting against the internals of `assertRedirects` since I was redirecting to something that redirected. While that might have been better (since the redirect was to the dispatcher for group), I short-circuited to what the current place the dispatcher for group sends rswg.

jennifer-richards (approved):

ietf/group/tests.py
jennifer-richards: Ok - if you think it was better as it was I don't object to that, just want to be sure that we test that the redirects will eventually land.

PR #5218, opened 2023-02-27
feat: Prefill the abstract, shepherd and AD info into a ballot writeup

larseggert:


ietf/templates/doc/mail/ballot_writeup.txt
```diff
@@ -1,11 +1,12 @@
-{% autoescape off %}
-Technical Summary
-
+{% load ietf_filters %}{% autoescape off %}Technical Summary
+{% if doc.abstract %}
+{{ abstract }}
```
rjsparks, line 3: `{{ doc.abstract.rstrip }}` should work here, saving the need for the manually split-out argument to the template.

ietf/templates/doc/mail/ballot_writeup.txt
```diff
@@ -26,12 +27,14 @@ Document Quality
    Review, on what date was the request posted?
 
 Personnel
-
+{% if doc.shepherd and doc.ad %}{% filter wordwrap:"76" %}
+   The Document Shepherd for this document is {{ doc.shepherd.person.name }}. The Responsible Area Director is {{ doc.ad.name }}.{% endfilter %}
+{% else %}
```
rjsparks, line 32: This would cause the reminder to the AD or shepherd to track the assignment of the IANA experts to be suppressed - is that intentional?

ietf/templates/doc/mail/ballot_writeup.txt
larseggert: Will do, thanks.

ietf/templates/doc/mail/ballot_writeup.txt
larseggert: I can't recall when an AD would have left that in/filled that out. Maybe that should be removed entirely.

rjsparks: wwfm

larseggert: I'll remove it.

Should I make the "IRTF Note" part conditional on the doc being on that stream? And ditto for the "IESG Note"?

And what is the "IANA Note" for?

rjsparks: Looks like IANA Note is frequently used, even now. See https://mailarchive.ietf.org/arch/browse/ietf-announce/?q=IANA%20Note for examples.

rjsparks (approved):

PR #5233, opened 2023-02-28
fix: Support time zones in agenda.txt; restore agenda-utc.txt

### LINKED ISSUE #5207: The meeting text agenda page is now only in unlabelled UTC
/meeting/agenda.txt should probably render in meeting local time.
We should probably make
/meeting/agenda-utc.txt work again.
In any case, the resulting text should be explicit about what timezone it is being displayed in.

(Right now, it displays in UTC with a "Last changed at" at the top of the page in meeting local time...)

jennifer-richards:
Fixes #5207


jennifer-richards:
Note: after merging this, the agenda note for 116 should be updated to remove the notice "Please note that times on the text agenda are in UTC."

rjsparks (approved):

PR #5254, opened 2023-03-01
fix: Don't expose existing emails via reset password and account creation forms

### LINKED ISSUE #5156: Password reset workflow reports when provided username doesn't exist
Instead, it should say nothing, and always note where to go for account creation.

larseggert:
Fixes #5156


rjsparks (approved):

rjsparks (changes requested):
Ah - needs a new test that exercises sending the new email.

jennifer-richards (changes requested):
Comments inline - one trivial, one substantive.

ietf/ietfauth/views.py
```diff
@@ -120,18 +120,21 @@ def create_account(request):
...
-            # The following is what to revert to should that lowered barrier prove problematic
-            # existing = Subscribed.objects.filter(email=to_email).first()
-            # ok_to_create = ( Allowlisted.objects.filter(email=to_email).exists()
-            #     or existing and (existing.time + TimeDelta(seconds=settings.LIST_ACCOUNT_DELAY)) < DateTime.now() )
-            # if ok_to_create:
-            #     send_account_creation_email(request, to_email)
-            # else:
-            #     return render(request, 'registration/manual.html', { 'account_request_email': settings.ACCOUNT_REQUEST_EMAIL })
+            if User.objects.filter(username=to_email).exists():
+                send_account_creation_exists_email(request, to_email)
```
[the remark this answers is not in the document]
larseggert: Please check the new revision?

ietf/ietfauth/views.py
```diff
@@ -113,35 +113,48 @@ def index(request):
...
-            to_email = form.cleaned_data['email'] # This will be lowercase if form.is_valid()
-
-            if User.objects.filter(username=to_email).exists():
-                send_account_creation_exists_email(request, to_email)
+            new_account_email = form.cleaned_data[
+                "email"
+            ]  # This will be lowercase if form.is_valid()
+
+            user = User.objects.filter(username=new_account_email).first()
+            if user:
```
rjsparks, line 126: This is the test that needs changing, not just choosing the email we send to.
We should enter the `if` code below if a matching User object exists _or_ a matching Email object exists.

rjsparks (changes requested):
I have been making changes in feat/postgres around this code, and there is a bunch that will have to change with what you're proposing now if we were to bring it into main and then merge it to postgres. It's painful, and I'm sorry to ask, but could you rebase your branch on feat/postgres and target the PR there instead?

ietf/ietfauth/views.py
```diff
@@ -113,35 +113,48 @@ def index(request):
...
+                email = (
+                    new_account_email
+                    if new_account_email in user.person.email_set.filter(active=True)
+                    else user.person.email_set.filter(active=True).order_by("-time").first()
```
jennifer-richards, line 130: Suggest ordering with `.order_by("-primary", "-time")` to prefer the primary email address

jennifer-richards: Also, we should not mail to `new_account_email` once it has matched a `user.username` or `user.person.email_set` entry. It opens the door to unicode spoofing.

larseggert: Should we always email `user.person.email_set.filter(active=True).order_by("-primary", "-time").first()` in that case?

ietf/ietfauth/views.py
larseggert: So something like this?
```python
            existing_user = User.objects.filter(username=new_account_email).first()
            existing_email = Email.objects.filter(email=new_account_email).first()
            if existing_user or existing_email:
                email = (
                    existing_user.person.email_set.filter(active=True).order_by("-primary", "-time").first()
                    if existing_user
                    else new_account_email
                )
                send_account_creation_exists_email(request, new_account_email, email)
```

ietf/ietfauth/views.py
rjsparks: Use `person.email()` instead of copying that logic everywhere please.
See
https://github.com/ietf-tools/datatracker/blob/1d87686a2bae67db0d45f6fea909c0e73856571c/ietf/person/models.py#L140-L152

jennifer-richards (approved):
lgtm, but @rjsparks mentioned wanting the PR pointed at feat/postgres

larseggert:
Yep, that is next. Wanted to make sure the basic logic is OK now though.

ietf/ietfauth/views.py
```diff
@@ -113,32 +113,49 @@ def index(request):
...
-            #     send_account_creation_email(request, to_email)
-            # else:
-            #     return render(request, 'registration/manual.html', { 'account_request_email': settings.ACCOUNT_REQUEST_EMAIL })
+            new_account_email = form.cleaned_data[
+                "email"
+            ]  # This will be lowercase if form.is_valid()
+
+            existing_user = User.objects.filter(username=new_account_email).first()
+            existing_email = Email.objects.filter(address=new_account_email).first()
+            if existing_user or existing_email:
```
rjsparks, line 127: This is fine as is, but it's a good opportunity to point to some readability options for the future.

You could also have left off the .first() in L125 and L126 and at L127 used:
```
if existing_user.exists() or existing_email.exists():
```

It's _slightly_ more than readability, but not enough to warrant another change in this code. The difference is that the database can use its index to say whether there's anything to return without actually loading the thing into memory. Impact would be larger if the the object being loaded is complex.

larseggert:
I messed up the rebase and had to do a clean PR. See #5288.

PR #5260, opened 2023-03-02
feat: Add calendar view to "important dates" page

### LINKED ISSUE #3512: Calendar view of important dates
`type_enhancement`   |    by cmorgan@amsl.com

___

Hi!

Would it be possible to get a calendar view on the important dates page (https://datatracker.ietf.org/meeting/important-dates/), similar to the one on the upcoming meetings page (https://datatracker.ietf.org/meeting/upcoming)?

Thanks!
Cindy

___
_Issue migrated from trac:3512 at 2022-03-04 09:24:40 +0000_

larseggert:
Fixes #3512


rjsparks (approved):
It would be good to show the actual meeting on this calendar, but we can add that later.

PR #5274, opened 2023-03-03
chore: Enable dependabot version updates

larseggert:


NGPixel:
Might be a good idea to change the schedule from weekly to monthly? It could get out of hand very quickly with the number of projects.

larseggert:
I guess it comes down to whether you prefer a few PRs each week or a lot of them each month :-) They should all be pretty straightforward though.

NGPixel (approved):

PR #5281, opened 2023-03-03
refactor: Tie Meetecho resources to Session pk

jennifer-richards:
There are two related parts:
* Replace the `UrlResource` representation of video/audio streams and the onsite tool with methods on `Session`. The URLs contain only the meeting number and session pk.
* Change the API views `api_set_session_video_url()` and `api_upload_bluesheet()` to identify sessions using pk.

This is, I believe, ready to go. Leaving as a draft because we don't plan to deploy until after IETF-116.


jennifer-richards:
This should also wait until after the feat/postgres merge. Will at least need to renumber the migrations.

jennifer-richards:
I believe this is now updated and ready to go

rjsparks (approved):

rjsparks:
@alexamirante - we just merged this and it will be part of the next release. Until you change to use the newer parameters, the older parameters will work. Please let us know when you've changed everything over (after the release) so we can remove the deprecated code.

PR #5287, opened 2023-03-07
fix(agenda): replace download cal dropdown with actual links

### LINKED ISSUE #3811: Confusing UI for agenda webcal subscribe/ics download buttons
The agenda page's UI for subscribing to webcal or downloading ics files looks something like:

image

(see #3786). This is shared between the main "agenda" tab and the "personalize" tab that permits filtering as well as selecting individual sessions to appear on the agenda.

These are a bunch of similar buttons: an ics file can be downloaded for a specific area, for all the "non-area" events, or for the currently filtered set of sessions. The latter can also be "subscribed" to as a webcal link.

The filtered calendars are only offered when there are filter parameters selected. These can also represent "filtered" views selected by the UI above these links, or "personal" views selected by combining the filtered views with individually selected sessions on the Personalize tab.

The current set of buttons is a bit haphazard. Really need to rationalize this and offer a more coherent presentation of these options. Just visually, the button sets are different sizes/styles and split across lines in a way that's not of obvious significance to a user.

NGPixel:
Replaces the agenda calendar download dropdown menu with actual links, so that they can be copied.

Fixes #3811

### DIFF
--- a/client/agenda/AgendaQuickAccess.vue
+++ b/client/agenda/AgendaQuickAccess.vue
@@ -74,7 +74,6 @@
           size='large'
           :show-arrow='true'
           trigger='click'
-          @select='downloadIcs'
           )
           n-button.mt-2(
             id='agenda-quickaccess-addtocal-btn'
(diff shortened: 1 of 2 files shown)


jennifer-richards (approved):

rjsparks:
Shouldn't a test have changed?

rjsparks:
> Shouldn't a test have changed?

Nevermind - I see it now.

rjsparks (approved):

PR #5288, opened 2023-03-07
fix: Don't expose existing emails via reset password and account creation forms

larseggert:
Replaces #5254. Rebase to feat/postgres.


ietf/ietfauth/views.py
```diff
@@ -112,33 +112,47 @@ def index(request):
...
-            # else:
-            #     return render(request, 'registration/manual.html', { 'account_request_email': settings.ACCOUNT_REQUEST_EMAIL })
+            new_account_email = form.cleaned_data[
+                "email"
+            ]  # This will be lowercase if form.is_valid()
+
+            user = User.objects.filter(username__iexact=new_account_email)
+            email = Email.objects.filter(address__iexact=new_account_email)
+            if user.exists() or email.exists():
+                to_email = user.first().person.email_address() if user else new_account_email
```
rjsparks, line 129: This could still leave us sending mail to the address provided in the form. Instead, I think we want the else here to be email.first().person.email_address()?

rjsparks (changes requested):

rjsparks:
_Thank you_ for the rebase effort

ietf/ietfauth/views.py
larseggert: Wouldn't that be the same email?

jennifer-richards: Nothing is easy... 

* even if `user is not None`, `user.email_address()` might be `''`
* an `Email` might exist but be disabled

I think we probably want `person_to_contact = user.person if user else email.person.user`. Then we contact `person.email_address()` if that's not empty, and finally throw it at an admin to deal with it because something fishy is going on.

jennifer-richards: > Wouldn't that be the same email?

Maybe not (e.g., https://0xsha.io/blog/the-fall-of-mighty-django-exploiting-unicode-case-transformations#exploiting-unicode-case-transformations).

The main thing here is that we want to alert a validated address that this attempt has been made, which means not trusting the POSTed value.

(Though I suspect a collision like this would mean an active attempt at hijacking. In that case, we're not really preventing an information leak any more, just trying to be sure a Real Person is informed.)

larseggert: Laptop died and am waiting for a replacement. I can pick this up later, or you can make the change if you need to merge before that.

rjsparks (approved):

jennifer-richards (approved):
Looks ok - I think it will leak the existence of an email / username in the case that the user has no active email addresses, but I think that's probably not an exploitable situation.

PR #5295, opened 2023-03-12
feat: endpoint for imapd to authenticate against

rjsparks:


rjsparks:
Feel free to suggest a different direction for this.

jennifer-richards (changes requested):
I think we need to fail if `User.objects.filter(username__iexact=username).count() > 1` (equivalently, if `User.objects.get(username__iexact=username)` raises `User.MultipleObjectsReturned`). Otherwise, if two `User` records are somehow created with a username collision under `iexact`, whichever `User` comes up as `first()` can log in using either username. Assuming imapd is simply passing the username over as it receives it, it would then allow that password holder to access either account.

This is different from the cases we've dealt with before. If we want to treat it the same way, we would need to report back to imapd not only "success" but the actual username we used in deciding it was a success.

jennifer-richards (approved):

VeselinPersinov (commented):
Hi nice to meet you

PR #5356, opened 2023-03-16
chore: change the way migrations are invoked for actions and dev builds

rjsparks:


NGPixel (approved):

jennifer-richards (approved):
Looks good. The INSTALL instructions are not changed here. If this change is not intended to be temporary then might want to update that.

rjsparks:
I think we have to play INSTALL by ear for the next deploy, then evaluate pulling these back out (that is, I suspect these are temporary).

Issue #4641, opened 2022-10-24
wording a bit off in "you've been nominated"


richsalz:
### Describe the issue

The nomination questionnaires are sent as part of the "you've been nominated" email. The text in that email implies they'll get them if/when they accept.  We should fix that.  And also fix the wording to allow for times when NomCom hasn't finished the questionnaires, maybe "if you do not see the questionnaire below, it will be available at shortly on the nomcom page, https://....."

See also #4637 

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

richsalz:
Quoting email I have from Robert:
"We could tweak the text in the nomination notification message to help avoid confusion, but the current implementation sends the questionaire at the same time as  the nomination notice when the nomcom is configured to send questionnaires automatically. The theory is that people should get a chance to start in on the questionaire while they're still working out with their sponsor whether or not they can accept the nomination.

The relevant code is here:

https://github.com/ietf-tools/datatracker/blob/8.18.0/ietf/nomcom/utils.py#L363
"

LPardue:
Sending the questionnaire at the same time is fine by me, so the suggestion of changing the text in the nomination announcement is reasonable. Just make it clearer like: "we've sent you the questionnaire at the same time. There's no obligation to complete or return it until you accept the nomination"

rjsparks:
We could look at the configuration of the nomcom at the time the message is sent, and say either "You will receive another message with the questionnaire immediately. There's no obligation to complete or return it until you accept the nomination." or use the current text if the questionnaire won't be sent until some other point in the future.

richsalz:
The proposed fix (#5357) changes the prompt to tell the nomcom chair to NOT click the "generate questionnaire email" button if the questionnaires are not ready.  That seems good enough.

richsalz:
I think this can be closed as fixed by #5357

PR #5357, opened 2023-03-16
fix: Edit "You've been nominated" text (#4641)

richsalz:
Mention that the schedule is posted to ietf-announce and give a link to subscribe to that.  Use vague wording to handle all cases.


richsalz:
Sorry for the noise; forced push/rebase etc.  Learning how to submit PR's here.

ietf/dbtemplate/fixtures/nomcom_templates.xml
rjsparks, line 1: This file is essentially legacy - it was used to populate the database long ago, and is not used by e.g. the tests. 

Changes here will not affect production.
What you really want is a migration that changes a row in the database. I can show you how to build that.

ietf/nomcom/migrations/0001_initial.py
rjsparks, line 1: This migration has long ago been run. Changing it here has no effect on production. Please revert this file in your branch.

ietf/dbtemplate/fixtures/nomcom_templates.xml
rjsparks: After we create the migration you want, (or edit the template in the database directly), you'll need to revert this file in your branch

ietf/nomcom/models.py
```diff
@@ -48,7 +48,7 @@ class NomCom(models.Model):
 
     group = ForeignKey(Group)
     send_questionnaire = models.BooleanField(verbose_name='Send questionnaires automatically', default=False,
-                                             help_text='If you check this box, questionnaires are sent automatically after nominations.')
+                                             help_text='If you check this box, questionnaires are sent automatically after nominations. DO NOT CHECK if they are not ready yet.')
```
rjsparks, line 51: This is fine, but it also needs a new companion migration.

rjsparks (changes requested):

ietf/nomcom/migrations/0001_initial.py
rjsparks: (again, hold until we create the migration you really wanted so we can translate your suggestion into the migration.)

ietf/nomcom/migrations/0001_initial.py
```diff
@@ -50,7 +50,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('public_key', models.FileField(blank=True, null=True, storage=ietf.utils.storage.NoLocationMigrationFileSystemStorage(location=None), upload_to=ietf.nomcom.models.upload_path_handler)),
-                ('send_questionnaire', models.BooleanField(default=False, help_text='If you check this box, questionnaires are sent automatically after nominations.', verbose_name='Send questionnaires automatically')),
+                ('send_questionnaire', models.BooleanField(default=False, help_text='If you check this box, questionnaires are sent automatically after nominations. DO NOT CHECK if they are not ready yet.', verbose_name='Send questionnaires automatically')),
```
jennifer-richards, line 53: Should add a new migration instead of modifying this one

rjsparks (approved):

rjsparks:
holding to bring in after the postgres transition is complete

richsalz:
I think this is ready for review now.

PR #5358, opened 2023-03-16
fix: Avoid creating a .yml-r file

richsalz:
These seem to be created if initial setup fails. Ignore them.

Not sure if this is something you want or not.

### DIFF
--- a/docker/run
+++ b/docker/run
@@ -41,8 +41,9 @@ done
 rm -rf .parcel-cache __pycache__
 
 # Create extended docker-compose definition
-cp docker-compose.extend.yml docker-compose.extend-custom.yml
-sed -i -r -e "s/CUSTOM_PORT/$CUSTOM_PORT/" docker-compose.extend-custom.yml
+sed -e "s/CUSTOM_PORT/$CUSTOM_PORT/" \
+	<docker-compose.extend.yml docker-compose.extend-custom.yml \
+	>docker-compose.extend-custom.yml
 cd ..
 
 # Set UID/GID mappings
--- a/docker/run
+++ b/docker/run
@@ -42,7 +42,7 @@ rm -rf .parcel-cache __pycache__
 
 # Create extended docker-compose definition
 sed -e "s/CUSTOM_PORT/$CUSTOM_PORT/" \
-	<docker-compose.extend.yml docker-compose.extend-custom.yml \
+	<docker-compose.extend.yml \
 	>docker-compose.extend-custom.yml
 cd ..
 


rjsparks:
Are you sure this isn't a side-effect of an editor you are using touching the file?
I haven't ever seen such a thing.
If you can reproduce (or guess at what led to) the initial setup failure, it might help with assessing.

richsalz:
I'll close this.  It came about, I think, when I didn't have 'docker compose' and the initial setup failed.  Not worth it.

rjsparks:
Other people at the sprint are running into this - ignoring the file won't hurt us, but it would be good to know what is creating it.

rjsparks:
We need to add code to the run script that detects the version of sed.
For people getting sed from os/x, the -i or -I takes an argument, and in this case it's eating -r as the argument
```
rjsparks@dhcp-80c4 foo % ls
rjsparks@dhcp-80c4 foo % echo "This is a test" > a
rjsparks@dhcp-80c4 foo % sed -i -r -e "s/ a / the /" a
rjsparks@dhcp-80c4 foo % ls -al
total 16
drwxr-xr-x    4 rjsparks  staff   128 Mar 25 14:44 .
drwxr-xr-x  236 rjsparks  staff  7552 Mar 25 14:43 ..
-rw-r--r--    1 rjsparks  staff    17 Mar 25 14:44 a
-rw-r--r--    1 rjsparks  staff    15 Mar 25 14:44 a-r
```
From `man sed` on osx:
```
     -I extension
             Edit files in-place, saving backups with the specified extension.
             If a zero-length extension is given, no backup will be saved.  It
             is not recommended to give a zero-length extension when in-place
             editing files, as you risk corruption or partial content in
             situations where disk space is exhausted, etc.

             Note that in-place editing with -I still takes place in a single
             continuous line address space covering all files, although each
             file preserves its individuality instead of forming one output
             stream.  The line counter is never reset between files, address
             ranges can span file boundaries, and the “$” address matches only
             the last line of the last file.  (See Sed Addresses.) That can lead
             to unexpected results in many cases of in-place editing, where
             using -i is desired.

     -i extension
             Edit files in-place similarly to -I, but treat each file
             independently from other files.  In particular, line numbers in
             each file start at 1, the “$” address matches the last line of the
             current file, and address ranges are limited to the current file.
             (See Sed Addresses.) The net result is as though each file were
             edited by a separate sed instance.
```

docker/run
```diff
@@ -41,8 +41,9 @@ done
 rm -rf .parcel-cache __pycache__
 
 # Create extended docker-compose definition
-cp docker-compose.extend.yml docker-compose.extend-custom.yml
-sed -i -r -e "s/CUSTOM_PORT/$CUSTOM_PORT/" docker-compose.extend-custom.yml
+sed -e "s/CUSTOM_PORT/$CUSTOM_PORT/" \
+	<docker-compose.extend.yml docker-compose.extend-custom.yml \
```
cabo, line 45:
```suggestion
	<docker-compose.extend.yml \
```

rjsparks (approved):

PR #5362, opened 2023-03-17
fix: Show "interesting" doc relations on HTMLized page

### LINKED ISSUE #5344: /doc/html replaced drafts
### Description

A replaced draft on /doc shows what it was replaced with.

A draft that replaces another shows what it replaced on /doc/html.  That is fine, though it only does so via the "Compare versions" feature.

A draft that has been replaced does not show what replaced it anywhere on /doc/html.

It seems like this might be a useful feature when looking at a draft, especially for drafts that have been adopted, where you might want to see the latest.

Story: I was trying to find my way from https://datatracker.ietf.org/doc/html/draft-xu-risav to the latest version.  I ended up going to GitHub (which the draft conveniently links) to find what replaced the draft (though the /doc link might have been faster).

### Code of Conduct

larseggert:
Fixes #5344


ietf/doc/views_doc.py
```diff
@@ -139,20 +139,17 @@ def render_document_top(request, doc, tab, name):
 def interesting_doc_relations(doc):
 
     if isinstance(doc, Document):
-        cls = RelatedDocument
```
rjsparks, line 142: This is here for a reason?
You will report the wrong relations if you are looking at an older version of a document if you take this out.

rjsparks:
Holding this for discussion at codesprint

rjsparks:
Bah - we missed talking about this Saturday - help watch for a chance to talk about it while here this week

rjsparks:
We should talk about this during 117.

Limpisey77922321 (commented):
[Build and Release](https://github.com/ietf-tools/datatracker/actions/workflows/build.yml)

larseggert:
Converting to draft until we have a chance to talk.

rjsparks:
closing this as asleep - it can be pulled back out later if needed.

PR #5363, opened 2023-03-17
fix: Add reviewer name to review assignment summary

### LINKED ISSUE #5360: Show the name of the directorate reviewer for drafts
### Description

I'm looking at the directorate reviews for https://datatracker.ietf.org/doc/draft-ietf-6man-rfc6874bis/ (which happens to have a nice collection of options) and I found myself wondering "who said that?"  Then I realized that it was me.  The current format is "$DIRECTORATE $stage review (of -$vv)? ($REVIEW_OUTCOME)".  Maybe adding "by $PERSON" might help contextualize in some cases.

### Code of Conduct

larseggert:
Fixes #5360

### DIFF
--- a/ietf/templates/doc/review_assignment_summary.html
+++ b/ietf/templates/doc/review_assignment_summary.html
@@ -8,7 +8,7 @@
                     {{ review_assignment.review_request.doc.name }}
                 {% endif %}
                 -{{ review_assignment.reviewed_rev }}){% endif %}
-            by {{ review_assignment.reviewer.person.name }}</a>
+            by {{ review_assignment.reviewer.person.plain_name() }}</a>
         {% if review_assignment.state_id == "part-completed" %}
             <span class="badge rounded-pill ms-1 bg-secondary">Partially completed</span>
         {% endif %}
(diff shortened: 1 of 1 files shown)


ietf/templates/doc/review_assignment_summary.html
```diff
@@ -7,7 +7,8 @@
                 {% if review_assignment.review_request.doc.name != current_doc_name %}
                     {{ review_assignment.review_request.doc.name }}
                 {% endif %}
-                -{{ review_assignment.reviewed_rev }}){% endif %}</a>
+                -{{ review_assignment.reviewed_rev }}){% endif %}
+            by {{ review_assignment.reviewer.person.name }}</a>
```
rjsparks, line 11: Consider using plain_name here instead?

larseggert: Can do, but why?

rjsparks: Because that's what we use fairly consistently everywhere else on pages like this. And it would be strange when someone's name appears twice on a page (for different reasons) to see a different form of the name.

I use this anecdote to help figure out which to use:
```
>>> ba = Person.objects.get(name__endswith=" Aboba")
>>> ba.plain_name()
'Bernard Aboba'
>>> ba.name
'Dr. Bernard D. Aboba'
```

rjsparks (approved):

PR #5389, opened 2023-03-21
fix: Show area of WG as part of document info

### LINKED ISSUE #1937: Make it easy to see what area (and group) an RFC came out of
`keyword_sprint` `type_defect`   |    by rjsparks@nostrum.com

___

The main document page should make it clearer what _area_ a document was published from, when it was published from an area. It currently indicates the group, but only points to the current notion of the group as far as the datatracker is concerned.

This is problematic with groups that have closed and reopened, possibly in different areas (see XMPP), groups that get renamed, and groups that move between areas (especially because of area restructuring like the recent creation of ART).

The main page should be sensitive to the doc being published as an RFC, and look at the history at the time of publication to speak about the group and area.

___
_Issue migrated from trac:1937 at 2022-03-04 04:56:23 +0000_

larseggert:
Fixes #1937

### DIFF
--- a/ietf/doc/views_doc.py
+++ b/ietf/doc/views_doc.py
@@ -303,7 +303,7 @@ def document_main(request, name, rev=None, document_html=False):
             submission = "individual in %s area" % group.acronym
         else:
             if group.features.acts_like_wg and not group.type_id=="edwg":
-                submission = "%s %s" % (group.acronym, group.type)
+                submission = "%s %s, %s %s" % (group.acronym.upper(), group.type, group.parent.acronym.upper(), group.parent.type)
             else:
                 submission = group.acronym
             submission = "<a href=\"%s\">%s</a>" % (group.about_url(), submission)


ietf/doc/views_doc.py
```diff
@@ -303,7 +303,7 @@ def document_main(request, name, rev=None, document_html=False):
             submission = "individual in %s area" % group.acronym
         else:
             if group.features.acts_like_wg and not group.type_id=="edwg":
-                submission = "%s %s" % (group.acronym, group.type)
+                submission = "%s %s, %s %s" % (group.acronym.upper(), group.type, group.parent.acronym.upper(), group.parent.type)
```
rjsparks, line 306: Not safe - not all groutypes with acts_like_wg will necessarily have parents. The extension needs to be guarded with "does the parent exist?"

rjsparks (changes requested):

larseggert:
Am fine with just closing.

rjsparks:
Lets close it then.

PR #5391, opened 2023-03-21
fix: Don't (re)sort obsoleted/updated info

### LINKED ISSUE #2441: Wrongly sorted Updates list for RFCs on the search result page
`type_defect`   |    by magnus.westerlund@ericsson.com

___

I searched for RFC 792 by entering just "792" in the search field at the right top of the tracker. Thus resulting in the result page:

https://datatracker.ietf.org/doc/search/?name=792&activedrafts=on&rfcs=on

Looking at the entry for RFC 792 at this page I gets an list of RFC that updates it: "Updated by RFC4884, RFC6633, RFC6918, RFC950"

As you can see that is wrongly sorted as RFC950 comes last instead of first. 
This is however correct looking at the RFC792 page:
https://datatracker.ietf.org/doc/rfc792/ where the list is given as:
Updated by RFC 950, RFC 4884, RFC 6633, RFC 6918

It would be good to get the same sorting in the search results as for the individual RFC page.

___
_Issue migrated from trac:2441 at 2022-03-04 06:38:15 +0000_

larseggert:
Fixes #2441


ietf/doc/utils_search.py
```diff
@@ -150,7 +150,6 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
         elif rel.relationship_id == "updates":
             l = d.updated_by_list
         l.append(rel_rfc_aliases[rel.source_id])
-        l.sort()
```
rjsparks, line 153: This sort was in an odd place anyhow (repeatedly sorting a list as things were appended to it), but _not_ sorting is likely not safe - you are relying on the natural order coming back from `RelatedDocument.objects.filter` and that ordering is undefined (casual tests may show it returning things in the order they were inserted into the RelatedDocument table, but that is not guaranteed to remain true).

rjsparks (changes requested):

ietf/doc/utils_search.py
rjsparks: an order_by on the query might be sufficient.

larseggert: So `interesting_doc_relations` also doesn't seem to sort, which is what we use elsewhere. Should it sort? What do I pass into `order_by`, I wasn't able to figure it out?

larseggert:
@rjsparks I think I addressed your review?

rjsparks:
in the backlog of the post-ietf week away, the transition, and now the tools retreat. I'll process these (and other outstanding PRs as time allows during the retreat and no later than early next week.

larseggert:
@rjsparks recheck?

ietf/doc/utils_search.py
```diff
@@ -143,14 +143,15 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
     xed_by = RelatedDocument.objects.filter(target__name__in=list(rfc_aliases.values()),
                                             relationship__in=("obs", "updates")).select_related('target')
     rel_rfc_aliases = dict([ (a.document.id, re.sub(r"rfc(\d+)", r"RFC \1", a.name, flags=re.IGNORECASE)) for a in DocAlias.objects.filter(name__startswith="rfc", docs__id__in=[rel.source_id for rel in xed_by]) ])
+    l = []
```
rjsparks, line 146: This doesn't work. l is a temporary that is switched between two things inside the loop below. The assignment here is ignored, and the sort as written happens on whatever list l got assigned to last (leaving the other unsorted).

ietf/doc/utils_search.py
```diff
@@ -143,14 +143,15 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
     xed_by = RelatedDocument.objects.filter(target__name__in=list(rfc_aliases.values()),
                                             relationship__in=("obs", "updates")).select_related('target')
```
rjsparks, line 144:
```suggestion
    # Revisit this block after RFCs become first-class Document objects
    xed_by = list(RelatedDocument.objects.filter(
        target__name__in=list(rfc_aliases.values()),
        relationship__in=("obs", "updates")
    ).select_related('target'))
```

ietf/doc/utils_search.py
```diff
@@ -143,14 +143,15 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
...
     rel_rfc_aliases = dict([ (a.document.id, re.sub(r"rfc(\d+)", r"RFC \1", a.name, flags=re.IGNORECASE)) for a in DocAlias.objects.filter(name__startswith="rfc", docs__id__in=[rel.source_id for rel in xed_by]) ])
```
rjsparks, line 145:
```suggestion
    rel_rfc_aliases = dict([ (a.document.id, re.sub(r"rfc(\d+)", r"RFC \1", a.name, flags=re.IGNORECASE)) for a in DocAlias.objects.filter(name__startswith="rfc", docs__id__in=[rel.source_id for rel in xed_by]) ])
    xed_by.sort(key=lambda rel: int(re.sub(r"rfc\s*(\d+)", r"\1", rel_rfc_aliases[rel.source_id], flags=re.IGNORECASE)))
```

ietf/doc/utils_search.py
```diff
@@ -143,14 +143,15 @@ def fill_in_document_table_attributes(docs, have_telechat_date=False):
...
+    l = []
     for rel in xed_by:
         d = doc_dict[rel.target.document.id]
         if rel.relationship_id == "obs":
             l = d.obsoleted_by_list
         elif rel.relationship_id == "updates":
             l = d.updated_by_list
         l.append(rel_rfc_aliases[rel.source_id])
-        l.sort()
+    l.sort(key=lambda x: int(re.sub(r"rfc\s*(\d+)", r"\1", x, flags=re.IGNORECASE)))
```
rjsparks, line 154:
```suggestion
```

ietf/doc/utils_search.py
rjsparks: I've made some complicated suggestions just typing into github's UI - there may be issues with them.

rjsparks (changes requested):

ietf/doc/utils_search.py
rjsparks: In case the UI suggest feature doesn't result in what I meant, the goal is to not have any sort statement inside or after the loop.

ietf/doc/utils_search.py
rjsparks: All of this will become much easier to do (the database will be able to do it for us) after we make RFCs first-class Document objects and the database knows rfc numbers as integers.

rjsparks (approved):

PR #5400, opened 2023-03-25
chore: Update test docs to postgresql; remove extraneous sqlite stuff

jimfenton:


docker/scripts/app-init.sh
```diff
@@ -54,17 +54,6 @@ else
...
-    echo "Setting up a default settings_local_sqlitetest.py ..."
-    cp $WORKSPACEDIR/docker/configs/settings_local_sqlitetest.py $WORKSPACEDIR/ietf/settings_local_sqlitetest.py
-else
-    echo "Using existing ietf/settings_local_sqlitetest.py file"
-    if ! cmp -s $WORKSPACEDIR/docker/configs/settings_local_sqlitetest.py $WORKSPACEDIR/ietf/settings_local_sqlitetest.py; then
-        echo "NOTE: Differences detected compared to docker/configs/settings_local_sqlitetest.py!"
-        echo "We'll assume you made these deliberately."
-    fi
-fi
-
```
rjsparks, line 67: @NGPixel I don't think any devs used settings_local_sqlitetest. Did and actions? We don't have a comparable settings_local_postgrestest, and I suspect we don't need one?

ietf/.gitignore
```diff
@@ -2,8 +2,8 @@
 /settings_local.py
 /settings_local.py.bak
 /settings_local_debug.py
-/settings_local_sqlitetest.py
 /settings_local_vite.py
 /settings_mysqldb.py
 /settings_postgresqldb.py
+/settings_postgrestest.py
```
rjsparks, line 8: This file is in git and does not need to be ignored. Please revert this change

rjsparks (changes requested):

docker/scripts/app-init.sh
NGPixel: @rjsparks It is used in some workflows, e.g.:

https://github.com/ietf-tools/datatracker/blob/main/.github/workflows/build.yml#L163-L165

ietf/.gitignore
jimfenton: Reverted by https://github.com/ietf-tools/datatracker/pull/5400/commits/012de38aaee39d4b19928717c151213066527388

docker/scripts/app-init.sh
jimfenton: But that's in the main branch, not feat/postgres. The corresponding lines are:

https://github.com/ietf-tools/datatracker/blob/feat/postgres/.github/workflows/build.yml#L156-L158

which use settings_postgrestest, not settings_local_sqlitetest

rjsparks:
Jim - I think something went wrong with a rebase here - can could look and clean it back to the intended changes?

jimfenton:
Yes, I did a sync of my fork to the main repository and it seems to have done bad things. I haven't run into that before, probably because I wasn't working on such an active project.

Let me see if I can fix it up; otherwise I'll just close this PR and create a new one.

jimfenton:
@rjsparks With help from Justin Richer, I got this pull request cleaned up. Please take another look.

ietf/.gitignore
rjsparks: settings_postgresqldb.py is source-controlled in docker/configs/, and is copied into ietf/ as the dev environment is built - it's runtime configuration for the actual app, not the tests. settings_postgrestest.py is source-controlled in ietf/.

That schism is an artifact of history, and we could normalize what we're doing, but that's a bigger task than this PR should take on.

rjsparks (approved):

Issue #4892, opened 2022-12-14
Format tweak to CSV download of volunteers


richsalz:
### Description

It would be useful to change the `"path_1, path_2"` syntax to avoid quotes so that Unix tools could more easily parse the output.  I suggest something like `p1+p2` but don't care much.

This may be a duplicate, sorry.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #5401, opened 2023-03-25
fix: Make "nomcom paths" field CSV-friendly (#4892)

richsalz:
When generating the NomCom volunteer paths, use a plus sign rather than a comma, so that the output is more CSV-friendly.

I decided not to change `path_N` into `pN` as that is more a matter of style than really required and can easily
by changed as needed by global replace. (Going the other way `pN` to `path_N` is trickier.)


rjsparks (approved):

rjsparks:
cut-paste to csv is fraught with peril. Making a "download as cvs" view would be better.

rjsparks:
(to be clear, I'll merge this as is (assuming the tests pass))

Issue #4727, opened 2022-11-07
Add "totals" to "view feedback" page


richsalz:
### Description

It would be useful to add totals for each column in the "view feedback" page.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

Issue #4627, opened 2022-10-21
ability to create a "test" nominee


richsalz:
### Description

It can be useful to create multiple "test" nominees (a bit that could only be set by the nomcom chair perhpas) to test that the workflow is working.  Biggest difference is that nominees with the "test" bit do not show up in the feedback list, or perhaps are marked as "TEST" somehow.

RFC 8713 says that not all nominees need to be made public. Does the tooling support that? If so, then using whatever that mechanism is there would probably handle this as well.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #5406, opened 2023-03-25
fix: Add column totals to feedback (#4727)

richsalz:
Create a new aggregate parameter that holds the totals and pass that into the HTML file as a parameter for display.
Not easy to test given #4627.

### DIFF
--- a/ietf/nomcom/views.py
+++ b/ietf/nomcom/views.py
@@ -782,12 +782,14 @@ def view_feedback(request, year):
 
     sorted_nominees = sorted(nominees,key=lambda x:x.staterank)
 
+    fbtotals = []
     for nominee in sorted_nominees:
         last_seen = FeedbackLastSeen.objects.filter(reviewer=request.user.person,nominee=nominee).first()
         nominee_feedback = []
         for ft in nominee_feedback_types:
             qs = nominee.feedback_set.by_type(ft.slug)
             count = qs.count()
+            fbtotals[ft.slug] = count + fbtotals.get(ft.slug, 0);
             if not count:
                 newflag = False
             elif not last_seen:
@@ -822,6 +824,7 @@ def view_feedback(request, year):
                                'topics_feedback': topics_feedback,
                                'independent_feedback': independent_feedback,
                                'nominees_feedback': nominees_feedback,
+                               'feedback_totals': fbtotals,
                                'nomcom': nomcom})
 
 
--- a/ietf/nomcom/views.py
+++ b/ietf/nomcom/views.py
@@ -782,7 +782,7 @@ def view_feedback(request, year):
 
     sorted_nominees = sorted(nominees,key=lambda x:x.staterank)
 
-    fbtotals = []
+    fbtotals = defaultdict(lambda: 0)
     for nominee in sorted_nominees:
         last_seen = FeedbackLastSeen.objects.filter(reviewer=request.user.person,nominee=nominee).first()
         nominee_feedback = []
--- a/ietf/nomcom/views.py
+++ b/ietf/nomcom/views.py
@@ -4,7 +4,7 @@
 
 import datetime
 import re
-from collections import OrderedDict, Counter
+from collections import OrderedDict, Counter, defaultdict
 import csv
 import hmac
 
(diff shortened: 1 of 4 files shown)


rjsparks:
Better to test it with something in nomcom/tests.py

ietf/templates/nomcom/view_feedback.html
```diff
@@ -54,6 +54,12 @@ <h3 class="mt-5" id="declined">Declined each nominated position</h3>
                     </tr>
                 {% endfor %}
             </tbody>
+	    <tr>
+		<td></td>
+		<td></td>
+		<td><b>Totals</b></td>
+		{% for fbtot in feedback_totals %}<td>{{ fbtot }}</td>{% endfor %}
+	    </tr>
```
larseggert, line 62: Should this be in a `tfoot`?

rjsparks: yes, please make that change, consider `<th>` instead of the explicit `<b>`

ietf/nomcom/views.py
```diff
@@ -782,12 +782,14 @@ def nominee_staterank(nominee):
 
     sorted_nominees = sorted(nominees,key=lambda x:x.staterank)
 
+    fbtotals = []
```
rjsparks, line 785: This is a list, but below you treat it as a dict.
What you want here is a defaultdict.
```
>>> from collections import defaultdict
>>> totals=defaultdict(lambda:0)
>>> totals['foo']
0
>>> totals['bar']+=1
>>> totals['bar']
1
```

rjsparks (changes requested):

ietf/templates/nomcom/view_feedback.html
richsalz: done.

rjsparks:
This didn't survive the automated change of branch target - I can work to straighten it out, or, if your local branch rebases easily, perhaps submit this again? (with apologies for the hassle).

richsalz:
I'll figure something out; you have enough to do.

richsalz:
fixed, please take a look.

ietf/nomcom/views.py
```diff
@@ -782,12 +782,14 @@ def nominee_staterank(nominee):
...
     sorted_nominees = sorted(nominees,key=lambda x:x.staterank)
 
+    fbtotals = defaultdict(lambda: 0)
     for nominee in sorted_nominees:
         last_seen = FeedbackLastSeen.objects.filter(reviewer=request.user.person,nominee=nominee).first()
         nominee_feedback = []
         for ft in nominee_feedback_types:
             qs = nominee.feedback_set.by_type(ft.slug)
             count = qs.count()
+            fbtotals[ft.slug] = count + fbtotals.get(ft.slug, 0);
```
rjsparks, line 792:
```suggestion
            fbtotals[ft.slug] += count
```

rjsparks:
I think URL mapping was about a different PR?

ietf/templates/nomcom/view_feedback.html
```diff
@@ -54,6 +54,14 @@ <h3 class="mt-5" id="declined">Declined each nominated position</h3>
                     </tr>
                 {% endfor %}
             </tbody>
+	    <tfoot>
+		<td></td>
+		<td></td>
+		<th>Totals</h></th>
+		{% for ft in nominee_feedback_types %}
+		  <td>{{ feedback_totals|lookup:ft }}</td>
```
richsalz, line 62: This line doesn't seem to work :(

richsalz: Of course that fixed it.  I missed the deref in the previous `for ft in` loops :(
Still need to write a test.

ietf/templates/nomcom/view_feedback.html
```diff
@@ -54,6 +54,14 @@ <h3 class="mt-5" id="declined">Declined each nominated position</h3>
                     </tr>
                 {% endfor %}
             </tbody>
+	    <tfoot>
+		<td></td>
+		<td></td>
+		<th>Totals</h></th>
```
larseggert, line 60: th needs scope

larseggert, line 60:
```suggestion
		<th scope="row">Totals</th>
```

richsalz: Do you think `col` is right?  I didn't  so I left it out

richsalz:
Replaced by #5977

PR #5407, opened 2023-03-25
fix: Update I-D submission instructions. Fix #5264

### LINKED ISSUE #5264: Update submit instructions template
At the very least text like what's at 
https://github.com/ietf-tools/datatracker/blob/5fc93bb8ca243be6aa8c4e848484109e7f3a4d4c/ietf/templates/submit/tool_instructions.html#L142
needs to be changed to match the current reality.
The whole template needs careful review.

russhousley:


ietf/templates/submit/tool_instructions.html
```diff
@@ -30,15 +30,15 @@ <h2 class="mt-3">I-D submission tool instructions</h2>
     </p>
     <p>
         An API for automated Internet-Draft submission is available as an alternative to this webpage at
-        <a href="{{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.submit.views.api_submit' %}">{{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.submit.views.api_submit' %}</a>.
+        <a href="https://datatracker.ietf.org/api/submission">https://datatracker.ietf.org/api/submission</a>.
```
rjsparks, line 33: This should use the settings and url reversals. The view is `ietf.submit.views.api_submission`

ietf/templates/submit/tool_instructions.html
```diff
@@ -30,15 +30,15 @@ <h2 class="mt-3">I-D submission tool instructions</h2>
...
     </p>
     <p>
         The specification for this tool can be found in
-        <a href="{% url 'ietf.doc.views_doc.document_html' name='rfc4228' %}">RFC 4228</a>.
+        <a href="{% url 'ietf.doc.views_doc.document_html' name='rfc4228' %}">RFC 4228</a>.  It has been updated over the years.
     </p>
```
rjsparks, line 38: I think we can just delete this now

ietf/templates/submit/tool_instructions.html
```diff
@@ -60,39 +60,39 @@ <h3 class="mt-4">Upload screen</h3>
...
     <p>
-        After a user uploads the document(s),  the tool will parse the plain-text version, validate the I-D, and display the validation results with option(s) for next steps. The validation includes: checking for all IPR-related notices and I-D boilerplate described in
-        <a href="https://www.ietf.org/ietf/1id-guidelines.html">Guidelines to Authors of Internet-Drafts</a>;
+        After you upload the file(s) and the tool parses the the metadata, the validation results are displayed, and you are offered option(s) for next steps. The validation checks include for IPR-related notices and I-D boilerplate described in
+        <a href="https://www.ietf.org/ietf/1id-guidelines.html">Guidelines to Authors of Internet-Drafts</a>,
         the required sections described in
-        <a href="https://www.ietf.org/ID-Checklist.html">the I-D Check List</a>;
-        the version number; and the creation date.
+        <a href="https://www.ietf.org/ID-Checklist.html">the I-D Check List</a>,
+        the version number, and the creation date.
```
rjsparks, line 89: I wonder if these should be pointing into the authors wiki instead?

rjsparks (changes requested):

ietf/templates/submit/tool_instructions.html
russhousley: Sure.  That is fine with me.

ietf/templates/submit/tool_instructions.html
russhousley: I guess that is a question for Greg Wood.

rjsparks: Those URLs all redirect now to https://authors.ietf.org/en/content-guidelines-overview.

ietf/templates/submit/tool_instructions.html
russhousley: I removed it.

rjsparks (approved):

PR #5409, opened 2023-03-25
fix: Lengthen IESG dashboard trend period to 120 days

### LINKED ISSUE #4450: AD dashboard should also display trend when current == 0
### Description

https://datatracker.ietf.org/doc/ad/ with the trending arrows would benefit by displaying the trend *even* if the current value is 0.

### Code of Conduct

larseggert:
Fixes #4450


evyncke:
Thanks !

rjsparks (approved):

PR #5411, opened 2023-03-25
fix: remove order from DocumentInfo as apparently unneeded (fix #5245)

### LINKED ISSUE #5245: Verify whether DocumentInfo.order is ever used
If it is not used, remove it.

Spectre17:
Includes migration, although I admit that I didn't run the reverse to see if I got that right.


rjsparks:
closing this to get a better one based on codesprint discussion

Issue #4732, opened 2022-11-08
Generating volunteer list


richsalz:
### Description

It would be nice to have an option to generate the volunteer list that just output text/plain with numbering.
```
1. Rich Salz
2. Robert Sparks
3. ...
```

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #5412, opened 2023-03-25
fix: Add "simple" list of nomcom volunteers (#4732)

richsalz:
For doing cut/paste to IETF announcements of the list of volunteers, it can be very convenient to just have a simple text list of name and affiliation. We implement this by adding a "mode" flag to the volunteer functions, and omitting columns in the template in non-full (i.e., simple) mode.


richsalz:
I am not sure if it is better to have multiple `{%if` blocks in the template, or use two different templates which would then have to be kept in sync.  Opinions?

ietf/nomcom/views.py
```diff
@@ -1313,14 +1313,17 @@ def eligible(request, year, public=False):
 def public_volunteers(request, year):
     return volunteers(request=request, year=year, public=True)
 
-def private_volunteers(request, year):
-    return volunteers(request=request, year=year, public=False)
+def private_volunteers(request, year, mode):
+    return volunteers(request=request, year=year, public=False, mode="full")
```
rjsparks, line 1317: This seems strange - why is mode an argument to this function if you're always going to ignore what's passed in?

ietf/templates/nomcom/nomcom_public_base.html
```diff
@@ -81,4 +81,4 @@ <h1>
                     });
             });
     </script>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
```
rjsparks, line 84: Please change your editor to not automatically change the end of file newline presence

rjsparks: (and revert this file since there are no  other changes in it?)

ietf/templates/nomcom/nomcom_private_base.html
```diff
@@ -132,7 +132,12 @@ <h1>
                     </li>
                     <li class="dropdown-item">
                         <a href="{% url 'ietf.nomcom.views.private_volunteers' year %}">
-                            View volunteers
+                            View volunteers (full)
+                        </a>
+                    </li>
+                    <li class="dropdown-item">
+                        <a href="{% url 'ietf.nomcom.views.private_volunteers' year "simple" %}">
```
rjsparks, line 139: This is not doing what you intend, I think. There's nothing in the urls to reverse this to. If possible, lets talk about this in yokohama (I'm at the hackathon today).

rjsparks (changes requested):

rjsparks:
> I am not sure if it is better to have multiple {%if blocks in the template, or use two different templates which would then have to be kept in sync. Opinions?

The general shape of where you're going reads well enough. The tradeoff of breaking this template into one big if/then or dispatching two two different templates is duplication of the presentation, making it more likely that edits take things out of sync.

Consider whether reordering columns so that the verbose columns were always rightmost would work - that would make the tradeoff easier to resolve.

ietf/nomcom/views.py
richsalz: copy-editing error.  the `="full"` should be in the parameter declaration.

richsalz:
superceeded by #5578.

PR #5418, opened 2023-03-25
fix: add review team setting allowing reviewers to reject assignments past the deadline

### LINKED ISSUE #3069: Review team members are unable to reject overdue reviews
`type_defect`   |    by magnus.westerlund@ericsson.com

___

Got a report from one of the TSVART reviewers that they where unable to reject a review request when it was overdue. I think they should be able to. In many cases there are still time to reassign a review request even if the last call data has passed. 

___
_Issue migrated from trac:3069 at 2022-03-04 07:59:29 +0000_

kivinen:
Fixes #3069.

allow_reviewer_to_reject_after_deadline that will allow rejecting review requests, even after the deadline is past. Also modified that the secretary, or whoever manages the reviews is always allowed to reject the review regardless of the deadline as he/she could change the deadline anyways.


jennifer-richards (dismissed):
This looks ok, but I think the new test cases need to verify that POSTs are allowed/disallowed in addition to checking what the GET claims can be done.

rjsparks (approved):

PR #5419, opened 2023-03-25
feat: Allow password reset with non username email address

### LINKED ISSUE #5057: Reset password only works for the primary email address
### Describe the issue

Users can login with any email address associated with the account.
But the reset password only looks at the primary email address.
If a user enters a non-primary email address, they get an error: 
> Didn't find a matching account. If you don't have an account yet, you can [create one](https://datatracker.ietf.org/accounts/create/).

### Code of Conduct

kesara:
This change allows password reset with any email address associated with the account.
The password reset will only be sent to the active email addresses associated with the account.

Fixes #5057

### DIFF
--- a/ietf/ietfauth/tests.py
+++ b/ietf/ietfauth/tests.py
@@ -563,6 +563,21 @@ class IetfAuthTests(TestCase):
         self.assertIn(active_address, to)
         self.assertNotIn(person.user.username, to)
 
+    def test_reset_password_without_username(self):
+        """Reset password using non-username email address"""
+        url = urlreverse('ietf.ietfauth.views.password_reset')
+        person = PersonFactory()
+        secondary_address = EmailFactory(person=person).address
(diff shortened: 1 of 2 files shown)


jennifer-richards (approved):
I think will allow an inactive secondary address to be used to request the reset but will not send the link to that address. As long as we don't allow an inactive address to be associated with a different `User` that seems reasonable to me (thinking of the case where someone has forgotten which addresses are tied to their account).

kesara:
> I think will allow an inactive secondary address to be used to request the reset but will not send the link to that address. As long as we don't allow an inactive address to be associated with a different `User` that seems reasonable to me (thinking of the case where someone has forgotten which addresses are tied to their account).

Yes, that was the idea. To allow people to request for password reset with an inactive email address as well but not to send the link to inactive email addresses.

If we allow inactive email addresses to be associated with a different user, I think we will have a lot of data integrity issues.

PR #5421, opened 2023-03-25
fix(doc): Replace iconography for position's email status with text label

### LINKED ISSUE #3855: IESG evaluation record: can't distinguish mail/no-mail icons
### What happened?

In an IESG evaluation record, there are small envelope-shaped icons in each ballot header that indicate whether mail is requested or not.

These icons are practically indistinguishable.  There are also three different ones, which two different tooltips.

https://datatracker.ietf.org/doc/draft-ietf-cbor-file-magic/ballot/

Chrome / Edge, Safari

### Code of Conduct

aph3rson:
Positions within a ballot, discuss, or comment within DT may-or-may-not have an email notification associated with them. This is at the discretion of the submitter, and is indicated within DT to viewers of the thread.

Previously, this was indicated via an icon+tooltip on the specific position, using an icon which was less-than-readable (especially on low-DPI/small displays).

This change replaces that iconography with a text pill, while retaining the previous tooltip for additional information.

Closes #3855.

### DIFF
--- a/ietf/doc/tests_ballot.py
+++ b/ietf/doc/tests_ballot.py
@@ -1232,7 +1232,6 @@ class BallotContentTests(TestCase):
         self.assertEqual(len(heading), 1)
         # <div.h5> is followed by a panel with the message of interest, so use next()
         next = heading.next()
-        print(f"looking for {expected} in:\n{next}")
         self.assertEqual(
             len(next.find(
                 f'*[title="{expected}"]'
--- a/ietf/doc/tests_ballot.py
+++ b/ietf/doc/tests_ballot.py
@@ -1231,8 +1231,10 @@ class BallotContentTests(TestCase):
         heading = q(f'div.h5[id$="_{slugify(balloter.plain_name())}"]')
         self.assertEqual(len(heading), 1)
         # <div.h5> is followed by a panel with the message of interest, so use next()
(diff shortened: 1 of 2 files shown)


ietf/doc/tests_ballot.py
```diff
@@ -1231,8 +1231,10 @@ def _assertBallotMessage(self, q, balloter, expected):
         heading = q(f'div.h5[id$="_{slugify(balloter.plain_name())}"]')
         self.assertEqual(len(heading), 1)
         # <div.h5> is followed by a panel with the message of interest, so use next()
+        next = heading.next()
+        print(f"looking for {expected} in:\n{next}")
```
rjsparks, line 1235: Leftover debugging?

rjsparks (changes requested):

ietf/doc/tests_ballot.py
aph3rson: Fixed in b430fa2583446581c2de942618b1efba5cb496fd.

aph3rson:
@rjsparks any thoughts on this?

rjsparks:
Been needing time to review - meeting startup has been intense.
There are some things to change, but I'll pick them up later.

rjsparks (approved):

Issue #4005, opened 2022-05-22
Finding the interim meeting request tool is difficult


rjsparks:
From Karen O'Donoghue:
>I wonder if there is a menu option or something that would make the virtual  meeting request tool could be made easier to find. I couldn't find it through the menus. I knew it was there somewhere because I had used it before. I ended up going to the chairs.ietf.org site and searching there. I then found the Upcoming Meetings menu option and the request a meeting tool is linked from there. But maybe it isn't worth taking up more menu space...

PR #5422, opened 2023-03-25
fix: move meeting management items from base menu Documents tab to Meetings tab (fixes #3897 and #4005)

### LINKED ISSUE #3897: Scheduling an interim

Spectre17:
Rearranges elements of the optional Manage block in Documents so that they appear in Meetings instead. This is for things like "xxxx wg meetings". Also, put role restrictions around whether "Request a session" and "Request an interim meeting" appear under the Meetings tab.

### DIFF
--- a/ietf/templates/base/menu.html
+++ b/ietf/templates/base/menu.html
@@ -145,14 +145,6 @@
                     </a>
                 </li>
             {% endfor %}
-            {% for g in user|matman_groups %}
-                <li>
-                    <a class="dropdown-item {% if flavor != 'top' %}text-wrap link-primary{% endif %}"
-                       href="{% url "ietf.group.views.meetings" g.acronym %}">
-                        {{ g.acronym }} {{ g.type.slug }} meetings
-                    </a>
-                </li>
-            {% endfor %}
         {% endif %}
         {% if user|has_role:"Review Team Secretary" %}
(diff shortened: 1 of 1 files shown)


rjsparks (changes requested):
The role based if clauses here do not match real needs (many other group types use interims for example). Lets talk about using leveraging existing utilities to decide when to offer links instead.

Spectre17:
The list of roles I used was based on what was being enforced in the code itself. I just added the role-based restriction to the HTML so that users who were going to trip the code restrictions instead hit HTML restrictions first. Two notes:

1) I'm happy to change the restrictions to a different set or use an existing utility if it can be incorporated into the HTML template. 

2) That part of the merge request is not critical to fixing the issue in any case - it was a secondary thing I hit while in that HTML files, so we could break it out to a separate pull request and allow the main fix to be incorporated without being tied to work on the restrictions.

rjsparks:
overtaken by #6586

PR #5423, opened 2023-03-25
fix: display last year reviews in the profile

### LINKED ISSUE #4706: IETF profile should include the reviews done

evyncke:
Replacing #4768 to use the recent feat/postgres

 Fixes #4706 : display a limited amount of directorate reviews.

Goal is to give some recognitions (or even incentives) to active directorate reviewers.

As discussed with @cjbc @jgscudder @AndrewAlston and a few others

### DIFF
--- a/ietf/templates/person/profile.html
+++ b/ietf/templates/person/profile.html
@@ -165,6 +165,21 @@
             <div id="chart-{{ forloop.counter }}">
             </div>
         {% endif %}
+        <h2 class="mt-5" id="reviews-{{ forloop.counter }}">
+            Directorate Reviews In The Last Year <small class="text-muted">({{ person.completed_reviews|length }})</small>
+        </h2>
+        {% if person.completed_reviews|length > 0 %}
+            <ul>
+                {% for doc in person.completed_reviews %}
+                    <li>
+                        <a href="{{ doc.get_absolute_url }}">{{ doc.title }}</a> on {{ doc.time}}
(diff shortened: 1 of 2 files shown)


ietf/templates/person/profile.html
```diff
@@ -165,6 +165,21 @@ <h2 class="mt-5">
             <div id="chart-{{ forloop.counter }}">
             </div>
         {% endif %}
+        <h2 class="mt-5" id="reviews-{{ forloop.counter }}">
+            Directorate Reviews In The Last Year <small class="text-muted">({{ person.completed_reviews|length }})</small>
```
rjsparks, line 169: Do you really want this to show for every user of the datatracker?
I think it would be better show this only if there are reviews to report on.
People who have accounts that are not the kind of person that does reviews (such as policy fellows who have an account so they can register for meetings) will find this line very confusing.

evyncke: Yet another good point of yours. As you may have noticed, I am working  at low intensity on this issue to take into account your previous comments (including re-using the code of ietfauth/views.py about review_overview()

rjsparks: maybe mark this draft until you're ready for full review again?

PR #5425, opened 2023-03-26
fix:  Remove automatically suggested document for document that is already has review request (fixes #3211)

### LINKED ISSUE #3211: When assignee rejects review secretary gets to see same review twice.
`type_defect`   |    by kivinen@iki.fi

___

I.e., when assigned reviewer rejects a review:

https://datatracker.ietf.org/doc/draft-ietf-core-yang-cbor/reviewrequest/14387/

And I as an secretary go to the 

https://datatracker.ietf.org/review/secdir/reviews/

page I can see the document twice on the "Unassigned review requests" section (see attached figure 1). 

Then when I try to do assignments and click "Manage unassigned reviews", I see the same document again twice (see attached figure 2). 

Now it is very easy to assign same document twice to different people, if I do not notice that the same document is in the list two times. 

If I close the review request that was rejected, both of the items disappear from the "Unassigned review requests". 

Now to get it back I need to go to the document again and request separate review for the document. I think I should have first assigned auto-suggested document to someone, and only after that close the previous review...

The datatracker should understand rejected assignments better, and combine the auto-suggestion and the rejected assignments to one assignment. I.e., most likely just show the rejected review request, and not create auto-suggested one for the same document if there is already rejected review to be assigned for the same document.

___
_Issue migrated from trac:3211 at 2022-03-04 08:12:26 +0000_

kivinen:
Added check that if there is already review request for the document in question, ignore the automatic suggestion for that document. Fixes #3211.


ietf/review/utils.py
```diff
@@ -588,10 +588,12 @@ def blocks(existing, request):
                    and existing.reviewassignment_set.filter(state_id__in=("assigned", "accepted")).exists()
                    and (not existing.requested_rev or existing.requested_rev == request.doc.rev))
         request_closed = existing.state_id not in ('requested','assigned')
+        # Is there a review request for this document already in system
+        requested = existing.state_id in ('requested')
```
rjsparks, line 592: Would it get the same result in the cases you are testing if you added `and existing.type_id == request.type_id` here, or is it specifically existing, say, early assignments that cause problems when LC is requested?

kivinen:
Robert Sparks writes:
> @rjsparks commented on this pull request.
> In ietf/review/utils.py:
> 
> > @@ -588,10 +588,12 @@ def blocks(existing, request):
>                     and existing.reviewassignment_set.filter(state_id__in=("assigned", "accepted")).exists()
>                     and (not existing.requested_rev or existing.requested_rev == request.doc.rev))
>          request_closed = existing.state_id not in ('requested','assigned')
> +        # Is there a review request for this document already in system
> +        requested = existing.state_id in ('requested')
> 
> Would it get the same result in the cases you are testing if you added and
> existing.type_id == request.type_id here, or is it specifically existing, say,
> early assignments that cause problems when LC is requested?
> 

Both can happen I think. I.e., I might have early review request that
got rejeced, but then the document went to LC, and got autosuggested,
thus there is old rejected early review that is open as a request, and
there is new LC request. I do not really care about the review type, I
only care that the review gets done at some point...

Same thing with LC review that got rejected and then it might already
be in the telechat.
-- 
***@***.***

rjsparks:
I want to add some tests to this before merging, but it will be week after next.

kivinen:
Robert Sparks writes:
> I want to add some tests to this before merging, but it will be week
> after next.

Thats ok, as the secretaries are already aware of this bug, and will
most likely still catch those cases before doing duplicate
assignments, so few weeks delay is not an issue.
-- 
***@***.***

rjsparks:
I am very sorry that I haven't gotten the remaining tests in and have this into main already. It will be my initial sprint task if it's not done before then.

rjsparks:
@kivinen - Please review the additional condition I added (and tested).

jennifer-richards (approved):

PR #5427, opened 2023-03-26
fix: Do not add user to the top of queue after reject (fixes #4505)

### LINKED ISSUE #4505: When assignee rejects a review request, it seems to mark him for next for assignment
### Describe the issue

When someone rejects an assignment, for example because they are too busy now, the system seems to automatically mark that parson to "someone who has indicated he wants to be next in queue", which they definately have not done. 

As a secretary I always need to go and edit those users information and remove that checkbox from them before I can continue my assignments, as otherwise it will try to assign next document to them, which they will reject as they are still busy etc.

### Code of Conduct

kivinen:
Added a checkbox in the reject review dialog to ask whether user wants to be added to the top of the queue or not. Default is off.

### DIFF
--- a/ietf/doc/views_review.py
+++ b/ietf/doc/views_review.py
@@ -347,6 +347,7 @@ def assign_reviewer(request, name, request_id):
 
 class RejectReviewerAssignmentForm(forms.Form):
     message_to_secretary = forms.CharField(widget=forms.Textarea, required=False, help_text="Optional explanation of rejection, will be emailed to team secretary if filled in", strip=False)
+    wants_to_be_next = forms.BooleanField(label="I want to be assigned new document immediately", required=False)
 
 @login_required
 def reject_reviewer_assignment(request, name, assignment_id):
@@ -388,11 +389,12 @@ def reject_reviewer_assignment(request, name, assignment_id):
             )
 
             policy = get_reviewer_queue_policy(review_assignment.review_request.team)
-            policy.return_reviewer_to_rotation_top(review_assignment.reviewer.person)
+            policy.return_reviewer_to_rotation_top(review_assignment.reviewer.person, form.cleaned_data['wants_to_be_next'])
 
             msg = render_to_string("review/reviewer_assignment_rejected.txt", {
                 "by": request.user.person,
-                "message_to_secretary": form.cleaned_data.get("message_to_secretary")
+                "message_to_secretary": form.cleaned_data.get("message_to_secretary"),
+                "wants_to_be_next" : form.cleaned_data['wants_to_be_next']
             })
 
             email_review_assignment_change(request, review_assignment, "Reviewer assignment rejected", msg, by=request.user.person, notify_secretary=True, notify_reviewer=True, notify_requested_by=False)
@@ -438,7 +440,7 @@ def withdraw_reviewer_assignment(request, name, assignment_id):
         )            
 
         policy = get_reviewer_queue_policy(review_assignment.review_request.team)
-        policy.return_reviewer_to_rotation_top(review_assignment.reviewer.person)
+        policy.return_reviewer_to_rotation_top(review_assignment.reviewer.person, True)
         
         msg = "Review assignment withdrawn by %s"%request.user.person
 
--- a/ietf/review/policies.py
+++ b/ietf/review/policies.py
@@ -476,9 +476,10 @@ class RotateAlphabeticallyReviewerQueuePolicy(AbstractReviewerQueuePolicy):
         # As RotateAlphabetically does not keep a full rotation list,
         # returning someone to a particular order is complex.
         # Instead, the "assign me next" flag is set.
-        settings = self._reviewer_settings_for(reviewer_person)
-        settings.request_assignment_next = wants_to_be_next
-        settings.save()
+        if wants_to_be_next:
+            settings = self._reviewer_settings_for(reviewer_person)
+            settings.request_assignment_next = wants_to_be_next
+            settings.save()
 
     def _update_skip_next(self, rotation_pks, assignee_person):
         """Decrement skip_next for all users skipped
@@ -570,7 +571,10 @@ class LeastRecentlyUsedReviewerQueuePolicy(AbstractReviewerQueuePolicy):
         # Reviewer rotation for this policy ignores rejected/withdrawn
         # reviews, so it automatically adjusts the position of someone
         # who rejected a review and no further action is needed.
-        pass
+        if wants_to_be_next:
+            settings = self._reviewer_settings_for(reviewer_person)
+            settings.request_assignment_next = wants_to_be_next
+            settings.save()
 
 
 QUEUE_POLICY_NAME_MAPPING = {
--- a/ietf/review/policies.py
+++ b/ietf/review/policies.py
@@ -84,7 +84,7 @@ class AbstractReviewerQueuePolicy:
             rotation_list = self._filter_unavailable_reviewers(rotation_list)
         return rotation_list
 
-    def return_reviewer_to_rotation_top(self, reviewer_person):
+    def return_reviewer_to_rotation_top(self, reviewer_person, wants_to_be_next):
         """
         Return a reviewer to the top of the rotation, e.g. because they rejected a review,
         and should retroactively not have been rotated over.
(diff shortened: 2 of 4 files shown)


rjsparks:
The test fails are because of
https://github.com/kivinen/datatracker/blob/fix-4505/ietf/review/tests_policies.py#L512
and
https://github.com/kivinen/datatracker/blob/fix-4505/ietf/review/tests_policies.py#L728

But consider changing
```
def return_reviewer_to_rotation_top(self, reviewer_person, wants_to_be_next):
```
to
```
def return_reviewer_to_rotation_top(self, reviewer_person, wants_to_be_next=False):
```
(or True - whichever way would have preserved older behavior.

ietf/doc/views_review.py
```diff
@@ -438,7 +440,7 @@ def withdraw_reviewer_assignment(request, name, assignment_id):
         )            
 
         policy = get_reviewer_queue_policy(review_assignment.review_request.team)
-        policy.return_reviewer_to_rotation_top(review_assignment.reviewer.person)
+        policy.return_reviewer_to_rotation_top(review_assignment.reviewer.person, True)
```
rjsparks, line 443: With the change I suggested in the top level comments, you could leave this line unchanged (don't pass the second argument). If you want to pass it explicitly, use the named instead of positional form please.

rjsparks (changes requested):

ietf/review/policies.py
```diff
@@ -84,7 +84,7 @@ def default_reviewer_rotation_list(self, include_unavailable=False):
             rotation_list = self._filter_unavailable_reviewers(rotation_list)
         return rotation_list
 
-    def return_reviewer_to_rotation_top(self, reviewer_person):
+    def return_reviewer_to_rotation_top(self, reviewer_person, wants_to_be_next):
```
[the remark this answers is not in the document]
kivinen: This is the abstractreviewerqueuepolicy so it does not do anything with the arguments.

ietf/doc/views_review.py
kivinen: So I had already added the wants_to_be_next to the test_policies.py, but the tests still fail. Thats why I asked yesterday, whether you wanted me to commit those changes to the github, even when I know the test will fail. 

Before adding them there, they failed with a line number, so it was easy to add them there, now when I added those parametrs they simply fail saying type error and no line number:

`dev ➜ /workspace $ ietf/manage.py test --skip-coverage --no-validate-html --traceback --settings=settings_postgrestest  ietf.review.tests_policies.LeastRecentlyUsedReviewerQueuePolicyTest.test_return_reviewer_to_rotation_top ietf.review.tests_policies.RotateAlphabeticallyReviewerQueuePolicyTest.test_return_reviewer_to_rotation_top
     Datatracker 8.0.0-dev test suite, 27 March 2023 02:01:29 PDT:
     Python 3.9.16 (main, Feb  4 2023, 12:10:12)  [GCC 10.2.1 20210110].
     Django 2.2.28, settings 'settings_postgrestest'
     Changing TEMPLATES[0]['OPTIONS']['string_if_invalid'] to '' during testing
     Changing INTERNAL_IPS to '[]' during testing.
     Running an SMTP test server on 127.0.0.1:2026 to catch outgoing email.
     Loading factory-boy random state from .factoryboy_random_state
     Not validating any generated HTML; please do so at least once before committing changes
     Creating test database...
     Loading global test fixtures: names, ietf.utils.test_data.make_immutable_base_data, nomcom_templates, proceedings_templates
System check identified no issues (1 silenced).
EE
======================================================================
ERROR: test_return_reviewer_to_rotation_top (ietf.review.tests_policies.LeastRecentlyUsedReviewerQueuePolicyTest.test_return_reviewer_to_rotation_top)
----------------------------------------------------------------------
TypeError: test_return_reviewer_to_rotation_top() missing 2 required positional arguments: 'reviewer' and 'wants_to_be_next'

======================================================================
ERROR: test_return_reviewer_to_rotation_top (ietf.review.tests_policies.RotateAlphabeticallyReviewerQueuePolicyTest.test_return_reviewer_to_rotation_top)
----------------------------------------------------------------------
TypeError: test_return_reviewer_to_rotation_top() missing 2 required positional arguments: 'person' and 'wants_to_be_next'

----------------------------------------------------------------------
Ran 2 tests in 0.276s

FAILED (errors=2)

Destroying test database for alias 'default'...
dev ➜ /workspace $ 

`

kivinen: Changing the parameters to be optional would solve the issue, but would not allow making tests that actually test that feature. On the other hand I do not think there is test currently that tests that code yet.

rjsparks: And in my clone of your PR, if I add at True argument to the two lines above, the tests pass without error.

kivinen:
Robert Sparks writes:
> If you check in what is now broken I can help diagnose better.

Done already.

> I don't see how making the parameter optional removes the ability to test the
> feature. You're in control of what the tests provide when they call the
> method...

Perhaps, but I would like to also know why the tests are failing even
when tests are updated.
-- 
***@***.***

rjsparks:
Ok - found it. The confusion is in the similarly named things.

The failure you are seeing now is because you changed the signature of the test functions. Note the error
```
TypeError: test_return_reviewer_to_rotation_top() missing 2 required positional arguments: 'reviewer' and 'wants_to_be_next'
```
It's not the return_reviewer_to_rotation_top function that's being complained about - it's the definition of the **test**_return_reviewer_to_rotation_top function.

The test framework searches classes derived from TestCase, looking for methods that have names that start with test_ and calls them with no arguments other than self.

ietf/review/tests_policies.py
```diff
@@ -507,9 +507,9 @@ def test_default_reviewer_rotation_list_with_nextreviewerinteam(self):
         rotation = self.policy.default_reviewer_rotation_list()
         self.assertEqual(rotation, available_reviewers[2:] + available_reviewers[:1])
 
-    def test_return_reviewer_to_rotation_top(self):
+    def test_return_reviewer_to_rotation_top(self, person, wants_to_be_next):
```
rjsparks, line 510: The signature here should  be just (self).

ietf/review/tests_policies.py
```diff
@@ -507,9 +507,9 @@ def test_default_reviewer_rotation_list_with_nextreviewerinteam(self):
...
         reviewer = self.append_reviewer()
-        self.policy.return_reviewer_to_rotation_top(reviewer)
+        self.policy.return_reviewer_to_rotation_top(reviewer, wants_to_be_next)
```
rjsparks, line 512: Here, you need to choose a value for wants_to_be_next and pass it in. Call this twice (appending a new reviewer each time), once with False and once with True and check that the right thing happened with likes like what are at 513 below.

It would be good to add tests like that to the other policy test as well

rjsparks (changes requested):

ietf/review/tests_policies.py
kivinen: Ok fixed now, and now the tests passes. I test both the wants to be next cases, in both policies. Actually this function is works identically for both policies.

ietf/review/tests_policies.py
```diff
@@ -724,8 +726,11 @@ def test_default_review_rotation_list_uses_assigned_on_date(self):
                          available_reviewers[2:] + [first_reviewer, second_reviewer])
 
     def test_return_reviewer_to_rotation_top(self):
-        # Should do nothing, this is implicit in this policy, no state change is needed.
-        self.policy.return_reviewer_to_rotation_top(self.append_reviewer())
+        reviewer = self.append_reviewer()
+        self.policy.return_reviewer_to_rotation_top(reviewer, False)
+        self.assertFalse(self.reviewer_settings_for(reviewer).request_assignment_next)
```
rjsparks, line 731: Between 731 and 732, I think you need to make a new reviewer for the test (by repeating what's at 729.
That should happen both here and for the other policy.

rjsparks (changes requested):

ietf/review/tests_policies.py
kivinen: Why? The first call to return_reviewer_to_rotation_top with parameter False, does not really do anything, it will not modify the request_assignment_next in any way, so it should still be false after that first call. Then on the second call we call it with wanting to be next set to True, and that will cause request_assignment_next to be set in both policies, and then we check that it is now set.

rjsparks: so the unit here is badly named now if all we're doing is setting and testing the "assign me next" bit. If the implementation is really _only_ setting that bit and no other queue manipulation is made, then we should rename this. Otherwise, we'd be missing something moving someone who is already at head of queue (rotation top) to the head of the queue.

rjsparks: I'll add an issue to make that change later.

rjsparks (approved):

ietf/review/tests_policies.py
kivinen: I agree that it is badly named. I think this is artifact that there is no clearly defined method of moving someone on the top, other than asking to be next. This function is called in two different situations, once when review request is rejected, and in that case I added the checkbox. It is also called when the document is withdrawn, and in that case this function will be called with wants_to_be_first set to true, as this happens when secretary messed up and assigned something that should not have been assigned, thus it is logical that the reviewer will get new assignment as soon as possible. On the other hand we could add similar checkbox to that location, but as that thing is only used by secretary and very rarely anyways, it might not be needed.

PR #5439, opened 2023-03-29
feat: Add requests summary

rpcross:


ietf/meeting/tests_views.py
```diff
@@ -6589,6 +6590,28 @@ def test_handles_notes_server_failure(self):
 
 class SessionTests(TestCase):
 
+    def test_get_summary_by_area(self):
+        meeting = make_meeting_test_data(meeting=MeetingFactory(type_id='ietf', number='100'))
+        sessions = Session.objects.filter(meeting=meeting).with_current_status()
+        data = get_summary_by_area(sessions)
+        assert data[0][0] == 'Duration'
```
rjsparks, line 6597: bare asserts don't work nicely with the test framework. self.assertEqual is what you should be using here.
The other bare asserts should also be updated.

rjsparks (changes requested):

ietf/meeting/tests_views.py
rpcross: That's my pytest muscle memory, will fix.

rjsparks (approved):

PR #5450, opened 2023-03-30
feat: Add list of sub-groups to group's "about" page

### LINKED ISSUE #5441: View WGs in a given area from the area group view.
### Description

From an area group view, [such as for ART](https://datatracker.ietf.org/group/art/about/), it would be nice to be able to view all WGs in that group. This is visible at [the list of all WGs](https://datatracker.ietf.org/wg/#art), but this requires going through the list of all WGs.

### Code of Conduct

larseggert:
Fixes #5441

### DIFF
--- a/ietf/templates/group/group_about.html
+++ b/ietf/templates/group/group_about.html
@@ -406,7 +406,7 @@ height: 100vh;
     {% endif %}
     {% if subgroups %}
         <h2 class="mt-4">
-            Subgroups under {{ group.acronym|upper }}
+            Groups in {{ group.acronym|upper }}
         </h2>
         <table class="table table-sm table-striped tablesorter">
             <thead>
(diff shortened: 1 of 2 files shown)


ietf/templates/group/group_about.html
```diff
@@ -401,8 +404,38 @@ <h2 class="mt-3">
             </p>
         {% endif %}
     {% endif %}
+    {% if subgroups %}
+        <h2 class="mt-4">
+            Subgroups under {{ group.acronym|upper }}
```
rjsparks, line 409: Normal people would expect to just see "Groups under ART" not "Subgroups under ART"

rjsparks: And consider "Group in ART" rather then "under"

rjsparks (approved):

rjsparks:
intentionally squashing this into main rather than feat/postgres

PR #5486, opened 2023-04-11
fix: Unbreak HTMLized table of contents

### LINKED ISSUE #5483: doc/html Contents empty
### Describe the issue

Some recent internet-drafts are showing up with an empty table of contents in the side panel.  I'm also getting a JS error that might have something to do with that:

```
properties of undefined (reading 'match')
    at wr (nav.js:11:20)
    at Array.map (<anonymous>)
    at xr (nav.js:27:10)
    at HTMLDocument.<anonymous> (document_html.js:45:5)
```

### Code of Conduct

larseggert:
Fixes #5483

### DIFF
--- a/ietf/static/js/document_html.js
+++ b/ietf/static/js/document_html.js
@@ -42,10 +42,8 @@ document.addEventListener("DOMContentLoaded", function (event) {
 
     // Set up a nav pane
     const toc_pane = document.getElementById("toc-nav");
-    populate_nav(toc_pane,
-        `#content h2, #content h3, #content h4, #content h5, #content h6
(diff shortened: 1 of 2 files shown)


rjsparks (approved):

rjsparks:
@NGPixel How can we get tests around this?

larseggert:
At some point I added code so that CI fails when there are unexpected messages in the JS console - but that doesn't seem to trigger anymore?

NGPixel:
We could add a simple Playwright test on an internet draft we choose. But if the bug only occurred on recent drafts rather than all of them, it might not be helpful at catching future bugs.

PR #5501, opened 2023-04-17
feat: Show how long someone has served in role on group page

larseggert:
@rjsparks the datatracker history for roles only seems to go back ~11 years, any idea how to indicate if someone has served longer?


larseggert:
Right, but the thing I'm trying to do here is quite a bit simpler and I was hoping it was doable given the current state of things: I'm trying to show for a person who is currently in a role when they started in that role for their current term only.

rjsparks:
>  I'm trying to show for a person who is currently in a role when they started in that role for their current term only.

I don't think you can do that.
When time allows, I'll find the cases that we already have that break.

larseggert:
@rjsparks let's chat about this sometime. I have tried to find a case where it is incorrect, but haven't managed to.

rjsparks:
Well, to start with - this view of /group/iab/about doesn't look right:
image

ietf/group/models.py
```diff
@@ -391,6 +393,32 @@ def formatted_ascii_email(self):
...
+            .first()
+        )
+        print("history", self.person.plain_name(), history)
+        current = (
+            Person.objects.filter(
+                pk=self.person.pk, role__name=self.name, role__group=self.group
+            )
+            .values_list("role__group__time")
+            .first()
+        )
```
rjsparks, line 415: What is this trying to accomplish that's different than just taking self.group.time?

rjsparks:
I'm also not understanding this:
```
>>> from ietf.group.models import Role
>>> r = Role.objects.get(person__name="Lars Eggert", name="chair", group__acronym="iesg")
>>> r.started_in_role()
history Lars Eggert (datetime.datetime(2020, 6, 15, 23, 40, 49, tzinfo=<UTC>),)
current Lars Eggert (datetime.datetime(2023, 3, 29, 10, 37, 32, tzinfo=<UTC>),)
start Lars Eggert 2020-06-15 23:40:49+00:00
(datetime.datetime(2020, 6, 15, 23, 40, 49, tzinfo=<UTC>), False)
```

Digging into the GroupHistory objects for the IESG group, it's clear that we didn't bother keeping roles there until quite recently and even then, there have been hiccups.

For the IESG group, you might have more success looking at the AD role across working groups, but that's going to be a mess as well.

The IAB is probably suffering similarly.

I'll go looking for a WG where the current approach fails later in the week.

larseggert:
@rjsparks good point about IESG and IAB. Let's restrict this to WG pages, because that's where I see most of the value. (One reason I want to add this is to nudge ADs to recognize cases where a change in chairs might be indicated.)

larseggert:
Closing this in favor of #6299

Issue #5504, opened 2023-04-18
Extract mail list should show when empty


richsalz:
### Describe the issue

```
Email lists
Nominating Committee 2022/2023
Nominees Pending Acceptance
Accepted Nominees

    [Al Morton] ([acmorton@att.com])
    ...
```

Adding `<i>None</i>` after the pending acceptance label will make things more clear.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #5505, opened 2023-04-18
fix: Add "empty-list" indicator (#5504)

richsalz:


rjsparks (approved):
This construct is fine - it's more readable than something else django provides explicitly for this purpose. I'm pointing to it here in case you ever run into it and wonder what it is all about: https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#for-empty

PR #5506, opened 2023-04-18
fix: Handle integrity violations when confirming email address

jennifer-richards:
This avoids server errors when an email confirmation link is used more than once or when the confirmed email address is in use for another User.

The latter case is mostly prevented by the user edit views, but can come up if there are concurrent attempts to add the same address to two different Users.

### DIFF
--- a/ietf/ietfauth/views.py
+++ b/ietf/ietfauth/views.py
@@ -53,6 +53,7 @@ from django.contrib.auth.models import User
 from django.contrib.auth.views import LoginView
 from django.contrib.sites.models import Site
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
+from django.db import IntegrityError
 from django.urls import reverse as urlreverse
 from django.http import Http404, HttpResponseRedirect, HttpResponseForbidden
 from django.shortcuts import render, redirect, get_object_or_404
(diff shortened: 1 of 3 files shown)


ietf/ietfauth/views.py
```diff
@@ -440,15 +441,27 @@ def confirm_new_email(request, auth):
...
+            new_email_obj, created = Email.objects.get_or_create(
+                address=email, 
+                person=person, 
+                defaults={'origin': username},
+            )
+        except IntegrityError:
+            can_confirm = False
+            form.add_error(
+                None, "Email address is in use by another user. Please contact the secretariat for assistance."
+            )
```
larseggert, line 456: Wouldn't this again leak the existence of email addresses?

jennifer-richards: Only if you can receive messages sent to that address - this code runs in response to following the confirmation link that was just emailed there.

Further, the duplicate confirmation link will only be generated if you attempt to add the address between the times its rightful owner adds it and confirms it.

My thought was that if you can engineer your way through that you probably already know the address is there. I think this is more likely to come up for someone having account issues so it is better to nudge them to get help. I could make it treat this like an invalid link, though, and that would leak the least possible information.

rjsparks: You could also get to this code by brute force posts, but it would require creating a valid signature over the username and email, and _that_ essentially requires knowing the site's secret key. If that's compromised, there are worse problems.

rjsparks (approved):

PR #5515, opened 2023-04-21
fix: Remove IESG-related about tabs from non-IETF-stream docs

### LINKED ISSUE #5514: Remove IESG-related tabs from document about page
### Description

For non-IETF streams. As discussed at tools retreat. Specifically, "IESG evaluation record" and "IESG writeups".

### Code of Conduct

larseggert:
Fixes #5514


ietf/templates/doc/document_top.html
```diff
@@ -9,11 +9,13 @@ <h1>
 </h1>
 <ul class="nav nav-tabs my-3">
     {% for name, t, url, active, tooltip in tabs %}
-        <li {% if tooltip %}title="{{ tooltip }}"{% endif %} class="nav-item">
-            <a class="nav-link {% if t == selected %}active{% endif %}{% if not active %}disabled{% endif %}"
-               {% if active %}href="{{ url }}"{% endif %}>
-                {{ name|capfirst_allcaps }}
-            </a>
-        </li>
+        {% if doc.stream.name|lower == 'ietf' or t not in 'ballot,writeup' %}
```
rjsparks, line 12: From a maintenance point of view it would be better to not put these tabs in the `tabs` list in the first place in the view code.

rjsparks: This will also bury old ballots (from the IESG Tracker days) on non-ietf-stream documents, when the conflict review was an IESG ballot on the doc instead of an IESG ballot on a conflict-review document.

rjsparks: I think research will be needed to see what will work - maybe we place the tabs if the stream is ietf _or_ there is an IESG ballot exists. And I wonder if we can treat the two tabs differently.

rjsparks: See. for example, https://datatracker.ietf.org/doc/rfc5345/ballot/

rjsparks (changes requested):

ietf/doc/views_doc.py
```diff
@@ -114,12 +114,12 @@ def render_document_top(request, doc, tab, name):
...
-        tabs.append(("IESG Evaluation Record", "ballot", urlreverse("ietf.doc.views_doc.document_ballot", kwargs=dict(name=name)), iesg_ballot,  None if iesg_ballot else "IESG Evaluation Ballot has not been created yet"))
-    elif doc.type_id == "charter" and doc.group.type_id == "wg":
-        tabs.append(("IESG Review", "ballot", urlreverse("ietf.doc.views_doc.document_ballot", kwargs=dict(name=name)), iesg_ballot, None if iesg_ballot else "IESG Review Ballot has not been created yet"))
+    if doc.type_id in ("draft","conflrev", "statchg") and iesg_ballot:
+        tabs.append(("IESG Evaluation Record", "ballot", urlreverse("ietf.doc.views_doc.document_ballot", kwargs=dict(name=name)), iesg_ballot,  None))
+    elif doc.type_id == "charter" and doc.group.type_id == "wg" and iesg_ballot:
+        tabs.append(("IESG Review", "ballot", urlreverse("ietf.doc.views_doc.document_ballot", kwargs=dict(name=name)), iesg_ballot, None))
     
-    if doc.type_id == "draft" or (doc.type_id == "charter" and doc.group.type_id == "wg"):
+    if (doc.type_id == "draft" and iesg_ballot) or (doc.type_id == "charter" and doc.group.type_id == "wg"):
```
rjsparks, line 122: should `and iesg_ballot` be part of the second clause? If so, maybe this should be 

`if iesg_ballot and (doc.type_id == "draft" or (doc.type_id == "charter" and doc.group.type_id == "wg"):

ietf/doc/views_doc.py
```diff
@@ -114,12 +114,12 @@ def render_document_top(request, doc, tab, name):
                 rsab_ballot,  
                 None if rsab_ballot else "RSAB Evaluation Ballot has not been created yet"
             ))
-    if doc.type_id in ("draft","conflrev", "statchg"):
-        tabs.append(("IESG Evaluation Record", "ballot", urlreverse("ietf.doc.views_doc.document_ballot", kwargs=dict(name=name)), iesg_ballot,  None if iesg_ballot else "IESG Evaluation Ballot has not been created yet"))
```
rjsparks, line 118: This makes it so that the tab doesn't get placed there's no ballot. Earlier feedback was that people got lost if the tabs shifted around depending on document _state_. (Changing what tabs appear based on document _type_ is different).

Will it not cause issues to be raised when a chair notices that there is no IESG evaluation tab even through they've sent the document to the IESG? The current view pushes against that by explaining why the tab doesn't have content...

larseggert: Yeah, good point. I'll need to think about what condition would be correct...

ietf/doc/views_doc.py
larseggert: `iesg_ballot` should probably be replaced with whatever the correct condition for your previous comment is?

rjsparks (approved):

ietf/doc/views_doc.py
```diff
@@ -114,13 +114,46 @@ def render_document_top(request, doc, tab, name):
...
+                    "IESG Evaluation Record",
+                    "ballot",
+                    urlreverse(
+                        "ietf.doc.views_doc.document_ballot", kwargs=dict(name=name)
+                    ),
+                    iesg_ballot,
+                    None,
+                )
+            )
+        elif doc.type_id == "charter" and doc.group.type_id == "wg":
```
rjsparks, line 131: This and L144 need a guard on doc.group being non-None before looking at the type_id.
See the test fail at https://github.com/ietf-tools/datatracker/actions/runs/5322219096/jobs/9638362857?pr=5515#step:6:88

rjsparks (changes requested):

rjsparks (approved):

PR #5540, opened 2023-04-26
fix: Avoid 500 error creating account when User.person is None

### LINKED ISSUE #5537: 500 error when creating an account for an email associated with a User that has no related Person
### Describe the issue

Attempting to register through the `create_account()` view fails with a 500 server error if the email is the username for a `User` that has no related `Person` record.

### Code of Conduct

jennifer-richards:
Fixes #5537

### DIFF
--- a/ietf/ietfauth/views.py
+++ b/ietf/ietfauth/views.py
@@ -151,7 +151,13 @@ def create_account(request):
                 # Either a User or an Email matching new_account_email is in the system but we do not have a
                 # "good" email to use to contact its owner. Fail so the user can contact the secretariat to sort
                 # things out.
-                form.add_error("email", ValidationError(f"Unable to create account for {new_account_email}. Please contact the secretariat."))
+                form.add_error(
+                    "email",
+                    ValidationError(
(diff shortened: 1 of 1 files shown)


ietf/ietfauth/views.py
```diff
@@ -123,16 +123,36 @@ def create_account(request):
...
+            to_email = person_to_contact.email_address() if person_to_contact else None
+
+            if to_email:
+                # We have a "good" email - send instructions to it
+                send_account_creation_exists_email(request, new_account_email, to_email)
+            elif email_is_known:
+                # Either a User or an Email matching new_account_email is in the system but we do not have a
+                # "good" email to use to contact its owner. Fail so the user can contact the secretariat to sort
+                # things out.
+                form.add_error("email", ValidationError(f"Unable to create account for {new_account_email}. Please contact the secretariat."))
```
rjsparks, line 154: May as well provide support@ietf.org here.

jennifer-richards: Good idea, will do

rjsparks (approved):

Issue #5261, opened 2023-03-02
ISE can't request conflict review


rjsparks:
The code that chooses to present the "Begin conflict review" button on the doc page needs adjustment.

See
https://github.com/ietf-tools/datatracker/blob/5fc93bb8ca243be6aa8c4e848484109e7f3a4d4c/ietf/doc/views_doc.py#L430-L431

Make sure the view is tested as the ISE

PR #5570, opened 2023-05-03
fix: allow ISE to initiate conflict review

elear:
This very small change should address #5261 .  Very small change to a list.


elear:
For numerous reasons I have **NOT** tested this code, directly (I haven't yet replicated the environment, and the fork of this code is MASSIVE from a remote location.

rjsparks:
Thanks for this. The PR has one part of the essential fix, but there are a couple of other places that need to be touched and this needs to have tests added to the test suite. I'll work these in.

rjsparks:
@elear: I pushed a commit that changes the conditional and adds tests.

jennifer-richards (approved):

PR #5580, opened 2023-05-05
feat: Process uploaded submissions asynchronously

jennifer-richards:
This replaces the validation/checking done in the `upload_submission()` view with asynchronous processing. This uses a celery task in a similar way to the `api_submission()` handling.

**Important** This renames the `process_uploaded_submission()` task to `process_and_accept_uploaded_submission()` and reuses the original name. I think these are the best names for the tasks, but it does introduce the possibility of mishandling tasks. To avoid this, the celery queue must be emptied before deploying the new code. This can be done by bringing the Django server down, waiting for any existing tasks to complete, then shutting down the celery containers and continuing with the deployment. It'd be best to deploy at a time when submissions are not frequent to minimize the chance that this requires down time.

This drops support for uploading PDF files. Text file uploads are still accepted through the interactive `upload_submission()` endpoint and processed as before. The `api_submission()` endpoint still only accepts XML.

When a draft is in the "validating" state, the `submission_status()` view now uses javascript to reload the page every 30 seconds. This can be disabled with a toggle switch and should be tolerant of javascript being disabled (but the page will not reload). It would be possible to make this more sophisticated, starting by using the API submission status endpoint to check whether a reload is needed, but that is left as a future exercise.


NGPixel (approved):

rjsparks (changes requested):
As discussed on a different channel - some of the text on the status page should be updated to call extra attention to the change of behavior.

rjsparks (approved):

PR #5599, opened 2023-05-10
fix: show htmlized views of draft revisions without DocHistory

### LINKED ISSUE #4933: Many old draft versions missing from Datatracker
### Describe the issue

This issue appears to be related to #4898.

Several versions of older drafts do not appear in the Data Tracker UI. While they appear in the timeline diagram, they do not appear in the list of versions right above it, and clicking on the timeline to open the missing versions redirects to the latest publication instead.

This is a problem because some old drafts are still widely in use, such as SFTP version 3.

Examples:

- https://datatracker.ietf.org/doc/draft-ietf-secsh-filexfer/
- https://datatracker.ietf.org/doc/rfc3696/

Note that not all Internet Drafts are affected, such as https://datatracker.ietf.org/doc/rfc8989/

Workaround: Access the draft through https://www.ietf.org/archive/id/[name]-[version].txt

image

### Code of Conduct

rjsparks:
fixes #4933

### DIFF
--- a/ietf/doc/tests.py
+++ b/ietf/doc/tests.py
@@ -617,7 +617,9 @@ Man                    Expires September 22, 2015               [Page 3]
                 f.write(self.draft_text)
 
     def test_document_draft(self):
-        draft = WgDraftFactory(name='draft-ietf-mars-test',rev='01')
+        draft = WgDraftFactory(name='draft-ietf-mars-test',rev='01', create_revisions=range(0,2))
+
+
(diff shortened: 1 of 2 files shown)


rjsparks:
This is egregious, but the document_html hack throughout document_main is already egregious. The view needs to be re-engineered.

jennifer-richards (approved):

jennifer-richards (approved):

PR #5619, opened 2023-05-15
fix: show meeting timezone when editing sessions

### LINKED ISSUE #5353: Another location where meeting time zone info is incorrect
### Describe the issue

Related to #5285 ; time data is incorrectly showing in UTC. This is not a showstopper since I'm probably the only one who would see this page, but I'm just alerting you to one more place the time zone is showing as UTC instead of meeting time: on the "Edit Session" page. 

See below for an example: these office hours are scheduled for 10:30 local time but the edit session page says 01:30.

Screen Shot 2023-03-15 at 2 24 42 PM

### Code of Conduct

rjsparks:
Fixes #5353

### DIFF
--- a/ietf/meeting/templatetags/session_filters.py
+++ b/ietf/meeting/templatetags/session_filters.py
@@ -1,4 +1,7 @@
+# Copyright The IETF Trust 2023, All Rights Reserved
 from django import template
+
+from ietf.name.models import SessionStatusName
 
 register = template.Library()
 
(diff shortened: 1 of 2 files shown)


jennifer-richards (approved):
This should work, though it'll print the timezone as "PST8PDT" or "America/Halifax" rather than a more typical brief description like "PDT" or "ADT". I don't think that's a problem, just pointing it out in case you happened to test with UTC where it might not have been obvious.

PR #5624, opened 2023-05-15
fix: tweak removal of document info order

rjsparks:


ietf/doc/migrations/0003_remove_document_info_order.py
```diff
@@ -1,34 +1,20 @@
...
-    migrations.AddField(
-        model_name='document',
-        name='order',
-    )
-
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("doc", "0002_auto_20230320_1222"),
+        ('doc', '0003_remove_document_info_order'),
```
jennifer-richards, line 8: Need to revert this change, the migration can't depend on itself!

jennifer-richards (changes requested):

jennifer-richards (approved):

PR #5628, opened 2023-05-15
feat: change fonts to Inter and Noto Sans Mono

NGPixel:
- Sets fonts to **Inter** and **Noto Sans Mono**, using the static.ietf.org CDN.
- Removes Google Fonts references


rjsparks (approved):

larseggert (commented):
I think you also need to change the htmlized doc base template?

NGPixel:
@larseggert fixed by #5636

Issue #5648, opened 2023-05-17
Enable localization with USE_L10N


jennifer-richards:
### Description

As of Django 4.0, the `USE_L10N` setting is deprecated and defaults to `True`. In Django 5, the setting will be removed and treated as though it were `True`.

Datatracker should be migrated to `USE_L10N=True` in advance of that.

Doing so now causes a handful of tests to fail because dates change format so that `str(some_datetime)` no longer matches `{% some_datetime|time %}` in the template. There are likely other places where dates/times will be rendered differently but are not caught by tests.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

---
Aggregated with agent assistance on 2026-08-18

Issue #5635, opened 2023-05-16
Migrate from pytz to zoneinfo


jennifer-richards:
### Description

As of Django 4.0, zoneinfo replaces pytz as Django's default timezone library. At Django 5.0, pytz support will be removed. In the effort to migrate Datatracker to Django 4 we have made use of the `USE_DEPRECATED_PYTZ` setting and the `pytz_deprecation_shim` package. Before we get to Django 5.0, we need to replace pytz entirely in the code base.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

---
Aggregated with agent assistance on 2026-08-18

jennifer-richards:
See https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html for suggestions on how to make the transition. Much of it will be little more than glorified search and replace.

PR #5649, opened 2023-05-17
chore: Upgrade to Django 4.0

jennifer-richards:
According to the release notes, the format of sessions, password reset links, and a number of other elements has changed. Any of these from pre-Django 3.1 will be invalidated when we switch to this release.

Two pieces of technical debt have been added to the pile in this PR:
* `USE_L10N = False` is set in `settings.py` to maintain our previous behavior. The default for this became `True` with this release, and there will be no option to disable it starting with Django 5. Ticket #5648 was created to track switching over to `True`
* `USE_DEPRECATED_PYTZ = True` is set. Ticket #5635 was created to track switching everything to `zoneinfo`

Several tests were failing (and likely the Datatracker was actually broken) because the `@require_api_key` decorator stopped being compatible with the `@method_decorator` meta-decorator. I am not entirely sure why this happened, but implementing the `@require_api_key` with `functools.wraps()` instead of `@decorator` has resolved this. The `@decorator` documentation notes that it does not support decorating generic callables, so I'm assuming it just happened to work before and changes in Django 4 broke that. I think we should consider getting rid of the handful of uses of the `@decorator` in favor of the functools approach, but have not done so here.


jennifer-richards:
The changes to `test_runner` improve our URL coverage stats a bit because they identify and ignore subclasses of `RedirectView` and `TemplateView` instead of only those classes themselves. We have a couple subclasses of `RedirectView` amounting to ~half a dozen uncovered views that are now excluded from the counts.

rjsparks:
Yay - we are closer to 100% of testing everything except what we are not testing. 🙄

rjsparks (approved):

PR #5657, opened 2023-05-18
chore: remove gather from CRSF_TRUSTED_ORIGINS

rjsparks:


jennifer-richards (approved):
As we discussed, not sure the '*' is doing what we think here for Django < 4.0, but better to leave well enough alone.

PR #5665, opened 2023-05-19
chore: Upgrade to Django 4.2

jennifer-richards:
This brings us to the current version of Django.

Tests all pass and all but one deprecation warning have been resolved or suppressed after creating a ticket to track eventual resolution. The one exception, caused by a GET to the OIDC end-session view, will need to be resolved in `django-oidc-provider` and has simply been suppressed. When we get to Django 5, we should re-enable the warning and run tests to be sure we are clear of the problem.

A few significant changes:

The session serializer has been changed to `JSONSerializer`. There was one problem with this - the nomcom private key was being stored in the session as `bytes`, which cannot be serialized directly as JSON. I fixed this by decoding as utf8 before stuffing into the session and encoding as utf8 when retrieving.

Django 5 will change the default rendering of formsets to use a div-based structure. I've opted in to adopting this change early by setting `FORM_RENDERER` in `settings.py`. I checked a few places to see that this did not affect our views, but we should probably check more thoroughly. Hopefully the test crawl will catch problems if there are any.

Logging out via GET has been deprecated. The "Sign out" button in the dropdown menu has been changed to a form with a button to POST to the logout view. This worked without needing to modify CSS, but a few tests that counted the number of forms or number of "submit" buttons needed to be tweaked to ignore the sign out form.

### DIFF
--- a/ietf/nomcom/utils.py
+++ b/ietf/nomcom/utils.py
@@ -172,6 +172,12 @@ def command_line_safe_secret(secret):
     return base64.encodebytes(secret).decode('utf-8').rstrip()
 
 def retrieve_nomcom_private_key(request, year):
+    """Retrieve decrypted nomcom private key from the session store
+
+    Retrieves encrypted, ascii-armored private key from the session store, encodes 
+    as utf8 bytes, then decrypts. Raises UnicodeError if the value in the session
+    store cannot be encoded as utf8.
+    """
     private_key = request.session.get('NOMCOM_PRIVATE_KEY_%s' % year, None)
 
     if not private_key:
@@ -183,6 +189,7 @@ def retrieve_nomcom_private_key(request, year):
             settings.OPENSSL_COMMAND,
             command_line_safe_secret(settings.NOMCOM_APP_SECRET)
         ),
+        # The openssl command expects ascii-armored input, so utf8 encoding should be valid
         private_key.encode("utf8")
     )
     if code != 0:
@@ -191,6 +198,12 @@ def retrieve_nomcom_private_key(request, year):
 
 
 def store_nomcom_private_key(request, year, private_key):
+    """Put encrypted nomcom private key in the session store
+    
+    Encrypts the private key using openssl, then decodes the ascii-armored output
+    as utf8 and adds to the session store. Raises UnicodeError if the openssl's
+    output cannot be decoded as utf8.
+    """
     if not private_key:
         request.session['NOMCOM_PRIVATE_KEY_%s' % year] = ''
     else:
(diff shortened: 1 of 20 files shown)


richsalz:
> There was one problem with this - the nomcom private key was being stored in the session as bytes, which cannot be serialized directly as JSON. I fixed this by decoding as utf8 before stuffing into the session and encoding as utf8 when retrieving.

I do not know enough about how the system works, but let me raise the issue.  The NomCom private key isn't characters, it's an array of bytes and trying to make it into and out-of UTF8 is probably not safe.  If you need to turn key material into json, the best thing is to do to/from base64.

Hope this helps.

jennifer-richards:
> > There was one problem with this - the nomcom private key was being stored in the session as bytes, which cannot be serialized directly as JSON. I fixed this by decoding as utf8 before stuffing into the session and encoding as utf8 when retrieving.
> 
> I do not know enough about how the system works, but let me raise the issue. The NomCom private key isn't characters, it's an array of bytes and trying to make it into and out-of UTF8 is probably not safe. If you need to turn key material into json, the best thing is to do to/from base64.
> 
> Hope this helps.

Thanks for the comment - I think it's in encrypted, ascii-armored state (i.e., already base64 coded) when it's put in the session store. In the tests, the value in the session is faked with this: https://github.com/ietf-tools/datatracker/blob/23ca7a90bc364e7054059f59268f26582669f086/ietf/nomcom/factories.py#L37

In actual operation, it's stored here https://github.com/ietf-tools/datatracker/blob/23ca7a90bc364e7054059f59268f26582669f086/ietf/nomcom/utils.py#L197

and retrieved here https://github.com/ietf-tools/datatracker/blob/23ca7a90bc364e7054059f59268f26582669f086/ietf/nomcom/utils.py#L180

(the `%s` in both command lines is `openssl`, so those are blowfish encryption and decryption command lines)

The output of that command that was being handled as `bytes` because that's how we get it from `pipe`. I think it should always be ascii and the change is more about bookkeeping in Python's string handling.

ietf/nomcom/utils.py
```diff
@@ -183,7 +183,7 @@ def retrieve_nomcom_private_key(request, year):
             settings.OPENSSL_COMMAND,
             command_line_safe_secret(settings.NOMCOM_APP_SECRET)
         ),
-        private_key
+        private_key.encode("utf8")
```
rjsparks, line 186: I don't think this is going to work. The nomcom private key is part of a keypair that is generated externally, and the byte sequence is not guaranteed to be valid utf8. Maybe uuencoding first would work around the issue?

Please also check that the instructions (and explanations) at  https://github.com/ietf-tools/datatracker/blob/main/ietf/templates/nomcom/chair_help.html#L9-L76 are still accurate.

rjsparks: Sorry - I hadn't seen the exchange with Rich

jennifer-richards: This should not change anything with respect to the instructions or explanations. The change here is only in handling the data that comes back from gpg. As long as the reverse operation is done when feeding it back to gpg to decrypt, it should be transparent.

I will see about switching it for a more robust handling of arbitrary bytes, otherwise someone somewhere will change a locale and we'll get a weird bug.

rjsparks (approved):

PR #5675, opened 2023-05-22
fix: optimize queries for DocumentChangesFeed

rjsparks:
Anecdote:

For `/feed/document-changes/draft-ietf-roll-useofrplinfo`, this cuts the number of queries from 1037 to 7, and reduces the overall time it takes to render the view to 31% of the current code in main.

Several other places pass a list instead of a queryset into `augment_events_with_revision`. If they can be modified to pass a queryset, they'll see a similar performance increase.


rjsparks:
I pursued this optimization based on the traces Scout collected - this endpoint is very heavily utilized.

jennifer-richards (approved):

PR #5693, opened 2023-05-26
Improve parsing of hyphenated titles broken across lines

### LINKED ISSUE #5691: Draft submission fails validation if title is split across lines at a hyphen in text format
### Describe the issue

I try to upload this draft, which has worked in the past. Now, I get a confusing screen that, if I read to the bottom, indicates that the draft was rejected because the title is too long. The error reads:
```
Submission rejected: Text Internet-Draft title (Unilateral Opportunistic Deployment of Encrypted Recursive-to- Authoritative DNS) disagrees with XML Internet-Draft title (Unilateral Opportunistic Deployment of Encrypted Recursive-to-Authoritative DNS)
```

GitHub is not allowing me to attach the XML file that causes this error. The relevant bit is:
```
<rfc ipr="trust200902" docName="draft-ietf-dprive-unilateral-probing-06" category="std" consensus="true" submissionType="IETF">
  <front>
    <title abbrev="Unilateral Encrypted Authoritative DNS">Unilateral Opportunistic Deployment of Encrypted Recursive-to-Authoritative DNS</title>
 ```

### Code of Conduct

jennifer-richards:
This fixes #5691.

As noted in https://github.com/ietf-tools/datatracker/issues/5691#issuecomment-1563477805, this does not handle all cases of hyphens at line breaks correctly. I think this change will do the right thing for the more common case, though, so is a net improvement.

### DIFF
--- a/ietf/utils/draft.py
+++ b/ietf/utils/draft.py
@@ -1106,8 +1106,9 @@ class PlaintextDraft(Draft):
             title = match.group(1)
             title = title.strip()
             title = re.sub(r'(?s)\n\s*\<?draft-.*$','', title)
-            title = re.sub(r'\s*\n\s*', ' ', title)
-            title = re.sub(r' +', ' ', title)
+            title = re.sub(r'-\s*\n\s*', '-', title)  # replace line break preceded by hyphen with hyphen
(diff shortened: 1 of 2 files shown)


jennifer-richards:
Going to close this - will need to be part of a more comprehensive change.

PR #5709, opened 2023-05-29
fix: Fix overscrolling of rfc2html `pre` tags with new fonts

larseggert:


larseggert:
Not sure where the CI failure comes from, seems unrelated?

larseggert:
I should be current?

rjsparks (approved):

PR #5718, opened 2023-05-30
feat: disable login button on submit

NGPixel:


rjsparks (approved):

jennifer-richards (approved):
Code looks fine. Would be nice to move away from inline scripts though.

jennifer-richards (approved):

PR #5721, opened 2023-05-31
fix: Remove "purpose" column, show purpose as badge

### LINKED ISSUE #5711: "Purpose" column on https://datatracker.ietf.org/meeting/requests
### Description

@flynnliz do we need the "purpose" column on https://datatracker.ietf.org/meeting/requests? It's empty for almost all requests and not very informative for those where it's not. Removing it would free up some precious horizontal space on that page.

### Code of Conduct

larseggert:
Fixes #5711


rjsparks:
The only downside of the proposal is that it removes the ability to sort by the column

rjsparks (approved):

PR #5724, opened 2023-05-31
fix: Make sure "./run" is the invocation

richsalz:
I have made this mistake far too often :)


rjsparks:
A nit and a suggestion:

Nit: This could do the wrong thing if there happened to be some other executable named run in the current working (but unintended) directory.

Suggestion: Instead of erroring, consider modifying the script to cd to the right place internally. Experiment with `cd "$(dirname "$0")"` and the like.

rjsparks (approved):

PR #5726, opened 2023-05-31
fix: Return NomComs, not Groups, from active_nomcoms filter

### LINKED ISSUE #5723: Don't crash when a nomcom is partially set up.
Right now, if a nomcom group is created, and the associated NomCom object is not, anyone with a role in the nomcom group cannot use the datatracker - the construction of the menu crashes.

The places that crash need to be protected against this misconfiguration (especially while the configuration of the Group and NomCom objects are done manually).

See also https://github.com/ietf-tools/datatracker/issues/3289

jennifer-richards:
Fixes #5723

### DIFF
--- a/ietf/group/templatetags/group_filters.py
+++ b/ietf/group/templatetags/group_filters.py
@@ -25,6 +25,8 @@ def active_nomcoms(user):
             group__type_id='nomcom',  # just in case...
             group__state__slug='active',
         )
+        .distinct()
+        .order_by("group__acronym")
     )
 
 @register.inclusion_tag('person/person_link.html')
--- a/ietf/group/templatetags/group_filters.py
+++ b/ietf/group/templatetags/group_filters.py
@@ -2,7 +2,7 @@ from django import template
 
 import debug                            # pyflakes:ignore
 
-from ietf.group.models import Group
+from ietf.nomcom.models import NomCom
 
 register = template.Library()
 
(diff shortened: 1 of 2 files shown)


ietf/group/templatetags/group_filters.py
```diff
@@ -19,14 +19,13 @@ def active_nomcoms(user):
...
-        type_id='nomcom',
-        state__slug='active').distinct().select_related("type"))
-
-    return groups
+    return list(
+        NomCom.objects.filter(
+            group__role__person__user=user,
+            group__type_id='nomcom',  # just in case...
+            group__state__slug='active',
+        )
```
rjsparks, line 27: probably want distinct() here - several people end up with multiple roles in a NomCom.group.

rjsparks: also consider an order_by, perhaps on `group__acronym` (but maybe making the implicit `pk` ordering explicit is enough.

rjsparks: Where was the `select_related("type")` used in the existing code?

rjsparks (changes requested):

ietf/group/templatetags/group_filters.py
rjsparks: (looks like it _wasn't_. Some other code must have gone away...)

jennifer-richards: That's right, it wasn't being used. Will add the `distinct()` and an ordering

rjsparks (approved):

PR #5727, opened 2023-05-31
test: ignore acceptable order change

rjsparks:


jennifer-richards (approved):

ietf/secr/sreq/tests.py
```diff
@@ -158,7 +158,7 @@ def test_edit(self):
             list(TimerangeName.objects.filter(name__in=['thursday-afternoon-early', 'thursday-afternoon-late']).values('name'))
         )
         self.assertFalse(sessions[0].joint_with_groups.count())
-        self.assertEqual(list(sessions[1].joint_with_groups.all()), [group3, group4])
+        self.assertEqual(set(list(sessions[1].joint_with_groups.all())), set([group3, group4]))
```
jennifer-richards, line 161: Does the `set()` need the `list()` first? I'd have thought it can work straight from the `QuerySet`.

Could also use a set literal `{group3, group4}`

Up to you whether you care enough about this to change.

jennifer-richards (approved):

PR #5715, opened 2023-05-30
fix: /meeting/materials splits out named sessions the same way /meeting/proceedings does

pselkirk:
fixes #5653

PR #5741, opened 2023-06-02
fix: /meeting/materials splits out named sessions the same way /meeting/proceedings does

pselkirk:
This is an alternative to #5715, which groups sessions in the template versus in the python view. In terms of complexity, it's about the same, but is a little more performant.

Note that these branches contain the same unit test, because it's testing the same condition.

At most one of these PRs should be adopted.

### DIFF
--- a/ietf/meeting/templatetags/materials_filters.py
+++ b/ietf/meeting/templatetags/materials_filters.py
@@ -1,0 +1,47 @@
+# Copyright The IETF Trust 2023, All Rights Reserved
+from django import template
+
+import debug       # pyflakes:ignore
+import itertools
+
+register = template.Library()
+
+def _uniq(l):
+    # https://www.peterbe.com/plog/fastest-way-to-uniquify-a-list-in-python-3.6
+    l = list(dict.fromkeys(l))
+    # Filter out None return values from Session.agendas() etc.
+    # Listify the result, because the template calls length() on it.
+    return list(itertools.filterfalse(lambda x: not x, l))
+
+def _flatten(ll):
+    return [x for l in ll for x in l]
+
+@register.filter
+def all_meeting_sessions_cancelled(ss):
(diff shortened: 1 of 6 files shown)


ietf/meeting/templatetags/materials_filters.py
```diff
@@ -0,0 +1,47 @@
...
+    l = list(dict.fromkeys(l))
+    # Filter out None return values from Session.agendas() etc.
+    # Listify the result, because the template calls length() on it.
+    return list(itertools.filterfalse(lambda x: not x, l))
+
+def _flatten(ll):
+    return [x for l in ll for x in l]
+
+@register.filter
+def all_meeting_sessions_cancelled(ss):
```
rjsparks, line 20: Are the functions on the model with the same names still used anywhere? If not, they should be removed from the model.

For each of these, have you confirmed that this reimplementation results in the same ordering (which matters)?

pselkirk: - `ietf.meeting.views.materials()` still uses `all_meeting_slides()` and `all_meeting_agendas()` (but not other material types?) to determine `last_update` time.

- `ietf.meeting.views.request_minutes()` uses `all_meeting_minutes()`

This applies to both version of the fix.

pselkirk: Good catch on the ordering. Area, group, and session name are correct, but I had to add a session-time sort for same-name sessions, where I found some agendas and/or minutes out of order.

 #5715 didn't need this fix, because `organize_proceedings_sessions` applies the correect ordering out of the box.

rjsparks:
On reflection, I think we are better sticking with the utilities where they are rather than having a reimplementations in the templatetags. We can push on improving performance a different way. (A Vue version of this view will act very differently).

Issue #5773, opened 2023-06-06
Investigate potentially unused management commands


rjsparks:
Right now, nothing appears to invoke these commands. Confirm that, and discuss with leadership whether they can be removed:

* ipr/management/commands/generate_draft_recursive_txt.py
* liaison/management/commands/check_liaison_deadlines.py
* meeting/management/commands/meetecho_conferences.py
* meeting/management/commands/schedwrite.py
* stats/management/commands/find_meetingregistration_person_issues.py
* stats/management/commands/repair_meetingregistration_person.py

The commands under ietf/utils/management/commands need to be evaluated separately.

PR #5774, opened 2023-06-06
chore: remove unused mgmt commands

rjsparks:


rjsparks:
See also #5773. This PR only gets the very obviously obviated (actually dangerous to run again) commands.

jennifer-richards (approved):

Issue #5759, opened 2023-06-05
Downref registry page should provide brief explanation of what a downref is


ajeanmahoney:
### Describe the issue

There was an earlier issue (#2788) that covered this, and it was closed as fixed (commit 13fee54). However, the text does not appear on the [Downref registry](https://datatracker.ietf.org/doc/downref/) page.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

becarpenter:
Note that the missing text mentioned in #2788 is slightly incomplete, since BCP97 consists of three RFCs: 3967, 4897 and 8067.

larseggert:
Closing, since #5789 got merged.

PR #5789, opened 2023-06-08
fix: Add definition of DOWNREF from RFC3967

larseggert:
Fixes #2788, #5759


rjsparks (approved):

rjsparks:
If anyone can explain what happened to Russ' original, it would be helpful. The only thing I can come up with is that it never got merged to trunk (back when we were in svn).

PR #5795, opened 2023-06-09
feat: render charters on about page using markdown

rjsparks:
Remember the `/group/<acronym>/about/rendertest` view for a before/after comparison. (This PR fixes a js issue in that view).

The last time we pressed people to look at rendertest, I don't think we were using `nl2br`, so some of the really old charters with strange linebreaks in their text files will still just look strange, but overall, using it will likely produce the least surprise.

cc: @larseggert

### DIFF
--- a/ietf/group/views.py
+++ b/ietf/group/views.py
@@ -166,6 +166,7 @@ def fill_in_charter_info(group, include_drafts=False):
         group.charter_text = get_charter_text(group)
     else:
         group.charter_text = "Not chartered yet."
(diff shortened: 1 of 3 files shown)


jennifer-richards (approved):

larseggert:
I don't think there is anything more I can easily do. Fixing the oddities would entail switching libs and/or writing custom plugins.

rjsparks:
Mutter: foot-gunned by the GH stateful type-of-merge button once again...

PR #5816, opened 2023-06-13
refactor: Do not use canonical_name() for charters

jennifer-richards:


jennifer-richards:
Branches got mixed up, going to try to straighten that out and create a cleaner PR.

PR #5821, opened 2023-06-14
refactor: change references from User to Person

### LINKED ISSUE #2142: Change references to User
`type_defect`   |    by rcross@amsl.com

___

Several objects have references to User, these should be changed to Person:

Feedback, Nomination, CommunityList

___
_Issue migrated from trac:2142 at 2022-03-04 05:39:44 +0000_

pselkirk:
Fixes #2142


pselkirk:
This first commit converts `CommunityList`. It passes relevant unit testing, but it could user more scrutiny.

I do see an unrelated test failure in `ietf.doc.tests.PdfizedTests.test_pdfized`: `OSError: [Errno 99] Cannot assign requested address`. I don't see how it could be related, but I haven't gone back to test the `main` branch.

ietf/community/models.py
```diff
@@ -13,13 +12,23 @@
 from ietf.utils.models import ForeignKey
 
 class CommunityList(models.Model):
-    user = ForeignKey(User, blank=True, null=True)
+    person = ForeignKey(Person, blank=True, null=True)
     group = ForeignKey(Group, blank=True, null=True)
     added_docs = models.ManyToManyField(Document)
 
+    @property
```
rjsparks, line 19: Why did you need this? We should purge the app of any knowledge of user. It really should only need to know about person.

ietf/community/models.py
```diff
@@ -13,13 +12,23 @@
...
+            return self.person.user
+        except AttributeError:
+            return None
+
     def long_name(self):
-        if self.user:
-            return 'Personal I-D list of %s' % self.user.username
+        if self.person:
+            try:
+                return 'Personal I-D list of %s' % self.person.user.username
```
rjsparks, line 29: Lets change this to use self.person.plain_name

ietf/community/models.py
```diff
@@ -30,8 +39,8 @@ def __str__(self):
 
     def get_absolute_url(self):
         import ietf.community.views
-        if self.user:
-            return urlreverse(ietf.community.views.view_list, kwargs={ 'username': self.user.username })
+        if self.person:
+            return urlreverse(ietf.community.views.view_list, kwargs={ 'username': self.person.user.username })
```
rjsparks, line 43: removing the exposure of username here is also worth pursuing. We could either change this to just hit the person pk (which I would prefer to avoid) or do the same kind of thing we do at the /person/ endpoint, where we match names or email addresses. Please take a look at doing that and see if there's a strong reason not to.

rjsparks (changes requested):

ietf/community/models.py
jennifer-richards: Might be useful as a stepping stone to make it raise `RuntimeError` and run the tests to see if it's being referenced.

ietf/doc/utils.py
```diff
@@ -1059,7 +1059,7 @@ def augment_docs_and_user_with_user_info(docs, user):
                 reviewteamsettings__isnull=False, role__person__user=user, role__name='reviewer')
 
         doc_pks = [d.pk for d in docs]
-        clist = CommunityList.objects.filter(user=user).first()
+        clist = CommunityList.objects.filter(person=Person.objects.get(user=user)).first()
```
jennifer-richards, line 1062: Another place to move `user` out of the `Community` app. I think this method is only used in a couple places, so it'd probably be better to move the authentication check and `Person` lookup out of this method and into the caller. The method name and signature both need a change.

ietf/community/models.py
pselkirk: It's still in use in the templates, e.g. `{% url "ietf.community.views.feed" username=clist.user.username %}`. But I see below that you'd prefer to remove username from urls.

ietf/community/models.py
pselkirk: Okay.

ietf/community/models.py
pselkirk: Okay, I'll look into it.

ietf/api/views.py
```diff
@@ -89,7 +89,7 @@ def get(self, request):
             'sendqueue', 'nominee', 'topicfeedbacklastseen', 'alias', 'email', 'apikeys', 'personevent',
             'reviewersettings', 'reviewsecretarysettings', 'unavailableperiod', 'reviewwish',
             'nextreviewerinteam', 'reviewrequest', 'meetingregistration', 'submissionevent', 'preapproval',
-            'user', 'user__communitylist', 'personextresource_set', ]
+            'user', 'communitylist', 'personextresource_set', ]
```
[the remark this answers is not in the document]
pselkirk: I wasn't able to get this working via `curl`, but pasting the url in the browser, I see a pile of json, where `communitylist` (without a `user__` prefix) has moved out from under `user` to be directly under `person`.

ietf/community/models.py
pselkirk: Fixed the templates and a few other things that refer to user. I think the remaining references are related to usernames in requests or replies.

ietf/community/models.py
pselkirk: This ticket is about refactoring, which should be (mostly) invisible to the user. Changing the URL structure is not only user-visible, but goes way beyond the scope of this ticket. Which is not to say it's not worth doing, but it should be its own ticket.

ietf/api/views.py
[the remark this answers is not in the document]
jennifer-richards: Thanks. Sounds like an improvement.

ietf/doc/utils.py
pselkirk: Technically, this is outside of the `Community` app, but it wasn't a huge push.

PR #5825, opened 2023-06-14
fix: Don't resort a table if sort query argument is present

larseggert:
Fixes #5686

### DIFF
--- a/ietf/static/js/list.js
+++ b/ietf/static/js/list.js
@@ -273,14 +273,20 @@ $(document)
                 $(table)[0]
                     .dispatchEvent(new Event("tablesorter:done"));
 
-                // if there is a data-default-sort attribute on a column, pre-sort the table on that
-                const presort_col = $(header_row).children("[data-default-sort]:first");
-                if (presort_col) {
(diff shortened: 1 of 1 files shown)


ietf/static/js/list.js
```diff
@@ -273,14 +273,20 @@ $(document)
...
+                // check if there is a sort query argument, and leave the table alone if so
+                const params = new Proxy(new URLSearchParams(window.location.search), {
+                    get: (searchParams, prop) => searchParams.get(prop),
+                });
+                if (!params.sort) {
+                    // else, if there is a data-default-sort attribute on a column, pre-sort the table on that
+                    const presort_col = $(header_row).children("[data-default-sort]:first");
+                    if (presort_col) {
+                        const order = presort_col.attr("data-default-sort");
+                        if (order == "asc" || order == "desc") {
```
NGPixel, line 285: Avoid using `==` unless specifically checking for null | undefined. Always use `===` instead.

NGPixel (changes requested):

NGPixel (approved):

PR #5839, opened 2023-06-16
feat: initial strawberry-graphql implementation

NGPixel:


ietf/schema.py
```diff
@@ -0,0 +1,17 @@
+from typing import List
+import strawberry
+from strawberry_django_plus.optimizer import DjangoOptimizerExtension
+
+from ietf.meeting.types import Meeting, MeetingFilter
+
+@strawberry.type
```
rjsparks, line 7: Is it possible for these declarations to live with each app instead of at the top level of the project? Similar to the declarations for the admin?

PR #5841, opened 2023-06-16
feat: Remove DocAlias from RelatedDocument

rjsparks:


ietf/submit/tests.py
```diff
@@ -1142,7 +1141,7 @@ def test_submit_new_logged_in_with_extresources(self):
         self.verify_bibxml_ids_creation(draft)
 
     def test_submit_update_individual(self):
-        IndividualDraftFactory(name='draft-ietf-random-thing', states=[('draft','rfc')], other_aliases=['rfc9999',], pages=5)
+        IndividualDraftFactory(name='draft-ietf-random-thing', states=[('draft','active'),('draft-iesg','approved')], pages=5)
```
rjsparks, line 1144: This is suspicious - will review - I suspect an RFC should be being built here instead.

ietf/templates/doc/document_references.html
```diff
@@ -51,12 +51,12 @@ <h1>References from {{ doc.canonical_name }}</h1>
                             </a>
                         </td>
                         <td>
-                            {% if ref.target.document.type_id == "rfc" %}
-                                {% with ref.target.document.std_level as lvl %}
+                            {% if ref.target.type_id == 'rfc' %}
```
rjsparks, line 54: ' -> "

ietf/doc/tests_conflict_review.py
```diff
@@ -70,12 +70,12 @@ def test_start_review_as_secretary(self):
...
         self.assertEqual(review_doc.notify,'ipu@ietf.org')
         doc = Document.objects.get(name='draft-imaginary-independent-submission')
-        self.assertTrue(doc in [x.target.document for x in review_doc.relateddocument_set.filter(relationship__slug='conflrev')])
+        self.assertTrue(doc in [x.target for x in review_doc.relateddocument_set.filter(relationship__slug='conflrev')])
 
         self.assertTrue(review_doc.latest_event(DocEvent,type="added_comment").desc.startswith("IETF conflict review requested"))
         self.assertTrue(doc.latest_event(DocEvent,type="added_comment").desc.startswith("IETF conflict review initiated"))
         self.assertTrue('Conflict Review requested' in outbox[-1]['Subject'])
-        
+
```
rjsparks, line 78: remove spaces

ietf/submit/tests.py
rjsparks: No - it's correct - this reflects replaces only pointing to draft objects, and the test is otherwise adjusted to exercise the "can't replace an RFC" and "can't replace an approved draft".

PR #5852, opened 2023-06-19
fix: More dark mode fixes

larseggert:


ietf/templates/doc/document_html.html
```diff
@@ -31,6 +31,15 @@
...
             <script src="{% static 'ietf/js/theme.js' %}"></script>
         {% endif %}
+            <style>
+                {# Try and recolor black colors in SVGs for dark mode. This will need further tweaking. #}
+                svg *[fill='black'], svg text:not([fill]) {
+                    fill: currentColor;
+                }
+                svg *[stroke='black'] {
+                    stroke: currentColor;
+                }
```
martinthomson, line 41: You will need to redo white as well.  Some SVGs explicitly include white boxes, as I discovered...

See https://github.com/martinthomson/rfc-css/blob/d6c36ce15c5cc2663dd7500509dd7edc36d6e8d1/rfc.css#L824-L836

martinthomson (commented):
https://dt-dark-mode.dev.ietf.org/static/ietf/images/ietf-logo.svg probably needs tweaking also.  See the home page.

larseggert:
@martinthomson there is https://github.com/ietf-tools/datatracker/blob/feat/dark-mode/ietf/static/images/ietf-logo-white.svg already on that feature branch. (And others.)

NGPixel:
Not a dark mode thing per say, but Bootstrap 5.3 deprecates the `.text-muted` class, which is used in many places across the project. It's being replaced with `.text-body-secondary` (need to make sure it's identical or similar enough?).

larseggert:
@NGPixel we should probably do another PR that replaces deprecated things.

PR #5869, opened 2023-06-22
ci: merge main to release

rjsparks:


.github/workflows/build.yml
```diff
@@ -135,7 +135,7 @@ jobs:
           exit 1
         fi
         echo "Running tests..."
-        if [[ "x${{ github.event.inputs.ignoreLowerCoverage }}" == "xtrue" ]]; then
+        if [[ "x${{ github.event.inputs.ignoreLowerCoverage }}" == "true" ]]; then
```
jennifer-richards, line 138: Can this ever be true? (note the "x" prefix on the LHS)

jennifer-richards (changes requested):
I think maybe an erroneous typo "fix" made it in to main?

.github/workflows/build.yml
NGPixel: Yeah that's wrong. I incorrectly assumed that extra x on the RHS was a typo but it's supposed to be there. I'll "unfix" the fix on main.

jennifer-richards (approved):

PR #5878, opened 2023-06-26
feat: Display and bulk-edit nomcom position flags

### LINKED ISSUE #4681: Setting a state for all positions at once
### Description

It would be nice to set (some?) states for all positions at the same time. I understand for "special elections" you need to tweak individual setting, but I think for the common case of "nominations are closed, feedback is open" the current scheme is a bit error-prone

### Code of Conduct

pselkirk:
Fixes  #4681

### DIFF
--- a/ietf/templates/nomcom/edit_position.html
+++ b/ietf/templates/nomcom/edit_position.html
@@ -31,5 +31,14 @@
                {% if position %}Save{% else %}Add{% endif %}</button>
         <a class="btn btn-secondary float-end" href="../">Back</a>
     </form>
+    {% if position.name %}
+        <h3>
+            Templates
+        </h3>
+        {% for template in position.get_templates %}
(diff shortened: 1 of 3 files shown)


pselkirk:
Screen Shot 2023-06-26 at 2 57 59 PM

ietf/templates/nomcom/edit_position.html
```diff
@@ -31,5 +31,14 @@ <h2>
...
+        </h3>
+        {% for template in position.get_templates %}
+            <a href="{% url 'ietf.nomcom.views.edit_template' year template.id %}">{{ template }}</a>
+            <br>
+        {% endfor %}
+    {% endif %}
 {% endblock %}
-{% block content_end %}{{ form.media.js }}{% endblock %}
\ No newline at end of file
+{% block content_end %}{{ form.media.js }}{% endblock %}
```
rjsparks, line 44: Please see if you can convince your editor to leave file endings alone (particularly for templates).

rjsparks (approved):

PR #5886, opened 2023-06-27
refactor: Refactor build_file_urls()

jennifer-richards:


ietf/doc/utils.py
```diff
@@ -1039,13 +1036,14 @@ def build_file_urls(doc: Union[Document, DocHistory]):
             file_urls.append(("pdfized", urlreverse('ietf.doc.views_doc.document_pdfized', kwargs=dict(name=doc.name, rev=doc.rev))))
         file_urls.append(("bibtex", urlreverse('ietf.doc.views_doc.document_bibtex',kwargs=dict(name=doc.name,rev=doc.rev))))
     else:
-        # TODO: look at the state of the database post migration and update this comment, or remove the block
-        # As of 2022-12-14, there are 1463 Document and 3136 DocHistory records with type='draft' and rev=''.
-        # All of these are in the rfc state and are covered by the above cases.
-        log.unreachable('2022-12-14')
+        if doc.type_id == "draft":
```
rjsparks, line 1039: Why this if?

rjsparks (approved):

ietf/doc/utils.py
jennifer-richards: Yes, that seemed just begging to skew out of sync with the if cases and turn into a bug later.

PR #5895, opened 2023-06-29
feat: support iab and iesg statements. Import iab statements.

rjsparks:
I think we should merge this into a feature branch first and let the secretariat explore it before merging it to main, but would like some review before we do even that.


ietf/doc/views_statement.py
```diff
@@ -17,4 +29,247 @@ def serve_pdf(self, name, rev=None):
...
+
+        submission_method = self.cleaned_data.get("statement_submission")
+        markdown_content = ""
+        if submission_method == "enter":
+            if require_field("statement_content"):
+                markdown_content = self.cleaned_data["statement_content"].replace(
+                    "\r", ""
+                )
+                default_content = render_to_string(
+                    "doc/statement/statement_template.md", {"settings": settings}
```
jennifer-richards, line 64: Better to pass in specific settings of interest? (Or maybe wait until we address this more generally, since this is not a particularly vulnerable usage)

ietf/doc/views_statement.py
```diff
@@ -17,4 +29,247 @@ def serve_pdf(self, name, rev=None):
...
+                    )
+                elif content_type != "application/pdf":
+                    markdown_content = get_cleaned_text_file_content(
+                        self.cleaned_data["statement_file"]
+                    )
+        if markdown_content != "":
+            try:
+                _ = markdown.markdown(markdown_content)
+            except Exception as e:
+                raise forms.ValidationError(f"Markdown processing failed: {e}")
```
jennifer-richards, line 100: Might be better to log the exception instead of (or in addition to) exposing it to the UI, particularly if you're catching `Exception` instead of something more specific.

ietf/doc/views_statement.py
```diff
@@ -17,4 +29,247 @@ def serve_pdf(self, name, rev=None):
...
+            ) as destination:
+                if writing_pdf:
+                    for chunk in form.cleaned_data["statement_file"].chunks():
+                        destination.write(chunk)
+                else:
+                    destination.write(markdown_content)
+            return redirect("ietf.doc.views_doc.document_main", name=statement.name)
+
+    else:
+        if statement.uploaded_filename.endswith("pdf"):
```
jennifer-richards, line 148: Might want to use something a little more robust, `Path(...).suffix.lower()` perhaps? (Not sure it's worth it, given that it'd still be identifying the format by its extension, so the endswith might be adequate.)

ietf/doc/views_statement.py
```diff
@@ -17,4 +29,247 @@ def serve_pdf(self, name, rev=None):
...
+                default_content = render_to_string(
+                    "doc/statement/statement_template.md", {"settings": settings}
+                )
+                if markdown_content == default_content:
+                    raise forms.ValidationError(
+                        "The example content may not be saved. Edit it to contain the next revision statement content."
+                    )
+                if (
+                    markdown_content
+                    == "The current revision of this statement is in pdf format"
```
jennifer-richards, line 72: Duplicating this magic string is fragile, might want to make it a const so someone doesn't break this protection when they rephrase the message.

jennifer-richards (approved):
I have a handful of comments, all minor, inline, and optional.

docker/configs/settings_local.py
```diff
@@ -40,7 +40,7 @@
 #    'ietf.context_processors.sql_debug',
 # ]
 
-DOCUMENT_PATH_PATTERN = '/assets/ietf-ftp/{doc.type_id}/'
+DOCUMENT_PATH_PATTERN = '/assets/ietfdata/doc/{doc.type_id}/'
```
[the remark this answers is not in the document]
rjsparks: There will be other fallout for the dev system, but this needs to happen.

ietf/doc/management/commands/import_iab_statements.py
```diff
@@ -0,0 +1,288 @@
...
+from ietf.doc.models import Document, DocAlias, DocEvent
+from ietf.utils.text import xslugify
+
+class Command(BaseCommand):
+    help = "Performs a one-time import of IAB statements"
+
+    def handle(self, *args, **options):
+        tmpdir = tempfile.mkdtemp()
+        process = subprocess.Popen(
+            ["git", "clone", "https://github.com/kesara/iab-scraper.git", tmpdir],
```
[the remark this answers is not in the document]
rjsparks: not worried

ietf/doc/models.py
```diff
@@ -186,7 +186,7 @@ def get_base_name(self):
             elif self.type_id == 'review':
                 # TODO: This will be wrong if a review is updated on the same day it was created (or updated more than once on the same day)
                 self._cached_base_name = "%s.txt" % self.name
-            elif self.type_id == 'bofreq':
+            elif self.type_id in ['bofreq', 'statement']:
                 self._cached_base_name = "%s-%s.md" % (self.name, self.rev)
```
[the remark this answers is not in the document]
rjsparks: yeah - it will only get used if uploaded_filename is empty and the code is always setting uploaded_filename for this document type.

ietf/doc/views_statement.py
rjsparks: well, what Exception might get thrown in this case that isn't "Markdown processing failed"?
It's exposed in the UI so the submitter can edit whatever they're sending in that markdown is blowing up on.

ietf/static/js/upload_statement.js
```diff
@@ -0,0 +1,29 @@
+$(document)
```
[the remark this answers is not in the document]
rjsparks: The existing thing that this is just a copy/paste of would also need to be changed.
A later refactor to remove the duplication would be useful.

ietf/doc/views_doc.py
```diff
@@ -810,7 +810,23 @@ def document_main(request, name, rev=None, document_html=False):
         )
 
     if doc.type_id == "statement":
-        content = markdown.markdown(doc.text_or_error())
+        if doc.uploaded_filename:
+            basename = doc.uploaded_filename.split(".")[0] # strip extension
```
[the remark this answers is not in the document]
rjsparks: We don't take uploaded_filename from users - we compute and set it ourselves with the same restrictions we have on document name for the given document type.

ietf/doc/views_statement.py
rjsparks: removing it from the dict

ietf/doc/views_statement.py
rjsparks: again, its our code that set the uploaded_filename.

ietf/static/js/upload_statement.js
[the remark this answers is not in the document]
NGPixel: image

PR #5911, opened 2023-07-04
chore: Remove temporary pin on pydantic

jennifer-richards:
Pin is not needed since inflect 6.0.5. As of inflect 6.1.0, pydantic 2.0 is supported.

### DIFF
--- a/ietf/secr/sreq/tests.py
+++ b/ietf/secr/sreq/tests.py
@@ -146,7 +146,7 @@ class SessionRequestTestCase(TestCase):
         self.assertRedirects(r, redirect_url)
 
         # Check whether updates were stored in the database
-        sessions = Session.objects.filter(meeting=meeting, group=mars)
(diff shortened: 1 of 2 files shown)


ietf/secr/sreq/tests.py
```diff
@@ -146,7 +146,7 @@ def test_edit(self):
         self.assertRedirects(r, redirect_url)
 
         # Check whether updates were stored in the database
-        sessions = Session.objects.filter(meeting=meeting, group=mars)
+        sessions = Session.objects.filter(meeting=meeting, group=mars).order_by("id")  # order to match edit() view
```
rjsparks, line 149: In case anyone looks back and notices this PR does more than remove the pin - this test happened to show instability while testing removal of the pin. We accepted taking the fix as part of this PR intentionally.

rjsparks (approved):

PR #5919, opened 2023-07-05
fix: Show rfc documents via document_main view

jennifer-richards:
This restores basic displaying of RFC documents, adding a separate path in the view for the rfc document type. Makes a first pass at removing irrelevant information from the view.

Does not yet fix up the document timeline or rearrange the display of the document stream.


ietf/doc/views_doc.py
```diff
@@ -494,7 +594,7 @@ def document_main(request, name, rev=None, document_html=False):
 
         augment_docs_and_user_with_user_info([doc], request.user)
 
-        published = doc.latest_event(type="published_rfc")
+        published = doc.latest_event(type="published_rfc")  # todo rethink this now that published_rfc is on rfc
```
rjsparks, line 597: Correct - this should always be None for a document of type draft. If we want the draft view to say anything about when the rfc it became was published we should be looking through the relationship.

ietf/templates/doc/document_info.html
```diff
@@ -132,91 +132,93 @@
...
-                        {{ possibly_replaced_by|urlize_related_source_list:document_html|join:"<br>" }}
+                        {{ replaced_by|urlize_related_source_list:document_html|join:"<br>" }}
                     {% else %}
-                        {{ possibly_replaced_by|urlize_related_source_list:document_html|join:", " }}
+                        {{ replaced_by|urlize_related_source_list:document_html|join:", " }}
                     {% endif %}
                 </td>
             </tr>
         {% endif %}
+        {% if can_view_possibly_replaces %}
```
rjsparks, line 176: look to see if this is already in a conditional that enforces type=draft. rfc objects won't have possibly replaces.

rjsparks (approved):

rjsparks:
This looks to be on the right track to me.

ietf/templates/doc/document_info.html
```diff
@@ -11,15 +11,15 @@
         <th scope="row">{% if document_html %}Document type{% else %}Type{% endif %}</th>
         <td class="edit"></td>
         <td>
-            {% if doc.get_state_slug == "rfc" and not snapshot %}
+            {% if doc.is_rfc %}
```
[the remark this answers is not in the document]
jennifer-richards: Ok - I'd started on this road and saw you were going the type route. Decided to stick with this for this PR rather than mixing things up. I'll add switching these to the todo list.

ietf/templates/doc/document_info.html
jennifer-richards: Can you clarify what you mean? 

With the changes to the view, the `possibly_replaces` will always be empty for an rfc. This template is also used for the draft case. I've wrapped this whole section in a check of the Document type to make it clear that an RFC won't ever show it. Are you suggesting dropping that redundancy?

rjsparks: no - I was asking that you double-check/confirm "wrapped this while section in a check of the Document type" so, this is resolved.

PR #5921, opened 2023-07-05
feat: Add bibxml button to document info page

### LINKED ISSUE #5908: add  xml reference as button on document page
### Description

I always struggle to get the proper `<reference>` for a draft that I'm referencing in my own draft. I know it is available like in bib.ietf.org (and maybe other places), but I always go to the datatracker for that draft (to know its status, get the right version, ...). It would be very valuable and probably not that lot of work (given already implemented the bibtex button) to have in datatracker/doc/draft-... a new button in the "Formats" line to provide such bibxml info of that draft. 

### Code of Conduct

pselkirk:
Fixes #5908

### DIFF
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -1024,6 +1024,7 @@ def build_file_urls(doc: Union[Document, DocHistory]):
             if doc.tags.filter(slug="verified-errata").exists():
                 file_urls.append(("with errata", settings.RFC_EDITOR_INLINE_ERRATA_URL.format(rfc_number=doc.rfc_number())))
         file_urls.append(("bibtex", urlreverse('ietf.doc.views_doc.document_bibtex',kwargs=dict(name=name))))
+        file_urls.append(("bibxml", urlreverse('ietf.doc.views_doc.document_bibxml',kwargs=dict(name=name))))
     elif doc.rev:
         base_path = os.path.join(settings.INTERNET_ALL_DRAFTS_ARCHIVE_DIR, doc.name + "-" + doc.rev + ".")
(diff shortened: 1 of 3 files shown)


ietf/doc/utils.py
```diff
@@ -1024,6 +1024,7 @@ def build_file_urls(doc: Union[Document, DocHistory]):
             if doc.tags.filter(slug="verified-errata").exists():
                 file_urls.append(("with errata", settings.RFC_EDITOR_INLINE_ERRATA_URL.format(rfc_number=doc.rfc_number())))
         file_urls.append(("bibtex", urlreverse('ietf.doc.views_doc.document_bibtex',kwargs=dict(name=name))))
+        file_urls.append(("bibxml", urlreverse('ietf.doc.views_doc.document_bibxml',kwargs=dict(name=name))))
```
rjsparks, line 1027: This isn't what we want - the bibxml3 endpoint will (and should) 404 on an rfc name. Please remove just this line.

It's not perfect since then there will be no way to get the bibxml for something that has become an rfc. The solution that would send someone off to bib.ietf.org would be better at this.

rjsparks (changes requested):

rjsparks (approved):

rjsparks:
I continue to have misgivings about not pointing to bib.ietf.org right away.
@kesara - could I get a second opinion?

kesara:
> I continue to have misgivings about not pointing to bib.ietf.org right away.
> 
> @kesara - could I get a second opinion?

There's a delay of about ~3 hours for new BibXML entries to be available of bib.ietf.org.

rjsparks:
ok - we'll go with this for now then.

PR #5922, opened 2023-07-05
fix: remove GMT offset labels from timezones.js

### LINKED ISSUE #5920: Incorrect time zone labeling for regions observing Daylight/Summer time
### Describe the issue

I'm in California, which is the same time zone that IETF 117 will take place in. On the agenda (https://datatracker.ietf.org/meeting/117/agenda), the Meeting and Local time zones are labeled for me as "(GMT-8:00) Pacific Time - Los Angeles."

We're currently observing Daylight Savings time, so the current offset for Pacific time is GMT-7:00.

When I look at the agenda in UTC time, it appears correctly as I would expect it to, with a 7 hour offset from Pacific time, so it appears that only the labeling is incorrect, and not the timezone itself. 

The label appears to be wrong for all regions current observing Daylight/Summer time.

### Code of Conduct

NGPixel:
Fixes #5920


client/shared/timezones.js
```diff
@@ -1,252 +1,252 @@
...
+  { label: 'Asia - Ulaanbaatar', value: 'Asia/Ulaanbaatar' },
+  { label: 'Western Time - Perth', value: 'Australia/Perth' },
+  { label: 'Asia - Pyongyang', value: 'Asia/Pyongyang' },
+  { label: 'Asia - Dili', value: 'Asia/Dili' },
+  { label: 'Asia - Jayapura', value: 'Asia/Jayapura' },
+  { label: 'Asia - Moscow+06 - Yakutsk', value: 'Asia/Yakutsk' },
+  { label: 'Pacific - Palau', value: 'Pacific/Palau' },
+  { label: 'Asia - Seoul', value: 'Asia/Seoul' },
+  { label: 'Asia - Tokyo', value: 'Asia/Tokyo' },
+  { label: 'Central Time - Darwin', value: 'Australia/Darwin' },
```
rjsparks, line 219: Should Australia be part of the label here?

client/shared/timezones.js
```diff
@@ -1,252 +1,252 @@
...
+  { label: 'Europe - Minsk', value: 'Europe/Minsk' },
+  { label: 'Europe - Moscow+00 - Moscow', value: 'Europe/Moscow' },
+  { label: 'Africa - Nairobi', value: 'Africa/Nairobi' },
+  { label: 'Asia - Qatar', value: 'Asia/Qatar' },
+  { label: 'Asia - Riyadh', value: 'Asia/Riyadh' },
+  { label: 'Antarctica - Syowa', value: 'Antarctica/Syowa' },
+  { label: 'Asia - Tehran', value: 'Asia/Tehran' },
+  { label: 'Asia - Baku', value: 'Asia/Baku' },
+  { label: 'Asia - Dubai', value: 'Asia/Dubai' },
+  { label: 'Mahe', value: 'Indian/Mahe' },
```
rjsparks, line 163: Should Indian be part of the label here?

rjsparks (approved):

PR #5930, opened 2023-07-06
fix: restored where an rfc came from to the main rfc view

rjsparks:


ietf/doc/views_doc.py
```diff
@@ -329,6 +329,32 @@ def document_main(request, name, rev=None, document_html=False):
...
+        elif group.type_id == "individ":
+            submission = "individual"
+        elif group.type_id == "area" and doc.stream_id == "ietf":
+            submission = "individual in %s area" % group.acronym
+        else:
+            if group.features.acts_like_wg and not group.type_id == "edwg":
+                submission = "%s %s" % (group.acronym, group.type)
+            else:
+                submission = group.acronym
+            submission = '<a href="%s">%s</a>' % (group.about_url(), submission)
```
jennifer-richards, line 345: We ought to get in the habit of using `mark_safe()` / `format_html()` when we create safe strings instead of `| safe` in the templates.

jennifer-richards (approved):
One comment that's more of a vent than anything...

ietf/doc/views_doc.py
rjsparks: Fair - this is copypasta (essentially) - the change would be good to apply in a comprehensive pass per view.

PR #5932, opened 2023-07-06
fix: Preserve RelatedDocuments for RFCs

jennifer-richards:
The new migration is a bit slow (but less than 2 minutes on my dev machine). May want to look into combining / optimizing those when we have a full set.


rjsparks (approved):

rjsparks:
I have some ideas about how to speed this up, but we are talking about a large number of new rows being added for the reference relationships.

PR #5935, opened 2023-07-06
fix: repair various doc view and tests

rjsparks:


ietf/doc/tests.py
```diff
@@ -1608,7 +1608,7 @@ def test_status_change(self):
         r = self.client.get(urlreverse("ietf.doc.views_doc.document_main", kwargs=dict(name=statchg.name)))
         self.assertEqual(r.status_code, 200)
         r = self.client.get(urlreverse("ietf.doc.views_doc.document_main", kwargs=dict(name=statchg.relateddocument_set.first().target.document)))
-        self.assertEqual(r.status_code, 302)
+        self.assertEqual(r.status_code, 200) # What was this even trying to prove?
```
jennifer-richards, line 1611: If you haven't, it probably worth checking that it's outputting the expected document here and not falling back to some other page (possibly with an error) and returning a 200. There are more than a few places where a redirect return code is expected.

The fact that this is following a related document makes me a bit wary of assuming it's returning the right thing with a changed code.

jennifer-richards (changes requested):

ietf/doc/tests.py
rjsparks: The test is pretty much a "did this not crash" test - there are _many_ things it could test that it is not currently testing. The 302 previously was because it passed the draft name to something that had become and RFC into the view - the answer to the comment I added is that it was testing that redirecting to the RFC worked, which had very little to do with this view.

jennifer-richards: Ok, as long as it's understood I'm happy.

jennifer-richards (approved):

PR #5940, opened 2023-07-07
feat: IAB statements

rjsparks:


requirements.txt
```diff
@@ -72,3 +72,4 @@ Unidecode>=1.3.4
 weasyprint>=59
 xml2rfc>=3.12.4
 xym>=0.6,<1.0
+pydantic<2 # Temporary pin until inflect can catch up.
```
jennifer-richards, line 75: Don't want this to creep back in

docker/configs/settings_local.py
```diff
@@ -40,7 +40,7 @@
 #    'ietf.context_processors.sql_debug',
 # ]
 
-DOCUMENT_PATH_PATTERN = '/assets/ietf-ftp/{doc.type_id}/'
+DOCUMENT_PATH_PATTERN = '/assets/ietfdata/doc/{doc.type_id}/'
```
jennifer-richards, line 43: Should we move the assets around on the dev container before merging this, or deal with it later? (Earlier you said there'd be other fallout -- just confirming whether this is intentional or was forgotten)

jennifer-richards (changes requested):
One fix needed, plus a question

jennifer-richards (approved):

PR #5941, opened 2023-07-07
fix: various fixes in ietf.doc

rjsparks:


ietf/doc/tests.py
```diff
@@ -2705,24 +2705,27 @@ def should_404(self, argdict):
         r = self.client.get(url)
         self.assertEqual(r.status_code, 404)
 
+    # This takes a _long_ time (32s on a 2022 m1 macbook pro) - is it worth what it covers?
```
jennifer-richards, line 2758: Seems like a lot of permutations being checked, maybe cut it down to just the pdf extension? Another option would be to mock out the pdf generation and just check that it calls the expected method.

jennifer-richards (approved):

PR #5963, opened 2023-07-13
fix: restore ability to create status change documents

### LINKED ISSUE #5962: status change workflow has case sensitivity errors
Creating a new status-change document currently fails with errors like this:

image

This is due to the form confusing how it displays a document name with how it searches for a document name, and postgresql searches being case sensitive. (A search for a DocAlias of "RFC3261" fails).

The tests have not exposed this, so they need to be inspected across the entire status-change workflow.

rjsparks:
Fixes #5962

### DIFF
--- a/ietf/doc/views_status_change.py
+++ b/ietf/doc/views_status_change.py
@@ -685,7 +685,7 @@ def last_call(request, name):
     form = LastCallTextForm(initial=dict(last_call_text=escape(last_call_event.text)))
 
     if request.method == 'POST':
-        if "save_last_call_text" in request.POST or "send_last_call_request" in request.POST and status_change.ad is not None:
+        if "save_last_call_text" in request.POST or ("send_last_call_request" in request.POST and status_change.ad is not None):
             form = LastCallTextForm(request.POST)
             if form.is_valid():
                 events = []
--- a/ietf/doc/views_status_change.py
+++ b/ietf/doc/views_status_change.py
@@ -418,7 +418,7 @@ def clean_helper(form, formtype):
         rfc_fields = {}
         status_fields={}
         for k in sorted(form.data.keys()):
-            v = form.data[k]
+            v = form.data[k].lower()
             if k.startswith('new_relation_row'):
                 if re.match(r'\d{1,4}',v):
                     v = 'rfc'+v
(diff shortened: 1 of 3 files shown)


pselkirk (approved):
As a purely stylistic matter, I tend to prefer "truthy" tests like `if status_change.ad` vs. `if status_change.ad is not None` or `{% if not doc.ad %}` vs. `{% if doc.ad is None %}`.

That said, it appears to do what it says on the label.

rjsparks:
> As a purely stylistic matter, I tend to prefer "truthy" tests
 
I've been getting pressure to make more explicit checks when the real point of consideration is whether the thing is None or not. In this particular case, I don't think it makes a difference, but if someone configured an AD with a _str_ that returned "", there are other kinds of conditionals or use of {{ ad }} that would likely do the wrong thing.

jennifer-richards:
> I've been getting pressure to make more explicit checks, particularly in templates, when the real point of consideration is whether the thing is None or not.

As a source of this pressure, I prefer it because it distinguishes the absence of value from the possible presence of a value that happens to evaluate as `False`. It's largely stylistic, but anecdotally I've run into more bugs caused by a surprising `False` value than by inappropriately testing for `is None`.

It's a little tricky in templates sometimes - Django represents the absence of a value in some of its fields (like `CharField`) as an empty string rather than `None`, so it's impractical to be too dogmatic.

jennifer-richards:
In #5962, you said

> The tests have not exposed this, so they need to be inspected across the entire status-change workflow.

Are you satisfied with having looked over that for this PR or should we plan that inspection separately?

rjsparks:
> Are you satisfied with having looked over (the tests) or should we plan that inspection separately?

I'm not really satisfied, but I did make a quick pass through looking for case-sensitivity troubles. I think this particular fail was a late surprise because of the way we test forms (poking hand-written POST values in that may not actually reflect what a browser would have provided exercising the defined form).  So I think it's enough of an edge that we move on, and watch for a framework that works that would allow more of an integrated test of the use of the form.

ietf/doc/views_status_change.py
```diff
@@ -685,7 +685,7 @@ def last_call(request, name):
     form = LastCallTextForm(initial=dict(last_call_text=escape(last_call_event.text)))
 
     if request.method == 'POST':
-        if "save_last_call_text" in request.POST or "send_last_call_request" in request.POST:
+        if "save_last_call_text" in request.POST or "send_last_call_request" in request.POST and status_change.ad is not None:
```
jennifer-richards, line 688: Parentheses might be in order here - this is `if A or (B and C):` I think, which makes sense but it'd help readability to make the intent more explicit

jennifer-richards (approved):
One suggested syntax change to consider, but looks good to me.

PR #5985, opened 2023-07-18
fix: #timeline -> #doc-timeline

### LINKED ISSUE #5923: #timeline CSS rule leaks
### Describe the issue

There is a rule somewhere in the datatracker code.
```css
#timeline {
  font-size: small;
}
```

This leaks in surprising ways, see https://datatracker.ietf.org/nomcom/2023/#timeline

This took me some time to diagnose.

### Code of Conduct

larseggert:
Fixes #5923

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -272,7 +272,7 @@ th,
 }
 
 // Styles for d3.js graphical SVG timelines
-#timeline {
+#doc-timeline {
     font-size: small;
 
     .axis path,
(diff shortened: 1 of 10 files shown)


rjsparks (approved):

rjsparks:
@larseggert - Something to think about: the current approach only allows one timeline on a page (because it is tied to id). I wonder if this should be a class instead. Perhaps a page showing the timelines for all documents adopted by a working group would be informative/useful?

PR #6002, opened 2023-07-21
feat: Reclassify nomcom feedback

### LINKED ISSUE #4669: nomcom chair should be able to reclassify feedback
Right now, once feedback has been classified, the classification can only be changed by directly manipulating the database.

Instead, there should be a view that lets the chair alter the classification of a message.

When providing this, also add a new classification of "overcome by events" or similar so that, for instance, a chair can replace a previously provided questionnaire response with an updated one. Any OBE bits of feedback must remain visible to the nomcom members - consider an OBE tab along with 
image

(There's almost certainly a better name than OBE)

cc @richsalz

pselkirk:
Fixes #4669


pselkirk:
Still to be done:

- Test cases.
- Robert asked for an "overcome by events" classification.

richsalz:
So this adds a "reclassify" action, that puts already-classified feedback back in the work queue.  That's a reasonable approach. This looks fine to me.

jennifer-richards (changes requested):
I think the UI needs to indicate that it's changed from "viewing" to "reclassifying". As it is, it's hard to see that the "Reclassify feedback" menu item actually had any effect or that clicking on the individual items will do other than view the feedback.

I wonder if a button on each individual feedback item's view page would be easier to implement in a way that's clearly visible to the user. I.e., open the "View feedback" tab, click the feedback item, then click the "reclassify" button on it to go to a confirmation view.

If you stick with the batch reclassify mode, there are also some glitches on the `view_feedback_nominee.html` template - the feedback type tabs don't seem to do anything. That should be cleaned up.

ietf/nomcom/views.py
```diff
@@ -956,15 +960,30 @@ def view_feedback_unrelated(request, year):
...
-                               'nomcom': nomcom})
+                               'nomcom': nomcom,
+                               'is_chair': is_chair,
+                               'reclassify': reclassify,
+                               })
 
 @role_required("Nomcom")
 @nomcom_private_key_required
-def view_feedback_topic(request, year, topic_id):
+def view_feedback_topic(request, year, topic_id, reflassify=False):
```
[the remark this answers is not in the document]
pselkirk: Right, that's from an earlier iteration.

ietf/templates/nomcom/view_feedback_nominee.html
```diff
@@ -20,11 +20,26 @@ <h2>Feedback about {{ nominee }}</h2>
...
+                <form class="form-inline" id="batch-action-form-{{ ft.slug }}" method="post">
+                {% csrf_token %}
+            {% endif %}
             <div role="tabpanel" class="tab-pane {% if forloop.first %} active{% endif %}"
                  id="{{ ft.slug }}">
                 {% for feedback in nominee.feedback_set.all %}
                     {% if feedback.type.slug == ft.slug %}
                         <dl class="row">
+                            {% if is_chair and nomcom.group.state_id == 'active' and reclassify and nominee.feedback_set.all|length > 0 %}
+                            <dt class="col-sm-2">
```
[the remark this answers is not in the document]
pselkirk: An earlier iteration had the reclassification turned on full-time for the chair, and I felt that checkboxes would be less visually intrusive than a `Reclassify` button on each message. I eventually convinced myself that reclassification would be (or at least should be) infrequent enough that it should need to be explicitly enabled. Also for safety. So maybe now we want a `Reclassify` button at the bottom of each message, instead of a checkbox at the top.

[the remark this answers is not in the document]
jennifer-richards: I later looked at a list that had more than one feedback item in the table and the checkbox was clearer than when I had only a single item, so maybe it's ok. Really the thing that concerned me was the lack of indication that the UI was in "reclassify mode". My suggestion of adding in the buttons could well add too much clutter - it was just a thought.

pselkirk:
Robert asked for an "overcome by events" feedback type, which was one of the motivators for this feature, but I feel like it's orthogonal to the actual functionality of reclassification. That might be better handled in a separate ticket.

rjsparks:
It's what things will often be reclassified _to_.
I'd like to see it as part of this PR, but a companion that merged near the same time would be sufficient

ietf/nomcom/views.py
```diff
@@ -924,47 +939,85 @@ def view_feedback_pending(request, year):
...
+        feedback_id = request.POST.get('feedback_id', None)
+        feedback = get_object_or_404(Feedback, id=feedback_id)
+        type = request.POST.get('type', None)
+        if type:
+            if type == 'unclassified':
+                feedback.type = None
+                messages.success(request, 'The selected feedback has been de-classified. Please reclassify it in the Pending emails tab.')
+            else:
+                feedback.type = FeedbackTypeName.objects.get(slug=type)
+                messages.success(request, 'The selected feedback has been reclassified.')
```
jennifer-richards, line 968: Minor, but maybe add the new classification type to the success message?

ietf/nomcom/views.py
```diff
@@ -976,17 +1029,50 @@ def view_feedback_topic(request, year, topic_id):
...
+        feedback = get_object_or_404(Feedback, id=feedback_id)
+        type = request.POST.get('type', None)
+        if type:
+            if type == 'unclassified':
+                feedback.type = None
+                feedback.nominees.clear()
+                messages.success(request, 'The selected feedback has been de-classified. Please reclassify it in the Pending emails tab.')
+            else:
+                feedback.type = FeedbackTypeName.objects.get(slug=type)
+                messages.success(request, 'The selected feedback has been reclassified.')
```
jennifer-richards, line 1056: Again add the type to the success message?

ietf/templates/nomcom/reclassify_feedback_item.html
```diff
@@ -0,0 +1,103 @@
...
+                                    </div>
+                                </div>
+                            </div>
+                        </div>
+                    </td>
+                </tr>
+            </tbody>
+        </table>
+        <input type="hidden" name="feedback_id" value="{{ feedback.id }}">
+        <button class="btn btn-primary" type="submit">Classify</button>
```
jennifer-richards, line 102: I think the button should perhaps say "Save" rather than "Classify"?

ietf/nomcom/views.py
```diff
@@ -924,47 +939,85 @@ def view_feedback_pending(request, year):
...
+@role_required("Nomcom Chair", "Nomcom Advisor")
+@nomcom_private_key_required
+def reclassify_feedback_topic(request, year, topic_id):
+    if request.method == 'POST':
+        feedback_id = request.POST.get('feedback_id', None)
+        feedback = get_object_or_404(Feedback, id=feedback_id)
+        feedback.type = None
+        feedback.topics.clear()
+        feedback.save()
+        messages.success(request, 'The selected feedback has been de-classified. Please reclassify it in the Pending emails tab.')
```
jennifer-richards, line 1011: Do you intentionally only handle clearing the feedback type for "topic" feedback? Why not allow reclassification as for the other flavors?

jennifer-richards (changes requested):
I have a number of small things, plus a couple non-trivial requests. One (the `HTTP_REFERER` business) I think needs to be sorted out now, the other (using a `ModelForm`) could wait if you're not enthusiastic to jump on it.

I like the "Reclassify" button format, but I'm wondering whether we really need the reclassify mode at all. Is there any reason not to, say, shrink the button a little and perhaps relocate it, then just always have it present? I think that would do away with the need to deal with the link in the base template and eliminate the need that led to the `HTTP_REFERER` business.

ietf/nomcom/views.py
pselkirk: The logic here is to reclassify a feedback item within its category (nominee, topic, unrelated). e.g. Nominee feedback can be reclassified from `questio` to `obe` (the canonical use case), while keeping its association with the nominee. Or it can be set to `unclassified`, in which case we wipe the type and nominees, and send it back to the pending queue.

Topic feedback only allows type `comment`, so the reclassification options would be `comment` (leave it unchanged) or `unclassified`. If the chair wants to reclassify a topic comment, the only reasonable option would seem to be to send it back to the pending queue, where it could be attached to a different topic, or marked as read, I don't know.

ietf/templates/nomcom/reclassify_feedback_item.html
pselkirk: IDK, the form for pending email says "Classify", and this form is reached by the "Reclassify" button.

pselkirk:
> I like the "Reclassify" button format, but I'm wondering whether we really need the reclassify mode at all. Is there any reason not to, say, shrink the button a little and perhaps relocate it, then just always have it present?

My feeling was that reclassification would be uncommon enough that a) there should be an explicit action to initiate it, and b) having the button there for the 99.99% of the time that you're not reclassifying would be visual clutter.

OTOH, I'm not the target user. @richsalz @martinthomson

ietf/nomcom/views.py
```diff
@@ -838,7 +851,9 @@ def nominee_staterank(nominee):
                                'topics_feedback': topics_feedback,
                                'independent_feedback': independent_feedback,
                                'nominees_feedback': nominees_feedback,
-                               'nomcom': nomcom})
+                               'nomcom': nomcom,
+                               'is_chair_task': reclassify,
+                               })
```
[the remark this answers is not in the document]
pselkirk: ??

martinthomson:
I'd have to see this in action to be able to reach any sort of conclusion.

richsalz:
It's easy to ignore a button if it's always there.  If it comes and goes, that's a problem.

ietf/templates/nomcom/reclassify_feedback_item.html
jennifer-richards: Ok - I guess my issue is with the existing UI then. :-) Leave it matching.

jennifer-richards:
> My feeling was that reclassification would be uncommon enough that a) there should be an explicit action to initiate it, and b) having the button there for the 99.99% of the time that you're not reclassifying would be visual clutter.

I agree it'll likely be uncommon, but I think that's a reason to keep it closer at hand in the usual view. With the button always present, when the chair comes across a piece of feedback that's incorrectly classified they won't have to remember to look through the "chair tasks" menu and then find the feedback again.

I think shrinking the button will reduce the clutter to a tolerable level. I can imagine we eventually associate a pulldown menu with each feedback item that has the reclassify view as one of a few options (but, emphatically, am not proposing that right now!).

As a bonus it'll simplify the views that are doing double-duty here.

jennifer-richards (approved):
Looks good to me!

Your worry about the "Reclassify" being a little obtrusive (even with the small button) is a reasonable one, but I think this is a good starting point. If it annoys the chair, we can restyle the view after we have some experience with how this is used.

richsalz (approved):
We don't need "button flavor of the month" :)  This is good.  Thanks!  And future nomcom chairs will thank you, too.

rjsparks (approved):

Issue #5972, opened 2023-07-16
/person/<name or email> endpoint should search case-insensitive


rjsparks:
Right now `/person/robert sparks` 404s.

Godseesall:
![Screenshot_20230717_162719_Chrome](https://github.com/ietf-tools/datatracker/assets/139814447/6516d2d2-63e5-4fe1-8ba6-b7eab620a626)

Godseesall:
I find this on my phone i have someone that has ben following me and i have the meaner on me okease help me

Godseesall:
![Screenshot_20230717_162712_Chrome](https://github.com/ietf-tools/datatracker/assets/139814447/4fbe7418-4118-43c3-8d5b-886b4d5f2f77)

Godseesall:
![Screenshot_20230717_162719_Chrome](https://github.com/ietf-tools/datatracker/assets/139814447/0c8af1ce-0c16-40c9-a9ab-5d6afc49759b)

rjsparks:
closed by #6007

PR #6007, opened 2023-07-22
fix: Make name/email lookups case-insensitive (#5972)

richsalz:
Use icontains so that looking up name or email is case insensitive Added a test

### DIFF
--- a/ietf/person/views.py
+++ b/ietf/person/views.py
@@ -69,11 +69,11 @@ def ajax_select2_search(request, model_name):
 
 
 def profile(request, email_or_name):
-    aliases = Alias.objects.filter(name=email_or_name)
+    aliases = Alias.objects.filter(name__icontains=email_or_name)
     persons = set(a.person for a in aliases)
 
     if '@' in email_or_name:
-        emails = Email.objects.filter(address=email_or_name)
+        emails = Email.objects.filter(address__icontains=email_or_name)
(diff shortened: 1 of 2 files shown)


jennifer-richards:
Would using `__iexact` instead of `__icontains` fix the issue without opening this up as you described in #6008 ?

jennifer-richards:
Would using `__iexact` instead of `__icontains` fix this without opening up a search like you describe in #6008?

richsalz:
Yes, that would fix the web search, but `ajax_select2_search` still uses `__icontains` so the API would still have the problem. And searching for people by things like `salz` is useful.  (I thought there was an open issue for that, but I can't find it.)

ietf/person/views.py
```diff
@@ -69,11 +69,11 @@ def ajax_select2_search(request, model_name):
 
 
 def profile(request, email_or_name):
-    aliases = Alias.objects.filter(name=email_or_name)
+    aliases = Alias.objects.filter(name__icontains=email_or_name)
```
rjsparks, line 72: how about iexact instead of icontains in both places?

richsalz: added commit to do this.
(It was useful to each just 'salz'  but not worth the #6008 risk)

rjsparks (approved):

PR #6014, opened 2023-07-22
fix: return 404 for non-existing revisions

### LINKED ISSUE #5934: doc url parser accepts revision for rfcs
/doc/rfc8989/99/ works by unintended consequence (redirecting to /doc/rfc8989.
/doc/rfc8989/99/bibtex works (producing bibtex for rfc8989.

Both should 404 instead.

maybe-hello-world:
Links to non-existing revisions to docs should return 404
Closes #5934


ietf/doc/views_doc.py
```diff
@@ -1103,6 +1106,8 @@ def document_bibtex(request, name, rev=None):
             if rev == h.rev:
                 doc = h
                 break
+        else:
+            raise Http404()
```
jennifer-richards, line 1110: This is a good place to return an error, but I think this view should also outright reject an rfc document with any rev (as you did for `document_main()`)

jennifer-richards (changes requested):
Looks good, but I think one small addition is needed. See inline comment.

ietf/doc/views_doc.py
maybe-hello-world: Addressed

maybe-hello-world:
Feel free to squash before merging to simplify history if needed :sweat_smile:

jennifer-richards:
> Feel free to squash before merging to simplify history if needed :sweat_smile: 

:-) that's our usual practice, otherwise our history would be more, uh, "interesting."

jennifer-richards (approved):
Looks good. Your last change, I think, restores the behavior of returning the current bibtex for a request for a (non RFC) draft with a nonexistent revision. That seems fine to me.

rjsparks (approved):

PR #6021, opened 2023-07-22
fix:no history entry when changing RFC Editor note for doc

PriyankaGitCom:


ietf/doc/views_ballot.py
```diff
@@ -721,7 +721,7 @@ def ballot_rfceditornote(request, name):
                 e = WriteupDocEvent(doc=doc, rev=doc.rev, by=login)
                 e.by = login
                 e.type = "changed_rfc_editor_note_text"
-                e.desc = "RFC Editor Note was changed"
+                e.desc = "RFC Editor Note was changed to " + t
```
rjsparks, line 724: I think we need a newline before t starts. I suggest using
f"RFC Editor Note was changed to \n{t}" instead.

rjsparks (changes requested):

ietf/doc/views_ballot.py
PriyankaGitCom: Done the changes.

rjsparks (approved):

PR #6022, opened 2023-07-22
fix: Shows requested reviews for doc fixes

### LINKED ISSUE #4881: Requested directorate reviews are not shown on the document status page until (?)
### Describe the issue

When as a chair I request directorate reviews for a draft, initially no indication that any review is requested is shown on the draft's status page until... I guess until the corresponding request is processed by the directorate's secretary and is assigned? Or something else? Anyway, later I can see that reviews are requested and see which of them are completed. 
The problem is that until that point there is no indication on document's status page that reviews are requested, apart from the document's history, which is a bit inconvenient. The fellow co-chair might accidently request the same reviews again when they are in fact already requested...
It would be good if the information that a review is requested (but probably not yet assigned) appears on the document's status page right after the request is done.

### Code of Conduct

kivinen:
Adds requested reviews to the document information page. Moved the due date of the incomplete reviews out from the badge to the actual text. 

Fixes #4881.

### DIFF
--- a/ietf/templates/doc/document_info.html
+++ b/ietf/templates/doc/document_info.html
@@ -351,6 +351,9 @@ href="{% url 'ietf.doc.views_draft.review_possibly_replaces' name=doc.name %}">E
                 {% for review_assignment in review_assignments %}
                     {% include "doc/review_assignment_summary.html" with current_doc_name=doc.name current_rev=doc.rev %}
                 {% endfor %}
+                {% for review_request in review_requests %}
+                    {% include "doc/review_request_summary.html" with current_doc_name=doc.name current_rev=doc.rev %}
+                {% endfor %}
                 {% if no_review_from_teams %}
                     {% for team in no_review_from_teams %}
                         {{ team.acronym.upper }}{% if not forloop.last %},{% endif %}
--- a/ietf/templates/doc/document_info.html
+++ b/ietf/templates/doc/document_info.html
@@ -349,10 +349,10 @@ href="{% url 'ietf.doc.views_draft.review_possibly_replaces' name=doc.name %}">E
             </td>
             <td>
                 {% for review_assignment in review_assignments %}
-                    {% include "doc/review_assignment_summary.html" with current_doc_name=doc.name current_rev=doc.rev %}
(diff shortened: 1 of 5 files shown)


ietf/templates/doc/document_info.html
```diff
@@ -351,6 +351,9 @@
                 {% for review_assignment in review_assignments %}
                     {% include "doc/review_assignment_summary.html" with current_doc_name=doc.name current_rev=doc.rev %}
                 {% endfor %}
+                {% for review_request in review_requests %}
+                    {% include "doc/review_request_summary.html" with current_doc_name=doc.name current_rev=doc.rev %}
```
larseggert, line 355:
```suggestion
                    {% include "doc/review_request_summary.html" with current_doc_name=doc.name current_rev=doc.rev only %}
```

larseggert (changes requested):

ietf/templates/doc/document_info.html
kivinen: I have no idea what that change would do so can't really comment on that.

rjsparks: it passes only the current_doc_name and current_rev to the review_request_summary template instead of the full context available to the template you are calling `{% include` from.

rjsparks: but I'm really confused - review_request_summary.html only uses `review_request`, why are you passing current_doc_name and current_rev to it?

kivinen: Because it is cut & paste from the above where it was giving same parameters to the doc/review_assignment_symmary.html.

I copied the review_assignment_summary.html as a base for the review_request_summary.html and at that point I did not know whether the lines using current_doc stays in or not.

In the end I did not need them as they are used to remove the version number from the line telling which version the review was done for in case it is current revision and document.

So most likely I should simply change it to 

{% include "doc/review_request_summary.html" with review_request only %}

or do I need to use review_request=review_request before only?

rjsparks: you need to use `review_request=review_request`

kivinen: Ok, now this should be done. I also did the same for the review_assignment_summary include.

rjsparks (approved):

PR #6024, opened 2023-07-22
feat: replace references to User with references to Person

rjsparks:


rjsparks:
Leaving some notes here from a deep dive into the edge cases.

There are users without persons. 

These migrations currently would lose some data about who (for a very poor value of who - all we would lose is a username since there's no person connected already) made nominations in very old nomcoms. That data is supposed to have been removed already anyhow, so I'm not particularly concerned about that.

There are communitylists attacted to users without persons. Unless these also have EmailSubscriptions, they don't do anything anyhow and deleting them is save. There is only one that has an EmailSubscription attached at the moment. I'll apppoach the owner and adjust that one away.

Then we  should add explicitly removing the personless objects instead of migrating them to person=None - the migtations will need to be updated.

jennifer-richards:
> There are communitylists attacted to users without persons. Unless these also have EmailSubscriptions, they don't do anything anyhow and deleting them is safe. There is only one that has an EmailSubscription attached at the moment. I'll appoach the owner and adjust that one away.
> 

This appears to have been done. On production:
```python
>>> CommunityList.objects.filter(user__person__isnull=True).exclude(user__isnull=True).exclude(emailsubscription__isnull=True).count()
0
```

jennifer-richards:
> These migrations currently would lose some data about who (for a very poor value of who - all we would lose is a username since there's no person connected already) made nominations in very old nomcoms. That data is supposed to have been removed already anyhow, so I'm not particularly concerned about that.

Looks like there are only 5 nominations affected if I'm understanding this correctly. The most recent was in 2018.
```sql
datatracker=> SELECT COUNT(*)
datatracker-> FROM nomcom_nomination nom
datatracker->          LEFT OUTER JOIN person_person p USING (user_id)
datatracker-> WHERE nom.user_id IS NOT NULL
datatracker->   AND p.id IS NULL;
 count
-------
     5
(1 row)
```
There are also 51 nominations with no `User` at all, the latest in 2019.

jennifer-richards:
Looks like there are also 199 `Feedback` instances tied to Personless users. I'm guessing these are in the same category as the `Nomination` instances - there's still a field indicating the author, there's just no link to a Person. The latest of these were in 2021.

jennifer-richards (approved):

PR #6027, opened 2023-07-22
fix: Add editorial stream to proceedings

kesara:
Fixes #5717

### DIFF
--- a/ietf/meeting/tests_views.py
+++ b/ietf/meeting/tests_views.py
@@ -7632,6 +7632,13 @@ class ProceedingsTests(BaseMeetingTestCase):
             'Correct title and link for each ProceedingsMaterial should appear in the correct order'
         )
 
+    def _assertGroupSessions(self, response, meeting):
+        """Checks that group/sessions are present"""
+        pq = PyQuery(response.content)
+        sections = ["plenaries", "gen", "iab", "editorial", "irtf", "training"]
+        for section in sections:
+            self.assertEqual(len(pq(f"#{section}")), 1, f"{section} section should exists in proceedings")
+
     def test_proceedings(self):
         """Proceedings should be displayed correctly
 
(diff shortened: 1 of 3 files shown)


ietf/meeting/tests_views.py
```diff
@@ -7645,6 +7652,20 @@ def test_proceedings(self):
...
+        # Add various group sessions
+        groups = []
+        parent_groups = [
+                GroupFactory.create(type_id="area", acronym="gen"),
+                GroupFactory.create(acronym="iab"),
+                GroupFactory.create(acronym="irtf"),
+                ]
+        for parent in parent_groups:
+            groups.append(GroupFactory.create(parent=parent))
+        for acronym in ["rsab", "edu"]:
```
rjsparks, line 7664: why edu?

kesara: To test the `training` section. But I just noticed that I haven't added `training` to the section test. I'll add that.

rjsparks: edu is closed. The current instantiation is emodir

ietf/templates/meeting/proceedings.html
```diff
@@ -147,6 +147,36 @@ <h2 class="mt-5" id="iab">
             </tbody>
         </table>
     {% endif %}
+    <!-- Editorial Sessions -->
```
rjsparks, line 150: This should go after irtf

rjsparks (changes requested):

ietf/meeting/tests_views.py
kesara: Existing code filter for **training** as follows: `sessions.filter(group__acronym__in=['edu','iaoc'], type_id__in=['regular', 'other',])`.
Does that need an update?

ietf/templates/meeting/proceedings.html
kesara: Fixed with 41aff3d42c62fd356c16fe42277b394ac6cdbb87

ietf/meeting/tests_views.py
kesara: @rjsparks [emodir](https://datatracker.ietf.org/group/emodir/about/) is conlcluded as well?

Adding `emodir` to the test fails the current logic for the learning section, So that logic definitely needs to be updated.

rjsparks: bah - the current is `eodir`

rjsparks (approved):

PR #6041, opened 2023-07-24
feat: allow openId to choose an unactive email if there are none active

rjsparks:


rjsparks:
cc: @rpcross @jennifer-richards

jennifer-richards:
I am not familiar with what is done with the openid claims, but is it a problem to be handing back an email address that someone inactivated? If the other end of the exchange can be changed to handle the empty claim I'd think that's preferable to propagating an address that was deactivated. If not, maybe better to transmit a canned and identifiable "not-an-address"?

Also, why "unactive" rather than "inactive"?

rpcross:
We currently use claims.email from Datatracker as the email address and unique (DT related) identifier of the registration, used when calling DT APIs. An empty or generic string would break this.

jennifer-richards (approved):
Assuming you'll change the "u" to "i" in the names and approving.

PR #6069, opened 2023-08-01
fix: Set colspan to actual number of columns

jennifer-richards:
Fixes failing template validation in https://github.com/ietf-tools/datatracker/actions/runs/5702257684/job/15487979369


jennifer-richards:
n.b., there is still a failing template validation related to `view_feedback_pending.html`. That has been fixed in #6002 so not re-fixing it here. May be worth cherry picking that to get tests passing on main quickly.

NGPixel (approved):

PR #6105, opened 2023-08-09
fix: make some scripts to be Python 3 compatible

VoltrexKeyva:
Since `print` is a keyword in Python 2, these Python scripts use it without parentheses which works but this does not work in Python 3 since `print` is now a function, so use parentheses to make the scripts compatible with Python 3 without dropping Python 2 support.

### DIFF
--- a/bin/add-old-drafts-from-archive.py
+++ b/bin/add-old-drafts-from-archive.py
@@ -27,7 +27,7 @@ system = Person.objects.get(name="(System)")
 expired = State.objects.get(type='draft',slug='expired')
 
 names = set()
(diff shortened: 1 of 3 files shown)


rjsparks:
Thank you for this contribution. 

Unfortunately the right thing to do with these files is mostly to delete them. They are queued for a coming code-cleanup phase where any relevant part of what they used to do might be extracted.

The project has been Python3 (and not python2.7 compatible) for some time.

Apologies that the cruft attracted your attention.

VoltrexKeyva:
I see, I suppose I can close this to let the said code-cleanup extract the important functionality.

PR #6125, opened 2023-08-10
feat: Create rfc documents from rfc index updates

jennifer-richards:


ietf/sync/rfceditor.py
```diff
@@ -367,80 +370,140 @@ def update_docs_from_rfc_index(index_data, errata_data, skip_older_than_date=Non
...
+            alias, _ = DocAlias.objects.get_or_create(name=doc.name)
             alias.docs.add(doc)
-            changes.append("created alias %s" % prettify_std_name(name))
+            rfc_changes.append(f"created alias {prettify_std_name(doc.name)}")
+
+        if draft_name:
+            try:
+                draft = Document.objects.get(name=draft_name)
+            except Document.DoesNotExist:
+                pass
```
rjsparks, line 425: We should probably log this - the RPC thinks the doc came from a draft, but we can't find it. That's probably a real error.

ietf/sync/rfceditor.py
```diff
@@ -367,80 +370,140 @@ def update_docs_from_rfc_index(index_data, errata_data, skip_older_than_date=Non
...
+                            draft_changes.append(
+                                f"changed {new_state.type.label} to {new_state}"
+                            )
+                            e = update_action_holders(draft, prev_state, new_state)
+                            if e:
+                                draft_events.append(e)
+                    elif t == "draft-iesg":
+                        draft.set_state(
+                            State.objects.get(type_id="draft-iesg", slug="idexists")
+                        )
```
rjsparks, line 475: This doesn't make sense to me.
Instead, verify that the draft stream matches the rfc stream, and then verify the states only for that stream.

ietf/sync/rfceditor.py
```diff
@@ -367,80 +370,140 @@ def update_docs_from_rfc_index(index_data, errata_data, skip_older_than_date=Non
...
 
         # check attributes
         if title != doc.title:
             doc.title = title
-            changes.append("changed title to '%s'" % doc.title)
+            rfc_changes.append(f"changed title to '{doc.title}'")
```
rjsparks, line 492: i'm a little itchy at "changed" here. I wonder if "set" would be better? Since doc is the new RFC document, you're not checking to see if the RFC had a different title than the draft did. Perhaps the gearing should change to make and report on _that_ instead? Or am I misreading?

ietf/sync/rfceditor.py
rjsparks: I see below that this is what the old code did, but I think that was wrong.

ietf/sync/rfceditor.py
```diff
@@ -467,35 +530,24 @@ def update_docs_from_rfc_index(index_data, errata_data, skip_older_than_date=Non
...
                 # if the STD had more than one RFC the wrong thing happens
                 #
-                #if x[:3] in ("NIC", "IEN", "STD", "RTR"):
+                # if x[:3] in ("NIC", "IEN", "STD", "RTR"):
                 #    # try translating this to RFCs that we can handle
                 #    # sensibly; otherwise we'll have to ignore them
                 #   l = DocAlias.objects.filter(name__startswith="rfc", docs__docalias__name=x.lower())
-                #else:
+                # else:
                 l = Document.objects.filter(name=x.lower())
```
rjsparks, line 564: I wonder if we should be filtering this to doc type rfc?

ietf/sync/rfceditor.py
```diff
@@ -504,56 +556,96 @@ def parse_relation_list(l):
...
             for a in also:
                 a = a.lower()
                 if not DocAlias.objects.filter(name=a):
                     DocAlias.objects.create(name=a).docs.add(doc)
-                    changes.append("created alias %s" % prettify_std_name(a))
+                    rfc_changes.append(f"created alias {prettify_std_name(a)}")
 
-        doc_errata = errata.get('RFC%04d'%rfc_number, [])
-        all_rejected = doc_errata and all( er['errata_status_code']=='Rejected' for er in doc_errata )
+        doc_errata = errata.get("RFC%04d" % rfc_number, [])  # rfc10k problem here
```
rjsparks, line 608: Lets get rid of the %04 here now?

rjsparks: do errata searches actually take leading zeros?

ietf/sync/rfceditor.py
```diff
@@ -504,56 +556,96 @@ def parse_relation_list(l):
...
+                DocEvent.objects.create(
+                    doc=doc,
+                    rev=doc.rev,
+                    by=system,
+                    type="sync_from_rfc_editor",
+                    desc=f"Received changes through RFC Editor sync ({', '.join(rfc_changes)})",
+                )
+            )
+            doc.save_with_history(rfc_events)
+            yield rfc_changes, doc, rfc_published  # yield changes to the RFC
```
rjsparks: will this yielding do the right thing if there are both draft_changes and rfc_changes?

ietf/sync/rfceditor.py
jennifer-richards: I can make this handle leading zeroes or not - that'll avoid having to closely coordinate the change.

ietf/sync/rfceditor.py
jennifer-richards: Yes:

jennifer-richards: (this is also confirmed *in situ* by `test_rfc_index()`)

jennifer-richards: (rather, will be when I commit that piece)

ietf/sync/rfceditor.py
jennifer-richards: I'd say so

ietf/sync/rfceditor.py
jennifer-richards: will do

ietf/sync/rfceditor.py
jennifer-richards: You've drawn my attention to a bigger bug. These fields should be populated from the draft to start with. With that done, I *think* the "change" language will continue to make sense.

jennifer-richards: (though I can make it choose "set" vs "change" based on the value of `created_rfc` if you prefer that)

jennifer-richards: (continuing to talk to myself here) Never mind - we aren't guaranteed to have a draft to copy the fields from. I'll have it report "set" vs "changed" as appropriate.

jennifer-richards:
I believe I addressed all your comments, plus some other changes. I suggest looking closely at 7535df3eeb67dddcb9756f17761828e7d98c27bf which brings more fields over from the draft to the new rfc document (these are also the fields set on the rfc document in the `...create_rfc_documents.py` migration). The refactored stream state handling is also worth checking in detail.

rjsparks (approved):

PR #6130, opened 2023-08-11
feat: add the IAB to the group menu

### LINKED ISSUE #6120: Make it possible to navigate to the IAB group itself via the menus
Right now Group->IAB lets you get to programs and administrative activities, but not the IAB itself.

rjsparks:
Fixes #6120

This is a _kludge_, but the IAB is unique in the set of groups we might want to navigate to. We could consider doing something similar for the IESG, especially after moving IESG statements and appeals to that page.
image


jennifer-richards (approved):
Eww, but well motivated.

PR #6168, opened 2023-08-18
fix: construct double-submit test correctly

rjsparks:


ietf/doc/tests_review.py
```diff
@@ -930,10 +931,9 @@ def test_double_submit_review(self):
         })
         self.assertEqual(r.status_code, 302)
         r2 = self.client.get(r.url)
-        # FIXME-LARS: this fails when the tests are run with --debug-mode, i.e., DEBUG is set:
-        if not settings.DEBUG:
-            self.assertEqual(len(r2.context['messages']),1)
-            self.assertIn('Attempt to save review failed', list(r2.context['messages'])[0].message)
+        self.assertEqual(r2.status_code, 200)
```
rjsparks, line 934: This get had been 404-ing because the original Document.create didn't build a DocAlias (and didn't set a title, which led to html validation failure once the alias was in place). Using the factory clears this all up.

--debug-mode was failing because 404s are overriden there and the message framework message wouldn't be displayed.

jennifer-richards (approved):

PR #6193, opened 2023-08-23
fix: Reword I-D announcement emails to make first sentence more useful

### LINKED ISSUE #6092: In the mail template to i-d-announce@ietf.org could the Draft title come first in the body please?
### Description
In the templated mail from the IETF to i-d-announce@ietf.org every post starts with the sentence: "A New Internet-Draft is available from the on-line Internet-Drafts directories."

It's proper english, sure. It's also redundant. The list is exclusively new drafts.

If the message started with the title/authors block then line view would include the draft 

image

shows how this looks in gmail. I can almost see the Title: but not quite.

I appreciate this is a UI problem, and I understand there is risk that some people expect and need to see the line, so I ask this knowing that not everyone will be satisfied with this change. 

### Code of Conduct

larseggert:
Fixes #6092

Also makes some other tweaks to the layout of related emails.

### DIFF
--- a/ietf/templates/submit/announce_to_authors.txt
+++ b/ietf/templates/submit/announce_to_authors.txt
@@ -1,24 +1,23 @@
-{% autoescape off %}
-A new version of I-D, {{ submission.name }}-{{ submission.rev }}.txt
-has been successfully submitted by {{ submission.submitter_parsed.name }} and posted to the
-IETF repository.
+{% autoescape off %}{% filter wordwrap:78 %}A new version of Internet-Draft {{ submission.name }}-{{ submission.rev }}.txt has been successfully submitted by {{ submission.submitter_parsed.name }} and posted to the
+IETF repository.{% endfilter %}
 
-Name:		{{ submission.name }}
-Revision:	{{ submission.rev }}
-Title:		{{ submission.title }}
-Document date:	{{ submission.document_date|date:"Y-m-d" }}
-Group:		{{ group }}
-Pages:		{{ submission.pages }}
-URL:            {{ settings.IETF_ID_ARCHIVE_URL }}{{ submission.name }}-{{ submission.rev }}.txt
-Status:         {{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.doc.views_doc.document_main' name=submission.name %}{% if submission.xml_version == "3" %}
-Html:           {{ settings.IETF_ID_ARCHIVE_URL }}{{ submission.name }}-{{ submission.rev }}.html{% endif %}
-Htmlized:       {{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.doc.views_doc.document_html' name=submission.name %}
-{% if submission.rev != "00" %}Diff:           {{rfcdiff_base_url}}?url2={{ submission.name }}-{{ submission.rev }}{% endif %}
+Name:     {{ submission.name }}
+Revision: {{ submission.rev }}
+Title:    {{ submission.title }}
+Date:     {{ submission.document_date|date:"Y-m-d" }}
+Group:    {{ group }}
+Pages:    {{ submission.pages }}
+URL:      {{ settings.IETF_ID_ARCHIVE_URL }}{{ submission.name }}-{{ submission.rev }}.txt
+Status:   {{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.doc.views_doc.document_main' name=submission.name %}{% if submission.xml_version == "3" %}
+HTML:     {{ settings.IETF_ID_ARCHIVE_URL }}{{ submission.name }}-{{ submission.rev }}.html{% endif %}
+HTMLized: {{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.doc.views_doc.document_html' name=submission.name %}
+{% if submission.rev != "00" %}Diff:     {{rfcdiff_base_url}}?url2={{ submission.name }}-{{ submission.rev }}{% endif %}
 
 Abstract:
-{{ submission.abstract }}
-                                                                                  
-{{ submission.note|default:"" }}
+
+{{ submission.abstract.rstrip }}
+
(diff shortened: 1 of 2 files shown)


ietf/templates/submit/announce_to_authors.txt
```diff
@@ -1,24 +1,23 @@
...
+Name:     {{ submission.name }}
+Revision: {{ submission.rev }}
+Title:    {{ submission.title }}
+Date:     {{ submission.document_date|date:"Y-m-d" }}
+Group:    {{ group }}
+Pages:    {{ submission.pages }}
+URL:      {{ settings.IETF_ID_ARCHIVE_URL }}{{ submission.name }}-{{ submission.rev }}.txt
+Status:   {{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.doc.views_doc.document_main' name=submission.name %}{% if submission.xml_version == "3" %}
+HTML:     {{ settings.IETF_ID_ARCHIVE_URL }}{{ submission.name }}-{{ submission.rev }}.html{% endif %}
+HTMLized: {{ settings.IDTRACKER_BASE_URL }}{% url 'ietf.doc.views_doc.document_html' name=submission.name %}
```
richsalz, line 13: What is sthe difference between `HTML` and `HTMLized`? Can we find a better word to make that more obvious?

rjsparks: For modern documents (anything submitted in xmlrfc), HTML is the html rendering from xml2rfc. HTMLized is the representation of the text rendering as html with some heursitic links created. We can HTMLize documents that are submitted only as text. We cannot make v3 HTML them.

What the datatracker /doc/html page shows _looks_ (mostly) like the htmlized version, whether it eats .txt and htmlizes it or uses css to format the v3 HTML to look like the HTMLized.

And yes, this is not easy to explain.

ietf/templates/submit/announce_to_lists.txt
```diff
@@ -1,26 +1,27 @@
...
+{% autoescape off %}{% filter wordwrap:78 %}Internet-Draft {{ submission.name }}-{{ submission.rev }}.txt is now available.{% if submission.group %} It is a work item of the {{ submission.group.name }} ({{ submission.group.acronym|upper }}){% if submission.group.type.name %} {{ submission.group.type.name }}{% endif %} of the {% if submission.group.type_id == "rg" %}IRTF{% else %}IETF{% endif %}.{% endif %}{% endfilter %}
 
-   Title           : {{ submission.title }}
-   Author{{ submission.authors|pluralize:" ,s" }}         : {% for author in submission.authors %}{{ author.name }}{% if not forloop.last %}
-                     {% endif %}{% endfor %}
-   Filename        : {{ submission.name }}-{{ submission.rev }}.txt
-   Pages           : {{ submission.pages }}
-   Date            : {{ submission.submission_date|date:"Y-m-d" }}
+   Title:   {{ submission.title }}
+   Author{{ submission.authors|pluralize:",s" }}: {% if submission.authors|length == 1 %} {% endif %}{% for author in submission.authors %}{{ author.name }}{% if not forloop.last %}
```
rjsparks, line 4: Why the extra space here when there is only one author?

ietf/templates/submit/announce_to_authors.txt
richsalz: Is it time to just drop the HTMLized?
Can we add it only if txt was submitted?

I guess the bigger question is when would we want to provide both?

ietf/templates/submit/announce_to_authors.txt
```diff
@@ -1,24 +1,23 @@
...
+{% if submission.rev != "00" %}Diff:     {{rfcdiff_base_url}}?url2={{ submission.name }}-{{ submission.rev }}{% endif %}
 
 Abstract:
-{{ submission.abstract }}
-                                                                                  
-{{ submission.note|default:"" }}
+
+{{ submission.abstract.rstrip }}
```
rjsparks, line 18: Did you consider a full strip instead of just an rstrip here (and below)?

rjsparks (approved):

ietf/templates/submit/announce_to_authors.txt
larseggert: I think we should do a pass over all email announcements and make sure we link to the metadata page in the datatracker in most cases, rather than including per-rendering links (or using names baking in renderings like `.txt`).

ietf/templates/submit/announce_to_lists.txt
larseggert: Because with multiple authors the string before the colon is `Authors`, i.e., one character longer.

ietf/templates/submit/announce_to_authors.txt
larseggert: Last time I touched this I added the `rstrip`, and there was some reason for it (and not `strip), but I can't recall.

ietf/templates/submit/announce_to_lists.txt
rjsparks: Consider the discussion of pluralizing cherry at https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#pluralize

ietf/templates/submit/announce_to_authors.txt
rjsparks: > Is it time to just drop the HTMLized

No. Your question isn't really just about "in the announcement of new Internet-Draft versions" its a thing that we have to deal with in general, including the pre-v3 RFCs. The /doc/html pages _use_ that format so that we can provide the same kind of view for older and newer documents. Unless we somehow convinced the community that the v3 html rendering's CSS change to be what is produced at /doc/html, we need two terms for the two ways of looking at things.

This conversation should move out of the ticket - probably onto tools-discuss@

rjsparks:
I'm merging this - we can improve the pluralization in the template in the next release.

Issue #6057, opened 2023-07-27
Finding author by email when parsing drafts at submission may not ignore case


rjsparks:
Investigate whether an author that puts "MyName@SomeCompany.foo" will match a Person with an Email.address of "myname@somecompany.foo".

rjsparks:
likely a duplicate of #5929

pselkirk:
Closed as duplicate of #5929

Issue #5929, opened 2023-07-06
Submission checks don't compare submitter/author email addresses case-insensitively


rjsparks:
There is at least one participant that is not able to use the submission tool because Email records have CamelCase@Address.Tld and the submitted draft has camelcase@address.tld.

pselkirk:
After extensive review, I've determined that author case sensitivity is not a problem with either the interactive submission tool or the API. The tool does a `Person` comparison, and the API does an explicitly case-insensitive string comparison.

rjsparks:
(for the benefit of historians): The original issue was actually a MySQL->Postgresql migration error where we left two Email objects with addresses that varied only by case. That combined with making Email.address a CIText field led to difficult to diagnose behavior. We corrected that (verifying that there were no other instances) and the reported issue for the submitter was resolved.

PR #6194, opened 2023-08-23
test: Add tests for case sensitivity

pselkirk:
This also includes changes to existing tests to deal with case-sensitivity issues that were revealed by temporarily hard-wiring PersonFactory to use an upper-case or mixed-case email address.

This testing was done while researching  #5929 and #6057. This testing didn't result in code changes outside of the test code, so it's not imperative that it be merged.

### DIFF
--- a/ietf/submit/tests.py
+++ b/ietf/submit/tests.py
@@ -125,7 +125,7 @@ def submission_file_contents(name_in_doc, group, templatename, author=None, emai
     if author is None:
         author = PersonFactory()
     if email is None:
-        email = author.email().address.lower() if author.email() else None
+        email = author.email().address if author.email() else None
     if title is None:
         title = "Test Document"
     if year is None:
@@ -204,7 +204,7 @@ class SubmitTests(BaseSubmitTestCase):
         # Submit views assume there is a "next" IETF to look for cutoff dates against
         MeetingFactory(type_id='ietf', date=date_today()+datetime.timedelta(days=180))
 
-    def create_and_post_submission(self, name, rev, author, group=None, formats=("txt",), base_filename=None, ascii=True):
+    def create_and_post_submission(self, name, rev, author, email=None, group=None, formats=("txt",), base_filename=None, ascii=True):
         """Helper to create and post a submission
 
         If base_filename is None, defaults to 'test_submission'.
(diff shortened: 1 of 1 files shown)


ietf/submit/tests.py
```diff
@@ -541,7 +555,7 @@ def test_submit_new_replaced_wg_as_author(self):
     def test_submit_new_wg_with_extresources(self):
         self.submit_new_draft_with_extresources(group=GroupFactory())
 
-    def submit_existing(self, formats, change_authors=True, group_type='wg', stream_type='ietf'):
+    def submit_existing(self, formats, author=None, email=None, change_authors=True, group_type='wg', stream_type='ietf'):
```
jennifer-richards, line 558: These new parameters don't seem to be used in the method, or am I missing something?

ietf/submit/tests.py
```diff
@@ -3547,6 +3575,26 @@ def test_cancel_stale_submissions(self):
...
+        """process_and_accept_uploaded_submission should be case-insensitive"""
+        xml, author = submission_file('draft-somebody-test-00', 'draft-somebody-test-00.xml', None, 'test_submission.xml')
+        xml_data = xml.read()
+        xml.close()
+
+        submission = SubmissionFactory(
+            name='draft-somebody-test',
+            rev='00',
+            file_types='.xml',
+            submitter=author.formatted_email().upper(),
```
jennifer-richards, line 3588: I'd suggest an assertion that `author.formatted_email().upper() != author.formatted_email()` just to be airtight

jennifer-richards (changes requested):
It looks to me like something is missing - see the comment about `do_submission()`.

There's an inline suggestion about keyword-only arguments. That's more a question about whether you think it'd be appropriate than a demand - we haven't used keyword-only args in the code so far, but I think it's something we should consider.

ietf/submit/tests.py
```diff
@@ -237,7 +237,7 @@ def create_and_post_submission(self, name, rev, author, group=None, formats=("tx
                     self.assertTrue(os.path.exists(os.path.join(self.staging_dir, "%s-%s.%s" % (name, rev, 'html'))))
         return r
 
-    def do_submission(self, name, rev, group=None, formats: Tuple[str, ...]=("txt",), author=None, base_filename=None, ascii=True):
+    def do_submission(self, name, rev, group=None, formats: Tuple[str, ...]=("txt",), author=None, email=None, base_filename=None, ascii=True):
```
jennifer-richards, line 240: I should probably have caught this the first time around, but I don't think this `email` parameter is used either. Worries me that the test is not testing what it means to test.

jennifer-richards (changes requested):
I probably should have caught this on the first pass, but I still don't think `email` is making it all the way down into the test.

rjsparks:
Closing this as OBE. We can come back to it as part of a future cleanup effort.

PR #6197, opened 2023-08-23
feat: import iab agendas from iab website

rjsparks:


ietf/meeting/management/commands/import_iab_agendas.py
```diff
@@ -0,0 +1,240 @@
...
+            e = DocEvent.objects.create(
+                type="comment",
+                doc=doc,
+                rev="00",
+                by_id=1,  # The "(System)" person
+                desc="Agenda moved into datatracker from iab wordpress website",
+            )
+            doc.save_with_history([e])
+
+            session = Session.objects.get(meeting=meeting)
```
jennifer-richards, line 225: If this were more than a one-off, I'd suggest handling possible failure of `get()` here. However, since it's not, I don't.

jennifer-richards (approved):
Looks ok to me. I would be inclined to add a clause checking for unexpected ".md" files in the scraped directory as an accuracy check on the `agendas_to_import()` list, but if you statically checked that it's unnecessary.

PR #6208, opened 2023-08-24
chore: Define RfcToBeFactory

jennifer-richards:


ietf/rpc/factories.py
```diff
@@ -64,6 +68,26 @@ class Meta:
     desc = factory.Faker("sentence")
 
 
+class RfcToBeFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = RfcToBe
+
+    draft = factory.SubFactory(WgDraftFactory)
```
rjsparks, line 75: Better to do this with ('ietf.doc.factories.WgDraftFactory')

ietf/rpc/factories.py
```diff
@@ -64,6 +68,26 @@ class Meta:
...
+    rfc_number = factory.Sequence(lambda n: n + 1000)
+    submitted_format = SourceFormatName.objects.get(slug="xml-v3")
```
rjsparks, line 77: You can avoid importing the classes from Name by using '_id' here (that is, `submitted_format_id="xml-v3"`)
etc.

rjsparks (changes requested):

rjsparks (approved):

PR #6217, opened 2023-08-25
fix: Move submission files using shutil.move()

### LINKED ISSUE #6216: post_submission makes poor assumptions about file systems
### Describe the issue

While working on #6185 and #6186, I tried setting `IDSUBMIT_REPOSITORY_PATH` to a path on a different filesystem from `IDSUBMIT_STAGING_PATH`. This caused a failure when posting a submission:
```
  File "/home/dev/.local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/home/dev/.local/lib/python3.9/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/workspace/ietf/submit/views.py", line 515, in submission_status
    post_submission(request, submission, desc, desc)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/workspace/ietf/submit/utils.py", line 451, in post_submission
    move_files_to_repository(submission)
  File "/workspace/ietf/submit/utils.py", line 654, in move_files_to_repository
    os.rename(source, dest)

Exception Type: OSError at /submit/status/136279/c82dbc3f1b66260b7762926c58beb6a7/
Exception Value: [Errno 18] Invalid cross-device link: 'test/staging/draft-fluhrer-cfrg-ntru-03.txt' -> '/assets/ietf-ftp/internet-drafts/draft-fluhrer-cfrg-ntru-03.txt'
```
The call to `os.rename()` assumes that the source and dest are on the same filesystem. This is unwarranted.

This will certainly need to be fixed (or avoided) when refactoring the artifact storage.

### Code of Conduct

jennifer-richards:
Fixes #6216 

The change to `rename_submission_files()` is not strictly required because both source and dest are in the same directory so cannot be on different filesystems. However, `shutil.move()` uses `os.rename()` in this case so I think it's better to use it consistently.

Also switches some `os.path.join` to `pathlib.Path` - this works for Python 3.9 and later, which I believe is all we're supporting now, but calling it out in case I'm confused.

### DIFF
--- a/ietf/submit/utils.py
+++ b/ietf/submit/utils.py
@@ -641,12 +641,14 @@ def cancel_submission(submission):
     submission.save()
     remove_submission_files(submission)
 
+
 def rename_submission_files(submission, prev_rev, new_rev):
     for ext in settings.IDSUBMIT_FILE_TYPES:
-        source = os.path.join(settings.IDSUBMIT_STAGING_PATH, '%s-%s.%s' % (submission.name, prev_rev, ext))
-        dest = os.path.join(settings.IDSUBMIT_STAGING_PATH, '%s-%s.%s' % (submission.name, new_rev, ext))
-        if os.path.exists(source):
-            os.rename(source, dest)
+        staging_path = Path(settings.IDSUBMIT_STAGING_PATH) 
+        source = staging_path / f"{submission.name}-{prev_rev}.{ext}"
+        dest = staging_path / f"{submission.name}-{new_rev}.{ext}"
+        if source.exists():
+            move(source, dest)
 
 
 def move_files_to_repository(submission):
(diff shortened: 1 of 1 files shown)


rjsparks:
> this works for Python 3.9 and later, which I believe is all we're supporting now

That is correct.

rjsparks (approved):

rjsparks:
Quirk of current deployment to keep in mind:

The archive is built from collections using ghostlinkd, which runs over a set of collections of drafts, one of which is the repository. Ghostlinkd necessarily works on a single filesystem (it watches the filesystem and makes hard links to construct the archive. (See https://github.com/ietf-tools/ghostlinkd)

This will change when we move off of a shared filesystem.

PR #6220, opened 2023-08-25
feat: Distinguish failed from not-applied checkers

### LINKED ISSUE #4346: Submission status display is sometimes misleading about checker results
### Describe the issue

The submission status page shows any passed/failed checker results (i.e., `SubmissionCheck` instances) for the submission. If there are no results or if all the instances are passing results, this is shown as a green box with the statement `Your draft has been verified to pass the submission checks.` If any are not passing, the box is red and the statement indicates that a check failed.

If there are no check results, this is still shown with the message indicating that the draft has been verified to pass the submission checks. This is misleading. It would be better to indicate that the validations have not been run.

### Code of Conduct

jennifer-richards:
Fixes #4346

With this change, passing checks are unchanged, still "Your Internet-Draft has been verified to pass the submission checks."
image
and the wording for a failed test is now "Your Internet-Draft failed at least one submission check."
image
If no checks were applied, it says so, "No submission checks were applied to your Internet-Draft." The visual style is still `alert-warning` because this is not a healthy situation for a submission.
image

This also refactors and reuses the `latest_checks()` method which should give better results if checks are applied multiple times (though afaik this does not happen often if at all).


rjsparks (approved):

rjsparks:
This feels like we're missing some tests.

rjsparks:
@jennifer-richards : please merge main in again

jennifer-richards:
> This feels like we're missing some tests.

I agree. The existing tests of `submission_status()` are mixed in with other tests. I've started a `SubmissionStatusTests` class with a test to exercise the new stuff.

PR #6229, opened 2023-08-28
feat: move IAB appeals into the datatracker

rjsparks:


ietf/group/admin.py
```diff
@@ -327,11 +327,19 @@ def value_from_datadict(self, data, files, name):
             bits = upload.read()
             return b64encode(bits).decode("ascii") # Who made this so hard?
             
+class RestrictContentTypeChoicesForm(forms.ModelForm):
+    content_type = forms.ChoiceField(
+        choices=(
+            ( "text/markdown;charset=utf-8", "Markdown"),
+            ( "application/pdf;charset=utf-8", "PDF")
```
jennifer-richards, line 334: As noted elsewhere with references, I don't think `charset` is valid for `application/pdf`

ietf/group/models.py
```diff
@@ -442,8 +443,13 @@ def display_title(self):
             return self.title
         else:
             return self.artifact_type.name
-        
+
+    def is_markdown(self):
+        return self.content_type == "text/markdown;charset=utf-8"
```
jennifer-richards, line 448: I'd be inclined to be general here and look only at the type (`text/markdown`) excluding the `charset`, but am not going to insist given the minimal likelihood that we are ever going to use that.

ietf/group/tests_appeals.py
```diff
@@ -0,0 +1,64 @@
...
+    AppealArtifactFactory,
+)
+class AppealTests(TestCase):
+
+    def test_appeal_list_view(self):
+        appeal_date = datetime.date.today()-datetime.timedelta(days=14)
+        response_date = appeal_date+datetime.timedelta(days=8)
+        appeal = AppealFactory(name="A name to look for", date=appeal_date)
+        appeal_artifact = AppealArtifactFactory(appeal=appeal, artifact_type__slug="appeal", date=appeal_date)
+        response_artifact = AppealArtifactFactory(appeal=appeal, artifact_type__slug="response", content_type="application/pdf;charset=utf-8", date=response_date)
```
jennifer-richards, line 22: `charset` on `application/pdf`

ietf/group/tests_appeals.py
```diff
@@ -0,0 +1,64 @@
...
+        url = urlreverse("ietf.group.views.appeals", kwargs=dict(acronym="iab"))
+        r = self.client.get(url)
+        self.assertEqual(r.status_code, 200)
+        q = PyQuery(r.content)
+        self.assertEqual(len(q("#appeals > tbody > tr")), 1)
+        self.assertEqual(q("#appeal-1-date").text(), f"{appeal_date}")
+        self.assertEqual(f"{appeal_artifact.display_title()} - {appeal_date}", q("#artifact-1-1").text())
+        self.assertEqual(f"{response_artifact.display_title()} - {response_date}", q("#artifact-1-2").text())
+        self.assertIsNone(q("#artifact-1-1").attr("download"))
+        self.assertEqual(q("#artifact-1-2").attr("download"), response_artifact.download_name())
```
jennifer-richards, line 33: It'd be nice to test somewhere that `download_name()` is behaving - I think all the tests so far would pass if it always returned "DoWnLoAd Me!" or (perhaps more plausibly) "some_prefix_.pdf".

ietf/group/tests_appeals.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def test_markdown_download(self):
+        artifact = AppealArtifactFactory()
+        url = urlreverse("ietf.group.views.appeal_artifact_markdown", kwargs=dict(acronym="iab", artifact_id=artifact.pk))
+        login_testing_unauthorized(self, "secretary", url)
+        r = self.client.get(url)
+        self.assertContains(r, "**Markdown**", status_code=200)
+
+    def test_pdf_download(self):
+        artifact = AppealArtifactFactory(content_type="application/pdf;charset=utf-8") # The bits won't _really_ be pdf
```
jennifer-richards, line 57: `charset` with `application/pdf`

jennifer-richards (changes requested):
Only real concern is whether `charset` is valid as a parameter to the `application/pdf` media type. The RFC and iana database suggest not. I've flagged it where I noticed it (and linked to the iana info page at the first place I noticed this.)

Otherwise, I have a nit-picky suggestion that I have no qualms about your ignoring, and a suggstion for an additional thing to test.

ietf/group/models.py
rjsparks: I want to leave it this brittle edge for now - it will cause anything that is encoded differently (or that has a different content-type provided) to _not render_ as html, and I think that's what we'd prefer to have happen.

jennifer-richards (approved):

Issue #6088, opened 2023-08-07
Capture that IPR disclosures are removed under the "Objectively False IPR Disclosure Policy"


rjsparks:
The datatracker needs an new IPR state reflecting removal under the Objectively False IPR Disclosure Policy, and the views/templates that display IPR disclosure status need to be updated to reflect the new state. 
```[tasklist]
### Tasks
- [ ] Use a migration to add a new `removed_objfalse` ietf.name.IprDisclosureStateName
- [ ] Create a PUBLISH_IPR_STATES setting of ["posted", "removed", "removed_objfalse"]. Rework all places that currently explicitly filter to ["posted, "removed"] to use the new setting
- [ ] Update the mapping of IPR state to DocEvent state and provide a distinct description for the new docevent state type (which is not an ietf.name, so no migration needed).
- [ ] Migrate the few disclosures that have already been removed under this policy to the new state. The DocEvent objects coupled to the 'removed' state will need to be migrated to use the type and description for the new DocEvent type coupled to `removed_objfalse`
```

rjsparks:
This turns out to be moderately complex.

It's tempting to rename the existing `removed` slug, but there may be API users that would have to change, so we'll leave it alone.

Relevant places to focus:

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/ipr/models.py#L227-L248

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/ipr/views.py#L270-L291

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/ipr/views.py#L777-L800

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/templates/ipr/details_view.html#L101-L110

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/templates/ipr/details_view.html#L120-L133

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/templates/ipr/search_doc_result.html#L27-L35

https://github.com/ietf-tools/datatracker/blob/5522de976c461e4d3b6862ea21dc2cf661a90863/ietf/templates/ipr/search_doc_result.html#L66-L89

[...]

PR #6231, opened 2023-08-28
feat: Capture that IPR disclosures are removed under the Objectively False IPR Disclosure Policy

pselkirk:
Implements #6088


ietf/settings.py
```diff
@@ -1301,3 +1301,5 @@ def skip_unreadable_post(record):
     CSRF_TRUSTED_ORIGINS += ['http://localhost:8000', 'http://127.0.0.1:8000', 'http://[::1]:8000']
     SESSION_COOKIE_SECURE = False
     SESSION_COOKIE_SAMESITE = 'Lax'
+
+PUBLISH_IPR_STATES = ['posted', 'removed', 'removed_objfalse']
```
rjsparks, line 1305: Minor tweak needed: Please move this above the `if not production` block. Just before is fine (unless you see some other puddle of ipr app related settings to attach it to).

rjsparks (changes requested):

jennifer-richards (approved):

rjsparks (approved):

PR #6248, opened 2023-08-30
chore: merge main and fix conflicts

rjsparks:


ietf/templates/doc/document_info.html
```diff
@@ -239,17 +239,7 @@
...
-                             class="w-25 mt-1"
-                             {% if doc.stream.name|lower == 'iab' %}
-                                 src="{% static 'ietf/images/iab-logo.svg' %}"
-                             {% elif doc.stream.name|lower == 'ietf' %}
-                                 src="{% static 'ietf/images/ietf-logo.svg' %}"
-                             {% elif doc.stream.name|lower == 'irtf' %}
-                                 src="{% static 'ietf/images/irtf-logo.svg' %}"
-                             {% endif %}
-                             >
+                        {% include "logo.html" with org=doc.stream.name|lower classes="w-25 mt-1" title=stream_desc nor=True only %}
```
jennifer-richards, line 242: Should `stream_desc` be changed back to `doc.stream.name` here? It's set over in `views_doc.py`, along with `stream`. I didn't see where / whether `stream` is used, though, but haven't searched everywhere.

jennifer-richards (commented):
One note where a change on `feat/rfc` was reverted - but it's not clear to me whether the reversion is good or bad.

ietf/templates/doc/document_info.html
rjsparks: Probably best as is. `doc.stream` could be None. and title shouldn't be the "not found" thing. It might be possible to make this cleaner, but that's probably a main problem.

I might have created confusion when merging - this was a conflict line.

jennifer-richards (approved):

Issue #6051, opened 2023-07-26
Edit state button leads to 404.


rjsparks:
From Murray Kucherawy:

>On an expired document (current example: https://datatracker.ietf.org/doc/draft-freed-smtp-limits/), the IESG state is showing “Dead”, which is correct. Next to that there’s an “Edit” button, which when pressed takes me to a 404 page.

>I imagine either the button should be removed or greyed out (there’s nowhere to go from “Dead” unless a new version is posted), but other solutions are possible.

rjsparks:
cc: @mskucherawy

rjsparks:
Again from Murray (about draft-ietf-ecrit-lost-planned-changes):

>This document was, I believe, in state "AD is watching" when it expired, which caused it to move to "Dead".  They've now posted a new version, but I'm not able to restore it to a state where I can do anything to it. 

The secretariat, at least, should be able to edit the IESG state away from Dead. Lets revisit why there was a 404 and if it's because ADs don't have permission, then change the template to not give ADs the button, but leave it for the Secretariat.

I'll work with Murray on what to do with this particular draft in the meantime.

Separately, we need to revisit the transition from "AD is watching" to "Dead" - it should probably be to "ID Exists".

rjsparks:
The original issue was raised while the document was expired, and editing iesg state was not available then because of this line:
https://github.com/ietf-tools/datatracker/blob/aa955f0e9b7d7c98b6c467d2cf42e65b2c0f68fc/ietf/doc/views_draft.py#L91

rjsparks:
cc: @mskucherawy

rjsparks:
See @5560.

What we'll do for _this_ issue in the meantime is make sure that the secretariat is always able to change the IESG state.

PR #6253, opened 2023-08-31
fix: Allow secretariat to change the state of a dead draft

pselkirk:
Fixes re-opened #6051


ietf/templates/doc/draft/change_state.html
```diff
@@ -10,6 +10,13 @@ <h1>
         <br>
         <small class="text-body-secondary">{{ doc }}</small>
     </h1>
+    {% if state.slug == "dead" %}
+        <p class="alert alert-warning my-3">
+           This document is in state "Dead". It is unusual to change this
```
rjsparks, line 15: IESG state "Dead"

ietf/doc/views_doc.py
```diff
@@ -261,7 +261,9 @@ def document_main(request, name, rev=None, document_html=False):
         iesg_state_summary = doc.friendly_state()
         irsg_state = doc.get_state("draft-stream-irtf")
 
-        can_edit = has_role(request.user, ("Area Director", "Secretariat"))
+        can_edit = has_role(request.user,
+                            ("Secretariat") if iesg_state_slug == 'dead' else
+                            ("Area Director", "Secretariat"))
```
rjsparks, line 266: The scope of this change is too big. We would need to review each of the many places this would change access for the IESG separately and get the IESG to buy in.

The switch should remain in the template, and should only affect editing the IESG state.

rjsparks (changes requested):

rjsparks (approved):

PR #6260, opened 2023-08-31
chore: make some old things kept for documentation a little safer

rjsparks:


NGPixel (approved):

jennifer-richards (approved):
Feels a little silly to ask for this change, but the `exit(0)` in `recalculate-rfc-authors-snapshot` should be `exit` or `exit 0` because of the change to `/bin/sh`.

Given the intention of the code, though, I'm approving it as it is. It's up to you how much effort you want to expend in "correctly" bailing out of code not meant to be executed. :-)

PR #6272, opened 2023-09-03
feat: Give AD the action in ad-f-up doc state

### LINKED ISSUE #6266: Review action holder manipulation on draft-revision submission.
From @cabo at https://mailarchive.ietf.org/arch/msg/tools-discuss/kRL2lI4KASkF040nPkpdP6I_Gu8:

> https://datatracker.ietf.org/doc/draft-ietf-lake-edhoc/ is in state Publication Requested::AD Followup
> 
> You would expect that this makes the AD action holder?
> 
> No.  When a new version was posted, the state changed, and the action holder was removed:
> 
> "Removed all action holders (IESG state changed)”
> 
> So the AD needs to do the followup now but isn’t an action holder?
> 
> (I had the same problem earlier with one of my drafts, so this is not a fluke.)

jennifer-richards:
Submitting as a draft PR because I have not updated any tests, but would like feedback on the basic premise here. 

Python-wise, I'm using a dataclass, a Python feature we have not used much if at all yet. It has been around since 3.7 and cuts down on boilerplate for a use like this.

In terms of the functional change, this adds the AD to the action holders when the `ad-f-up` state is added. It does not remove them when that tag is removed. It could if this is going to be needed, but will take some care because it will interact more strongly with action holder updates on other state changes. I.e., it may not be as easy to consider the substate tags separately from the states when deciding who the next action holder are.

Fixes #6266


ietf/doc/utils.py
```diff
@@ -491,16 +509,19 @@ def update_action_holders(doc, prev_state=None, new_state=None, prev_tags=None,
...
-        elif ('need-rev' not in prev_tag_slugs) and ('need-rev' in new_tag_slugs):
-            # Added the 'need-rev' tag - add authors to the action holders list
-            for auth in doc.authors():
-                if not doc.action_holders.filter(pk=auth.pk).exists():
-                    doc.action_holders.add(auth)
+    # If we have added or removed the need-rev tag, add or remove authors as action holders
+    if tags.removed("need-rev"):
+        # Removed the 'need-rev' tag - drop authors from the action holders list
+        DocumentActionHolder.objects.filter(document=doc, person__in=doc.authors()).delete()
+    elif tags.added("need-rev"):
```
rjsparks, line 516: Only a few seconds thought given to it, but I'm not seeing a case where this tag being added shouldn't relieve the AD of being an action holder. (In case that's too many nested nots, I think the ad should be removed as an action holder when this tag is added.

@larseggert - am I missing any use cases that say the AD should remain as an action holder when this tag is added?

rjsparks:
> Python-wise, I'm using a dataclass, a Python feature we have not used much if at all yet. It has been around since 3.7 and cuts down on boilerplate for a use like this.

No objection.

ietf/doc/utils.py
jennifer-richards: Sounds reasonable to me - I'll add

larseggert: When _which_ tag is added?

jennifer-richards: the "need-rev" (identified as the "Revised I-D needed" in the UI)

larseggert: When that is added, IMO the AD should never remain an action holder.

ietf/doc/utils.py
```diff
@@ -473,34 +489,45 @@ def update_action_holders(doc, prev_state=None, new_state=None, prev_tags=None,
...
+        # Remove the AD if we're asking for a new revision
+        DocumentActionHolder.objects.filter(document=doc, person=doc.ad).delete()
+
+    # Add new action holders
+    if doc.ad:
+        # AD is an action holder unless specified otherwise for the new state
+        if iesg_state_changed and new_state.slug not in DocumentActionHolder.CLEAR_ACTION_HOLDERS_STATES:
```
rjsparks, line 522: I think this needs a test around it proving that #6266 has been addressed?

rjsparks (changes requested):

ietf/doc/utils.py
jennifer-richards: You're right, thanks for catching that. I've added a test.

rjsparks (approved):

PR #6283, opened 2023-09-05
fix: use type instead of state in templates

rjsparks:


ietf/templates/doc/document_rfc.html
```diff
@@ -13,7 +13,7 @@
           title="Document changes"
           href="/feed/document-changes/{{ name }}/">
     <meta name="description"
-          content="{{ doc.title }} {% if doc.get_state_slug == 'rfc' and not snapshot %}(RFC {{ rfc_number }}{% if published %}, {{ doc.pub_date|date:'F Y' }}{% endif %}{% if obsoleted_by %}; obsoleted by {% for rel in obsoleted_by %}{{ rel.source.name|prettystdname}}{% if not forloop.last%}, {% endif %}{% endfor %}{% endif %}){% endif %}">
+          content="{{ doc.title }} {% if doc.type_id == 'rfc' and not snapshot %}(RFC {{ rfc_number }}{% if published %}, {{ doc.pub_date|date:'F Y' }}{% endif %}{% if obsoleted_by %}; obsoleted by {% for rel in obsoleted_by %}{{ rel.source.name|prettystdname}}{% if not forloop.last%}, {% endif %}{% endfor %}{% endif %}){% endif %}">
```
jennifer-richards, line 16: Do we still need the `not snapshot` check here and elsewhere? (Should be harmless at worst, so I don't mind leaving it for now.)

jennifer-richards (approved):
Looks good - one question that's not really a change request but might be future cleanup

ietf/templates/doc/document_rfc.html
rjsparks: My next PR (work already in progress) will address that.

PR #6285, opened 2023-09-07
fix: rationalize document_rfc.html and the view driving it

rjsparks:


ietf/doc/views_doc.py
```diff
@@ -315,18 +305,12 @@ def document_main(request, name, rev=None, document_html=False):
...
-                    js = Path(finders.find("ietf/js/document_html.js")).read_text()
-                    css = Path(finders.find("ietf/css/document_html_inline.css")).read_text()
-                    if html:
-                        css += Path(finders.find("ietf/css/document_html_txt.css")).read_text()
+            html = doc.html_body()
+            if request.COOKIES.get("pagedeps") == "inline":
+                js = Path(finders.find("ietf/js/document_html.js")).read_text()
+                css = Path(finders.find("ietf/css/document_html_inline.css")).read_text()
+                if html:
+                    css += Path(finders.find("ietf/css/document_html_txt.css")).read_text()
```
larseggert, line 313: I wonder if these should be read into constants instead, to avoid the per-request overheads?

jennifer-richards: I think we really need to avoid inlining them, especially the JS

jennifer-richards (approved):

PR #6292, opened 2023-09-11
fix: Reject nul characters in ipr search parameters

jennifer-richards:
Really ought to rework this as a form, but in the meantime this should prevent 500s. Could probably reduce the number of places we check the value.

### DIFF
--- a/ietf/ipr/views.py
+++ b/ietf/ipr/views.py
@@ -10,7 +10,7 @@ from django.contrib import messages
 from django.db.models import Q
 from django.forms.models import inlineformset_factory, model_to_dict
 from django.forms.formsets import formset_factory
-from django.http import HttpResponse, Http404, HttpResponseRedirect, HttpResponseNotAllowed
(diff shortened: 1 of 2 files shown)


rjsparks (approved):

ietf/ipr/views.py
```diff
@@ -664,13 +675,12 @@ def search(request):
...
-                        q = normalize_draftname(q)
-                        start = DocAlias.objects.filter(name__icontains=q, name__startswith="draft")
-                    elif search_type == "rfc":
-                        start = DocAlias.objects.filter(name="rfc%s" % q.lstrip("0"))
-                
+                elif search_type == "draft":
+                    q = normalize_draftname(q)
+                    start = DocAlias.objects.filter(name__icontains=q, name__startswith="draft")
+                else:  # search_type == "rfc"
+                    start = DocAlias.objects.filter(name="rfc%s" % q.lstrip("0"))
```
rjsparks, line 682: When we replace this, if django is still directly in the path of the request, this spot could use better validation.

rjsparks (approved):

PR #6298, opened 2023-09-12
fix: migration, main view, htmlized view repair

rjsparks:


ietf/templates/doc/document_history_form.html
```diff
@@ -11,34 +11,22 @@
...
-                                {% if diff_revisions|length > 1 and forloop.counter == 2 %} selected{% endif %}
-                            {% endif %}>
+                {% for name, rev, time, url, is_this_doc, is_previous_doc in diff_revisions %}
+                    <option value="{{ url }}"{% if is_previous_doc %} selected{% endif %}>
                         {{ name|prettystdname }}{% if rev %}-{{ rev }}{% endif %}
                         {% if not document_html %}({{ time|date:"Y-m-d" }}){% endif %}
                     </option>
                 {% endfor %}
-                {% if doc.name|rfcbis %}
+                {% if doc.name|rfcbis %} {# TODO - move this into the diffable versions logic #}
```
rjsparks, line 27: Intentionally leaving this TODO for later

ietf/doc/views_doc.py
```diff
@@ -1153,6 +1126,9 @@ def document_email(request,name):
 
 
 def get_diff_revisions(request, name, doc):
+    """ returns list of (name, rev, time, url, is_this_doc, is_previous_doc)
```
jennifer-richards, line 1129: Would be nice to follow the [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings) convention for docstrings as much as we can

ietf/doc/views_doc.py
```diff
@@ -1179,16 +1155,17 @@ def get_diff_revisions(request, name, doc):
...
         )
     )
-    diff_documents.extend(
-        Document.objects.filter(
-            relateddocument__target=doc,
-            relateddocument__relationship="became_rfc"
-        )
-    )
+    if doc.came_from_draft():
+        diff_documents.append(doc.came_from_draft())
```
jennifer-richards, line 1159: I think you might be hitting the DB once for the `if` and again for the `append()` - better to grab the value and reuse it?

Same applies for `became_rfc()` below (and for both it's possible that Django is clever enough to reuse its query results, but I don't think so)

ietf/templates/doc/revisions_list.html
```diff
@@ -6,31 +6,46 @@
...
-        {% elif doc.doc.type_id == 'rfc' %}
-            <li class="page-item rfc">
-                <a class="page-link"
-                   href="{% if document_html %}{% url 'ietf.doc.views_doc.document_html' name=doc.doc.name %}{% else %}{% url 'ietf.doc.views_doc.document_main' name=doc.doc.name %}{% endif %}">
-                    RFC {{ doc.doc.rfc_number }}
-                </a>
-            </li>
+        {% else %}
+            {% for rev in revisions %}
+                {% if rev %} {# I think this guard is now unnecessary #}
```
jennifer-richards, line 31: Remove it and see if anything breaks?

ietf/doc/migrations/0016_move_dochistory.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+    ).update(
+        doc_id=F("rfc_id"), type_id="rfc"
+    )
+    assert not DocHistory.objects.filter(
+        name__startswith="rfc", type_id="draft"
+    ).exists()
+
+
+def reverse(apps, schema_editor):
```
jennifer-richards, line 28: Could just remove the reverse from the `RunPython` call and it'll have this effect

ietf/doc/utils.py
```diff
@@ -936,6 +936,8 @@ def get_predecessors(doc, predecessors=None):
                 if document not in predecessors:
                     predecessors.append(document)
                     predecessors.extend(get_predecessors(document, predecessors))
+        if doc.came_from_draft():
```
jennifer-richards, line 939: See comment in `views_doc` - I think this and the `became_rfc()` call below are making extra DB hits by testing and appending separately

jennifer-richards (changes requested):
A few little things but I'm not opposed to merging as-is if you think they're too trivial at this stage.

ietf/doc/views_doc.py
rjsparks: we could memoize came_from_draft and became_rfc later.

ietf/doc/views_doc.py
```diff
@@ -303,34 +290,23 @@ def document_main(request, name, rev=None, document_html=False):
...
-        additional_urls = doc.documenturl_set.exclude(tag_id='auth48')
-
         html = None
         js = None
         css = None
         diff_revisions = None
         simple_diff_revisions = None
         if document_html:
-            diff_revisions=get_diff_revisions(request, name, doc if isinstance(doc,Document) else doc.doc)
+            diff_revisions=get_diff_revisions(request, name, doc)
```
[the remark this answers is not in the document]
rjsparks: Instead of would mean hitting the database before rejection. (the test would have to be after the get_object_or_404 before we could inspect type_id.)

I don't think "in addition to" would really buy us anything other than an assert like check that we haven't let a document of type RFC slip in that has a name that starts with something other than rfc.

So, how about a logassertion instead?

ietf/templates/doc/revisions_list.html
rjsparks: Leaving it for the moment in case there's something going on outside draft/rfc - I don't think there is, but I need to dig more before taking it away.

jennifer-richards (approved):

PR #6311, opened 2023-09-14
fix: use remastered ietf logos

NGPixel:
Use the same updated logos as the IETF website


larseggert:
Should t we just replace the exiting logos, rather than adding to them?

Also, shouldn't we have the other logos "remastered" as well to keep the look consistent?

rjsparks:
closing this as asleep - it can be reopened if needed.

Issue #6169, opened 2023-08-19
Cannot assign document to someone who has rejected the review


kivinen:
### Describe the issue

I have situation where the David Mandelberg rejected the review because he thought it was assigned to him by mistake. I explained the situation and tried to assign it back to him. The datatracker does not allow me to assign it to him, as he has rejected it before. I tried to remove the request and recreate the request and tried to assign that to him, but that failed too. His name is simply not listed in the list of reviewers I can assign to him.

I know we did ask that we will never automatically assign document back to person who has rejected it once, but the secretary should still be able to assign document to that person.

I had similar issue when I wanted to assign document to area director (at his request), but I could not as he was marked as being unavailable. I think the datatracker should still allow such assignments, but should give warning explaining the situation.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
They are slightly different, but could be combined - the distinction is who the actor is (reviewer or secretary) in overriding the rejection.

Issue #5978, opened 2023-07-17
Allow review rejections to be undone.


rjsparks:
Right now, when a reviewer rejects an assignment (or if a secratary mistakingly rejects and assingment for them), there's no recovery path other than using the admin to change the assignment state.

A reviewer should be able to change their mind and take a review back on (even if there's been another assignment already made in the team).

A secretary should be able to change an assignment from rejected back to requested or even accepted  - if they are doing the wrong thing, social pressure will fix it.

cc: @cindymorgan

PR #6312, opened 2023-09-14
feat:  Allow review rejections to be undone

pselkirk:
Addresses most of #5978 and part of #6169

### DIFF
--- a/ietf/review/policies.py
+++ b/ietf/review/policies.py
@@ -10,6 +10,7 @@ from simple_history.utils import bulk_update_with_history
 from ietf.doc.models import DocumentAuthor, DocAlias
 from ietf.doc.utils import extract_complete_replaces_ancestor_mapping_for_docs
 from ietf.group.models import Role
+from ietf.name.models import ReviewAssignmentStateName
 from ietf.person.models import Person
 import debug                            # pyflakes:ignore
 from ietf.review.models import NextReviewerInTeam, ReviewerSettings, ReviewWish, ReviewRequest, \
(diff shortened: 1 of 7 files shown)


rjsparks:
There's at least one rough edge (and this all needs tests).

I ran through having cellar-flac assigned to me, and then rejecting/accepting/rejecting a few times. In my dev instance I've rejected the assignment there are no other assignements, and the open assignments management page doesn't allow the secretary to assign it to me again. The request, as I understand it, is to have me show up on the list here, but marked "has previously rejected reviewing this doc":

image

(right side truncated to not overshare current states)

rjsparks:
@russhousley @ajeanmahoney @kivinen 

This is what the assignment page will look like for a reviewer that has rejected a review before. This is Russ' view after both of us have been assigned and rejected the document.

Is the UI here intuitive enough? And if you have this as secretaries (if you were logged in, Jean, you would see accept buttons for both of us), would you be ok _not_ having the person show up in the list above?

image

ajeanmahoney:
I would typically work from the list on the **Manage unassigned open review requests** page rather than an individual review request page. As secretary, I would expect to see all available reviewers. 

On the individual review request page, the secretary is reassigning to Russ, yes?

Maybe add a date to "Assignment rejected" under the reviewer name, e.g., "Assignment previously rejected 18 Sept 2023". Maybe there should be a **Reject again** button?

ietf/review/policies.py
```diff
@@ -70,6 +69,12 @@ def assign_reviewer(self, review_req, reviewer, add_skip):
         """Assign a reviewer to a request and update policy state accordingly"""
         # Update policy state first - needed by LRU policy to correctly compute whether assignment was in-order
         self.update_policy_state_for_assignment(review_req, reviewer.person, add_skip)
+        assignment = review_req.reviewassignment_set.filter(reviewer=reviewer).first()
+        if assignment:
+            assignment.state = ReviewAssignmentStateName.objects.get(slug='assigned', used=True)
+            assignment.assigned_on = timezone.now()
+            assignment.save()
+            return assignment
         return review_req.reviewassignment_set.create(state_id='assigned', reviewer=reviewer, assigned_on=timezone.now())
```
rjsparks, line 78: Consider putting this in an `else:` clause

rjsparks (approved):

PR #6315, opened 2023-09-15
fix: Linkify email address lists one-by-one

### LINKED ISSUE #6307: Name rendering fail on email expansion page
See
https://datatracker.ietf.org/doc/draft-tsirtsis-aatn-mech/email/
at
image

larseggert:
Seems to work around a bug in mozilla/bleach.

Fixes #6307

### DIFF
--- a/ietf/templates/doc/document_email.html
+++ b/ietf/templates/doc/document_email.html
@@ -51,11 +51,11 @@
                     </td>
                     <td>
                         {% for addr in to %}
-                            {{addr|linkify}}{% if not forloop.last %}, {% endif %}
+                            {{ addr|linkify }}{% if not forloop.last %}, {% endif %}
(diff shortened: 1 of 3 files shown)


rjsparks (approved):

rjsparks:
A different solution would have been to make linkify also operate on and return lists (or make a new filter named linkify_list), but unless this starts exploding joins in many places, the return isn't that high.

rjsparks:
Test fails are in selenium, which we'll investigate separetely.

larseggert:
> A different solution would have been to make linkify also operate on and return lists (or make a new filter named linkify_list), but unless this starts exploding joins in many places, the return isn't that high.

I tried that, but Django seemed to stringify the list when it gets passed to the filter.

PR #6332, opened 2023-09-18
Change button style

vanillajonathan:
Not all buttons can be primary. Change to secondary, and since there is many of them, use the outline variant.

### DIFF
--- a/ietf/templates/community/list_menu.html
+++ b/ietf/templates/community/list_menu.html
@@ -1,19 +1,19 @@
 <div class="text-end">
     <label id="list-feeds">Atom feed:</label>
     <div class="btn-group" role="group" aria-labelledby="list-feeds">
-        <a class="btn btn-primary"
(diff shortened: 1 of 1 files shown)


lunaroco33 (commented):
srdrtfkjj bñhlihop lkljoiphpiig  mm.hilhñhoblk

rjsparks:
This doesn't look complete (the focus on the community app is odd), and should be supported by an issue that outlines the concern it's really trying to address that we can discuss. Closing this PR rather than suggesting changes at this point..

vanillajonathan:
My concern is that in UI/UX design you expect there to be one primary button, not many. Hence I propose to change `btn-primary` to `btn-secondary`. Since none of them really are primary. Since there are many, I propose using the outline variant `btn-outline-secondary` which is less prominent.

rjsparks:
Understood. We have a professional UI/UX review in the long term plan and this will be addressed application-wide with that.

PR #6345, opened 2023-09-21
fix: Add link to public pages to private NomCom pages

### LINKED ISSUE #6343: NomCom - link to public area from private area
### Description

There are a few things in the public area that are not replicated in the private one.  A link (maybe on the tab bar) would be helpful.

### Code of Conduct

larseggert:
Fixes #6343

Also fix highlighting of the active tab in the nav tab bar while I'm here, and remove old links that 404. Oh, and force-break some long/bogus UTF8 names.


larseggert:
I opened #6355 for the spurious CI failure.

ietf/templates/nomcom/nomcom_public_base.html
```diff
@@ -9,13 +9,16 @@
 {% endblock %}
 {% block content %}
     {% origin %}
+    {% with parts=request.path|split:'/' %}
+    {% with l0=parts|slice:'1:'|last l1=parts|slice:'-2:-1'|last %}
+    {% with selected=l0|default:l1 %}
```
jennifer-richards, line 14: Clever but pushing the limits of what's a good idea in template logic. Might be better as a template filter if there's any chance we'd want this logic elsewhere.

ietf/templates/nomcom/nomcom_public_base.html
```diff
@@ -48,21 +51,24 @@ <h1>
         </li>
         {% if user|is_chair_or_advisor:year or user|has_role:"Secretariat" %}
             <li class="nav-item dropdown">
-                <a class="nav-link dropdown-toggle" data-bs-toggle="dropdown" href="#">Volunteers</a>
+                <a class="nav-link dropdown-toggle {% if selected in 'eligible,volunteers' %}active{% endif %}" data-bs-toggle="dropdown" href="#">Volunteers</a>
```
jennifer-richards, line 54: Similar issue to before, though less likely to be surprising since the second match doesn't contain the first.

jennifer-richards (changes requested):
I'm a little worried about hard-to-maintain template logic, see inline comments. I think probably the best thing would be to replace `if selected in "a,b"` with `if selected == "a" or selected == "b"` and leave the logic that computes the `selected` as it is.

jennifer-richards (approved):

jennifer-richards:
I think I gave some bad advice - I was looking at creating the template tag I suggested as an alternative to the explosion of "or" clauses in the `selected` tests and realized that it can be done like
```
{% if selected in "a,b,c,d,e,f"|split:"," %}
```
using what we've already got. Sorry for pointing you toward exploding the logic - if you'd like I can clean that up here or in a separate PR.

ietf/templates/nomcom/nomcom_public_base.html
```diff
@@ -49,7 +49,7 @@ <h1>
         </li>
         {% if user|is_chair_or_advisor:year or user|has_role:"Secretariat" %}
             <li class="nav-item dropdown">
-                <a class="nav-link dropdown-toggle {% if selected == 'eligible' or selected == 'volunteers' %}active{% endif %}" data-bs-toggle="dropdown" href="#">Volunteers</a>
+                <a class="nav-link dropdown-toggle {% if selected in 'eligibl,volunteers'|split:',' %}active{% endif %}" data-bs-toggle="dropdown" href="#">Volunteers</a>
```
jennifer-richards, line 52: Typo - `eligible` became `eligibl`

jennifer-richards (changes requested):

jennifer-richards (approved):

PR #6356, opened 2023-09-22
fix: Prevent duplicate HTML IDs on photo page

### LINKED ISSUE #6355: Duplicate HTML ID in`templates/group/all_photos.html`
### Describe the issue

See https://github.com/ietf-tools/datatracker/actions/runs/6263672611/job/17008584937?pr=6345

My guess is that there is an internationalized last name with a `last_initial` that gets Romanized to "c", and also some other name with a `last_initial` of "c".

### Code of Conduct

larseggert:
Fixes #6355

### DIFF
--- a/ietf/templates/group/all_photos.html
+++ b/ietf/templates/group/all_photos.html
@@ -8,7 +8,7 @@
     <h1>{{ group_type | upper }} {{ role }} photos</h1>
     {% regroup roles|dictsort:"last_initial" by last_initial as alphabet_blocks %}
     {% for letter in alphabet_blocks %}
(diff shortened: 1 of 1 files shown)


rjsparks:
This will break anyone that is used to typing http://localhost:8000/wg/photos/#h to get to the h's.
http://localhost:8000/wg/photos/#h8 is just not something they're going to guess.

It also just pushes the problem down the road (this problem exists on the current page) on doing the right thing with real names. Someone with a non [a-zA-Z] first character will end up with a section that is labelled with an [a-zA-Z] heading that is sorted in nothing matching _any_ standardized collation. Are ids restricted to [a-zA-Z]? Could we just use the first unicode character instead and make sure the results are sorted with the system's collation setting? This template is only used for working group chairs, so we're not likely to have random garbage thrown at production.

larseggert:
Turns out HTML5 IDs can be Unicode :-)

rjsparks (approved):

PR #6371, opened 2023-09-25
fix: Enable editing of related liaison statement information

### LINKED ISSUE #6264: The Related liaison field of the edit liaison form is broken
To see the issue, as the secretariat, choose a liaison statement such as /liaison/1854/ that has a related liaison statement and edit it. The related liaison statements field will be blank.

The form for creating one of these is likely broken in a similar way - the statement above ended up initially pointing to a completely unrelated statement, suggesting the wrong value was put into the field by the code.

larseggert:
Fixes #6264

### DIFF
--- a/ietf/liaisons/tests.py
+++ b/ietf/liaisons/tests.py
@@ -24,7 +24,7 @@ from ietf.utils.mail import outbox
 
 from ietf.group.factories import GroupFactory, RoleFactory
 from ietf.liaisons.factories import ( LiaisonStatementFactory, 
-    LiaisonStatementEventFactory, LiaisonStatementAttachmentFactory, )
+    LiaisonStatementEventFactory, LiaisonStatementAttachmentFactory, RelatedLiaisonStatementFactory)
 from ietf.liaisons.models import (LiaisonStatement, LiaisonStatementPurposeName,
     LiaisonStatementAttachment)
 from ietf.person.models import Person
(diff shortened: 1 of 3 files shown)


ietf/liaisons/tests.py
```diff
@@ -402,6 +404,13 @@ def test_edit_liaison(self):
         self.assertEqual(r.status_code, 200)
         q = PyQuery(r.content)
         self.assertEqual(len(q('form input[name=from_contact]')), 1)
+        json_data = q('form select[name=related_to]').attr('data-pre')
+        try:
+            decoded = json.loads(json_data)
+        except json.JSONDecodeError as e:
+            self.fail('data-pre contained invalid JSON data: %s' % str(e))
```
rjsparks, line 411: This is a little odd. The test is "Does the code that builds the form put JSON in the field". I guess it's better to break of there's something that's not clearly JSON there, but a lot could go in that would survive `json.loads` that isn't useful.

`self.fail` outside of something like `setUp` where you are using it to check that the test framework and factory/fixture data is sane is always a bit of a red flag.

A better test would look to see that the fieid contained the expected data?

rjsparks (approved):

ietf/liaisons/tests.py
larseggert: That pattern exists elsewhere in the tests, so we should fix it everywhere.

PR #6384, opened 2023-09-26
fix: State descriptions can have HTML in them, mark them safe

larseggert:


rjsparks:
Whiie this may be better in the short run, it moves us further away from the right place to be.

State descriptions (and DocEvent descriptions) should _NOT_ have html in them. If we want this stuff to be marked up, we should use a safer markup language (limited markdown perhaps). Or we should stop trying to mark it up.

rjsparks:
To be clear, since state descriptions can only be provided by code or by someone with admin access, we can take this PR, but we _really_ need to squeeze these pockets of html fragments out of the database.

rjsparks (approved):

PR #6387, opened 2023-09-26
feat: Remove document notes from the datatracker

### LINKED ISSUE #6013: Remove the IESG Note from the IESG writeups tab
The field for an RFC Editor note has not been used for a long time.
Lets remove it and adjust the places (like email sent) that used it.
cc: @larseggert @evyncke @wkumari

larseggert:
Fixes #6013


rjsparks:
Do we really want to remove the note from the database? (I've been on the fence about this). We have the option of removing the ability to edit/make new ones, but show old ones when they existed.

If we want _do_ want to remove it from the database, do we want to move the existing notes somewhere (a dump kept in  long-term offline storage, or shifting them to be comments on the document (using DocEvent)? Or is it ok just to throw the information away?

larseggert:
No, let's keep them in the db. What do I need to change?

rjsparks:
don't remove the column from the models and remove the migration that came from that change. Then look at the places where you removed how its shown - the emails and all can stay as you've changed them I think, but the html views should show them still if the note is non-empty.

The editing views should all stay removed as the PR currently removes them - if any need to change in the future, the secretariat can[^1] change them using the admin.

---

^1 : We have a separate issue that the secretariat's access to the admin is spotty and needs tuning.

rjsparks (approved):

jennifer-richards (approved):

PR #6392, opened 2023-09-27
refactor: Return objects from get_persons api

jennifer-richards:
This is an alternative implementation of the `rpc_persons()` api endpoint. The motivation is to preserve `Person.id` as an integer rather than forcing it to be converted to a string so it can act as a JSON object key.

(Incidentally, I've called the name `plain_name` in the response to reuse the existing `rpc_person()` return value schema. I have no objection to calling it `name` instead, I just haven't changed that here.)

### DIFF
--- a/ietf/api/views.py
+++ b/ietf/api/views.py
@@ -462,9 +462,12 @@ def rpc_persons(request):
         return HttpResponseForbidden()
     
     pks = json.loads(request.body)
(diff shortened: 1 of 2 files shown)


jennifer-richards:
Closing, going to stick with the initial API until there's a stronger reason to complicate it.

PR #6397, opened 2023-09-28
feat: Add the ability to rerun a test suite until a failure occurs

larseggert:
More quickly than looping in the shell around a non-zero error code. This only incurs all the setup and scaffolding overheads once.

### DIFF
--- a/ietf/utils/test_runner.py
+++ b/ietf/utils/test_runner.py
@@ -1141,7 +1141,7 @@ class IetfTestRunner(DiscoverRunner):
 
         self.test_apps, self.test_paths = self.get_test_paths(test_labels)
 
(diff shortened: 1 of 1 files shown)


ietf/utils/test_runner.py
```diff
@@ -1129,6 +1141,7 @@ def run_tests(self, test_labels, extra_tests=None, **kwargs):
 
         self.test_apps, self.test_paths = self.get_test_paths(test_labels)
 
+        self.test_labels = test_labels  # these are used in out run_suite() and not available to it otherwise
```
larseggert, line 1144:
```suggestion
        self.test_labels = test_labels  # these are used in our run_suite() and not available to it otherwise
```

rjsparks (approved):

jennifer-richards (approved):

PR #6398, opened 2023-09-28
refactor: Allow existing Person in create_demo_person

jennifer-richards:
Instead of returning a 403 Forbidden error, just return the existing `Person` if a `create_demo_person()` API call tries to create a person who already exists. This avoids having to clean out the datatracker DB when re-running `create_rpc_demo` in the rpc tool.

Could instead handle the 403 (or return a distinct indicator that the name is in use, perhaps a 409 Conflict) and deal with this on the API caller side, but this is the lower hanging fix


rjsparks (approved):

jennifer-richards:
Oops, thought I'd merged

PR #6407, opened 2023-09-29
chore: Pin WeasyPrint to avoid bug in v60.0

jennifer-richards:
Keep the pin until the fix for https://github.com/Kozea/WeasyPrint/issues/1973 is released and confirmed to work with datatracker.


kesara (approved):

jennifer-richards:
https://pypi.org/project/weasyprint/60.1/ is out, so closing this - we should just have to update requirements.

PR #6410, opened 2023-09-29
fix: restore interim announce ability

rjsparks:
Intentionally targeting the release branch - this is an emergency patch.

The changes to use disabled instead of readonly cause the browsers to not include the fields in the POST, rendering the form invalid. (The page doesn't show the form errors - I'll open an issue for that).

It would be very nice to find a way to have this kind of breakage detected in tests.

cc: @larseggert

### DIFF
--- a/ietf/templates/meeting/interim_send_announcement.html
+++ b/ietf/templates/meeting/interim_send_announcement.html
@@ -13,7 +13,7 @@
         {% csrf_token %}
         <div class="row mb-3">
             <label for="{{ form.to.id_for_label }}" class="col-md-2 fw-bold col-form-label">To</label>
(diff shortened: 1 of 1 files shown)


EllaMartinez1981 (commented):
ietf/templates/meeting/interim_send_announcement.html

Fitzjust4ullc:
> Intentionally targeting the release branch - this is an emergency patch.
> 
> The changes to use disabled instead of readonly cause the browsers to not include the fields in the POST, rendering the form invalid. (The page doesn't show the form errors - I'll open an issue for that).
> 
> It would be very nice to find a way to have this kind of breakage detected in tests.
> 
> cc: @larseggert 

c5a955d7984d053bc949dfefd35ae8bd9ebc4365

PR #6424, opened 2023-10-03
feat: Update IRTF state when entering/exiting IETF conflict review

### LINKED ISSUE #3469: IRTF state doesn't change when draft sent for IESG conflict review
`type_defect`   |    by csp@csperkins.org

___

I have a draft in IRTF state "Waiting for IRTF Chair" and press "Begin IETF Conflict Review". The conflict review is correctly started and appears in the datatracker, but the IRTF state remains unchanged as "Waiting for IRTF chair". I would expect the state to change – probably to "In IESG review"?

Then, when the IESG conflict review is completed, the IRTF state should change back to "Waiting for IRTF Chair" with tag "IESG Review Completed".

___
_Issue migrated from trac:3469 at 2022-03-04 09:20:16 +0000_

pselkirk:
Fixes #3469

### DIFF
--- a/ietf/doc/views_conflict_review.py
+++ b/ietf/doc/views_conflict_review.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2012-2020, All Rights Reserved
+# Copyright The IETF Trust 2012-2023, All Rights Reserved
 # -*- coding: utf-8 -*-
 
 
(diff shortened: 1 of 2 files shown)


ietf/doc/views_conflict_review.py
```diff
@@ -536,3 +551,24 @@ def start_review_as_stream_owner(request, name):
                                'doc_to_review': doc_to_review,
                               },
                           )
+
+def start_review_irtf_state(doc, by):
+    prev_state = doc.get_state('draft-stream-irtf')
+    new_state = State.objects.get(type_id='draft-stream-irtf', slug='iesg-rev')
+    doc.set_state(new_state)
```
rjsparks, line 558: Consider adding the new_state != prev_state guard here as well

rjsparks (approved):

PR #6428, opened 2023-10-04
fix: Change some colors to make agenda tool work in dark mode

larseggert:
This is not a proper fix. This tool doesn't use bootstrap and is hence full of custom CSS. Really needs a redo.


larseggert:
I have no idea why Selenium thinks changing colors affects the clickability of things.

NGPixel (approved):

PR #6430, opened 2023-10-04
feat: New endpoint to get Person by OIDC subject id

jennifer-richards:


jennifer-richards:
Going to close - we don't need this (at least not yet, but maybe not ever)

PR #6432, opened 2023-10-04
fix: Remove incorrect line from submission API info web page

DavidSchinazi:
Now that https://github.com/ietf-tools/datatracker/pull/4037 has landed and https://github.com/ietf-tools/datatracker/issues/3286 has been resolved, this line can be removed.

### DIFF
--- a/ietf/templates/submit/api_submit_info.html
+++ b/ietf/templates/submit/api_submit_info.html
@@ -27,6 +27,7 @@
     </p>
     <ul>
         <li>Only XML-only uploads are supported, not text or combined.</li>
+        <li>Document replacement information cannot be supplied.</li>
         <li>
             The server expects <code>multipart/form-data</code>, supported by <code>curl</code> but <b>not</b> by <code>wget</code>.
         </li>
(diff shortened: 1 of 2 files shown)


ietf/templates/submit/api_submit_info.html
```diff
@@ -27,7 +27,6 @@ <h1 class="mb-3">Internet-Draft submission API instructions</h1>
     </p>
     <ul>
         <li>Only XML-only uploads are supported, not text or combined.</li>
-        <li>Document replacement information cannot be supplied.</li>
```
jennifer-richards, line 30: This one needs to stay - the replaces is only supported through the asynchronous endpoint.

jennifer-richards (changes requested):
Thank you for the patch - as noted inline, only the asynchronous endpoint (`api_submission`) accepts replacement information so we should probably keep the comment on the old endpoint's docs.

ietf/templates/submit/api_submit_info.html
DavidSchinazi: Ah good catch, thanks! Fixed.

jennifer-richards (approved):

PR #6438, opened 2023-10-05
chore: Prevent test suite artifact creation in work directory

### LINKED ISSUE #3520: temporary directories not cleaned up after tests
`type_defect`   |    by jennifer@painless-security.com

___

Two temporary directories are being left around after running the tests:
```
$ ls -ld tmp-*
drwxr-xr-x 55 root root 4096 Jan 24 17:26 tmp-nomcom-public-keys-dir
drwxr-xr-x  5 root root 4096 Jan 24 17:06 tmp-storage-meetinghosttests-dir
```
These should be cleaned up when the tests are done with them.

___
_Issue migrated from trac:3520 at 2022-03-04 09:25:31 +0000_

### LINKED ISSUE #4020: Make test suite generate files outside of the source tree
### Description

The tests seem to generate directories in the source tree names `tmp-*` and they also seem to generate files under `test/media` and possibly `test` in general.

It would be nice if that as reworked so all that happened under `/tmp`.

### Code of Conduct

larseggert:
Also remove a few other stale test assets while I'm here.

Fixes #4020
Fixes #3520


dev/tests/settings_local.py
```diff
@@ -17,8 +17,8 @@
 }
 
 IDSUBMIT_IDNITS_BINARY = "/usr/local/bin/idnits"
-IDSUBMIT_REPOSITORY_PATH = "test/id/"
-IDSUBMIT_STAGING_PATH = "test/staging/"
+IDSUBMIT_REPOSITORY_PATH = "/assets/ietf-ftp/internet-drafts/"
```
rjsparks, line 20: This should go into `/assets/ietfdata/doc/draft/archive` to match the tree currently in production at ietfdata (and things that are currently in www6 will move into that tree as we move all of this into a blob-store).

Similarly the repository should be at `/assets/ietfdata/doc/draft/repository`

and other document types are simpler:

```
./doc/bofreq
./doc/charter
./doc/conflict-review
./doc/review
./doc/slides # This one is strange - it's edu + iab ws slides at the moment, not meeting materials slides
./doc/status-change
./doc/statement
```

dev/tests/settings_local.py
```diff
@@ -17,8 +17,8 @@
...
+IDSUBMIT_STAGING_PATH = "/assets/www6s/staging/"
```
rjsparks, line 21: in the long run, staging needs to be reworked anyhow to be in db, not on disk. I'd like to squeeze www6s out of the dev container configuration, but until we pull everything out of the proceedings dir (where the meeting materials are currently living), we can keep /assets/www6s to reflect the production reality.

rjsparks (changes requested):

dev/tests/settings_local.py
larseggert: "This should go" - which this? `IDSUBMIT_REPOSITORY_PATH` or `IDSUBMIT_STAGING_PATH`?

rjsparks: `IDSUBMIT_REPOSITORY_PATH`

Thanks for asking - my direction above was a little off.
In production right now:
https://github.com/ietf-tools/datatracker/blob/a02812bf7c7177471af18141bcf6cae6375d9ee8/ietf/settings.py#L674
and 
https://github.com/ietf-tools/datatracker/blob/a02812bf7c7177471af18141bcf6cae6375d9ee8/ietf/settings.py#L799

We should mirror that in the dev instance (it's important that the submisssion tool post draft into the repository, not the archive).

That is, we should have this in the dev instance:
`IDSUBMIT_REPOSITORY_PATH = /assets/ietfdata/doc/draft/repository`

larseggert:
@rjsparks 
>CRITICALS:
?: (datatracker.E0006) A directory used by the I-D submission tool does not
exist at the path given in the settings file.  The setting is:
    IDSUBMIT_REPOSITORY_PATH = /assets/ietfdata/doc/draft/repository

larseggert:
There are a ton of directories that are being created in `docker/scripts/app-create-dirs.sh` - it's unclear to me how many of them are actually needed to run a dev instance and the CI.

larseggert:
@rjsparks I think I made all the changes you requested?

rjsparks:
Yes, I'm holding this for more work post the feat/rfc merge.

rjsparks:
This didn't get returned to post feat/rfc. We should review it fresh to see if it's still the right thing to do.
cc: @jennifer-richards

jennifer-richards:
I left a couple comments - cleaning up after the tests, even in `/tmp` or wherever those paths end up, would be nice but also tricky. If the tests pass against modern main, I would be inclined to accept this as an improvement and treat those future work.

ietf/settings_test.py
```diff
@@ -48,11 +49,12 @@ def __getitem__(self, item):
 if TEST_CODE_COVERAGE_CHECKER and not TEST_CODE_COVERAGE_CHECKER._started: # pyflakes:ignore
     TEST_CODE_COVERAGE_CHECKER.start()                          # pyflakes:ignore
 
-NOMCOM_PUBLIC_KEYS_DIR=os.path.abspath("tmp-nomcom-public-keys-dir")
+NOMCOM_PUBLIC_KEYS_DIR = tempfile.mkdtemp(suffix="-nomcom-public-keys-dir")
```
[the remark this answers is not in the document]
jennifer-richards: Fixed in e9bc0ae29ecedf9eb1caedf3e9b7c29ff0f6a21a

jennifer-richards:
Hopefully not stepping on @larseggert's toes, but just pushed changes that address my concerns. I think it's ready for @rjsparks to re-review?

larseggert:
No toes in the area :-) I won't get to this any time soon, so please do what you need to!

ietf/person/factories.py
```diff
@@ -103,10 +103,9 @@ def default_photo(obj, create, extracted, **kwargs): # pylint: disable=no-self-a
             media_name = "%s/%s.jpg" % (settings.PHOTOS_DIRNAME, photo_name)
             obj.photo = media_name
             obj.photo_thumb = media_name
-            photosrc = os.path.join(settings.TEST_DATA_DIR, "profile-default.jpg")
             photodst = os.path.join(settings.PHOTOS_DIR,  photo_name + '.jpg')
-            if not os.path.exists(photodst):
-                shutil.copy(photosrc, photodst)
+            img = Image.new('RGB', (200, 200))
+            img.save(photodst)
```
rjsparks, line 108: remind me why we're not just making a copy of the profile-default.jpg again?
(I'm sensitive to the PIL import - it's large)

media/photo/profile-default.jpg
rjsparks, line 1: Doesn't the non-test code _use_ this for displaying the profile page when no picture is available?

rjsparks (commented):
I am a little nervous at the "clean up on exit" vs "clean up as you go" approach as it adds to the recovery for the developer if the process goes away via `kill -9` or the like, but am willing to try it.

Just one question about why we're trading a file that I _think_ we should be keeping for PIL.

jennifer-richards:
> I am a little nervous at the "clean up on exit" vs "clean up as you go" approach as it adds to the recovery for the developer if the process goes away via `kill -9` or the like, but am willing to try it.

I'm sympathetic. The `kill -9` doesn't worry me terribly because it's rare and the impact is just a couple of directories in `/tmp` left over and `atexit` is robust against most normal exits.

I'd _much_ rather create the temp dir separately for each test to maintain isolation. Before adding the `atexit` business I tried using the temp directory mechanism we have in our `TestCase` class and it kinda sorta worked, but ran into bad interactions with the test startup sequence. It definitely interacted with our checks for the directories existing, but IIRC the statics serving for tests was also broken by this because it's only set up once for the entire test run.

media/photo/profile-default.jpg
jennifer-richards: (and for the thumbnail, we use an icon - it probably resembles the profile-default image so maybe that's what you're thinking of?)

https://github.com/ietf-tools/datatracker/blob/a6377997953ced70c78c8c3a3a33ee17a31dc8f3/ietf/templates/person/photo.html#L13-L15

ietf/person/factories.py
jennifer-richards: I think it's just avoiding a file that (as far as I can tell) exists only for the tests.

I don't have an opinion one way or the other, but maybe there's a lighter weight library for generating an empty image for testing.

jennifer-richards: There's PyPNG - much smaller it seems, but requires less clearly intentioned code to generate a blank image and is somewhat slower on a per-image basis. The import of PIL on my system looks to be a few tens of ms - and I think we probably only hit that once per test run. (Though I haven't actually done a profile to see if the import is being exercised more than once; not sure how to do that)

rjsparks (approved):

ietf/person/factories.py
rjsparks: I suspect there's a valid, small, literal we can use as a PNG, but not going to slow this down over it.
I think we should, however, spend time scrubbing our full dependency set and seeing if we can make it smaller, and PIL is one of the things I'm wondering if we can scrub completely away.

jennifer-richards: We use PIL for resizing images in `Person.photo`. If we do away with that (mis?)feature, then we can probably ditch even needing to fake an image. 

Checked and PyPNG is a < 40kb python file. It can read and write PNGs, so if we _do_ need to create one for some reason, it's a pretty minimal bit of baggage.

PR #6442, opened 2023-10-06
fix: Add remote instructions and date/time to interim agenda page

### LINKED ISSUE #6249: Interim meetings don't show a date and time
### Describe the issue

https://datatracker.ietf.org/doc/agenda-interim-2023-sframe-01-sframe-01/ does not include a date and time, except via the agenda.

However, the event is created with a date and time.  Email from datatracker shows:

> The Secure Media Frames (sframe) WG will hold a virtual interim meeting on 2023-09-13 from 15:00 to 16:00 UTC.

Maybe the web page could include this rather critical information.

### Code of Conduct

### LINKED ISSUE #6433: It's counterintuitive that remote participation links are under "Materials"
### Describe the issue

Several times I've been confused and not able to find remote participation information for a MeetEcho meeting I'm in, because I'm not expecting it to be under "Materials".  (I'd expect this to just be agenda, slides, eventually minutes, etc.). I was in another session today where someone else had the same problem.

Can this either be named better, or else could we have this link duplicated under "Agenda" or somewhere else?  Or maybe even a button for it on the main meetings page for the meetings currently in progress.

### Code of Conduct

larseggert:
Fixes #6433
Fixes #6249


rjsparks:
We should also put the date and time on the group meetings page.
For future and in-progress sessions, the time should show here somewhere:
image

rjsparks (changes requested):
(see the above comments on the discussion tab)

rjsparks (changes requested):
Thank you for continuing to improve things for meeting participants.

I think we need a few more touches before merging this -

at http://localhost:8000/meeting/interim-2023-cbor-16/session/cbor I see:

image

It would be good to put the timezone selectors there as well (instead of the current showing of timezone declared for the meeting), again with default to local (?).

At http://localhost:8000/doc/agenda-interim-2023-cbor-16-cbor-01/ I see:
image

The time happens to be in Local (I think) but the Local cell is not selected so it's not clear what the timezone is until clicking around a bit.

Also, when the `Meeting` option is pressed, the result is:
image
which says nothing about what timezone that actually is?

larseggert:
@rjsparks last commit should fix your issues. I couldn't reproduce "Local cell is not selected", it is selected in my dev instance. (Did you `yarn legacy:rebuild`?)

rjsparks (approved):

Issue #5440, opened 2023-03-29
Let people know when their slides have been accepted


paulehoffman:
### Description

When someone submits slides in the Datatracker, and the WG chair accepts their submission, the slides go into the WG list but the submitter doesn't know that. It would be nice if the submission causes an email to the submitter.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
This will involve a new mailtrigger, and the code to cause mail to be emitted towards the recipients (which should be the slide submitters and the group chairs) added just before https://github.com/ietf-tools/datatracker/blob/9a09bf1f7b2b6ab100482090012422263f85207a/ietf/meeting/views.py#L4558

rjsparks:
Fixed by #6470

PR #6470, opened 2023-10-12
feat: Send mail when slides are approved

pselkirk:
Implements #5440

### DIFF
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -4555,6 +4555,17 @@ def approve_proposed_slides(request, slidesubmission_id, num):
                 submission.status = SlideSubmissionStatusName.objects.get(slug='approved')
                 submission.doc = doc
                 submission.save()
+                (to, cc) = gather_address_lists('slides_approved', group=submission.session.group, proposer=submission.submitter).as_strings()
+                msg_txt = render_to_string("meeting/slides_approved.txt", {
+                    "to": to,
+                    "cc": cc,
+                    "submission": submission,
+                    "settings": settings,
+                })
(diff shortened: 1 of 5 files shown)


ietf/meeting/views.py
```diff
@@ -4555,6 +4555,17 @@ def approve_proposed_slides(request, slidesubmission_id, num):
...
+                msg_txt = render_to_string("meeting/slides_approved.txt", {
+                    "to": to,
+                    "cc": cc,
+                    "submission": submission,
+                    "settings": settings,
+                })
+                msg = infer_message(msg_txt)
+                msg.by = request.user.person
+                msg.save()
+                send_mail_message(request, msg)
```
rjsparks, line 4568: I see this pattern is used in a few places in the code but it is unnecessary here (and in one other spot that I you probably based this off of). Using infer_message is really for flows where we've received an entire message from outside or have a message stored for review where headers should be editable as part of that review (some liaison flows work this way), leading to the message being sent later. 

Using it for a message that you're building and sending right away does a lot of overwork (and runs the risk of missing some message conditioning that should usually be performed).

Instead, change the template to just generate the body. Build the subject in the code here.
Then make a call like:
```
    send_mail_text(request, to_email, None, subject, body, cc=cc)
```
Its worth reading through what send_mail_text does.

In particular, the None for frm will be turned into something sensible here:
https://github.com/ietf-tools/datatracker/blob/fa00abf9de8f68c72c9a5f7c72752e1d480324b6/ietf/utils/mail.py#L58-L65

rjsparks (changes requested):

rjsparks (approved):

PR #6496, opened 2023-10-18
fix: Don't allow group chair to change group parent

### LINKED ISSUE #6037: Group edit form destroys group parent for rswg chairs
When an rswg chair edits the group (to add extresources for example), the control for parent is present (it should not be) with the only choice available set to "None".

pselkirk:
Fixes #6037

### DIFF
--- a/ietf/group/tests_info.py
+++ b/ietf/group/tests_info.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2009-2022, All Rights Reserved
+# Copyright The IETF Trust 2009-2023, All Rights Reserved
 # -*- coding: utf-8 -*-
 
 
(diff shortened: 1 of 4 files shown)


ietf/group/tests_info.py
```diff
@@ -950,6 +950,31 @@ def test_edit_description_field(self):
...
+    def test_edit_parent_field(self):
+        group = GroupFactory(acronym='mars', parent=Group.objects.get(acronym='farfut'))
+        RoleFactory(group=group, name_id='chair')
+        url = urlreverse('ietf.group.views.edit', kwargs={'acronym': group.acronym, 'action': 'edit', 'field': 'parent'})
+
+        # parent is not shown to group chair
+        login_testing_unauthorized(self, 'mars-chair', url)
+        r = self.client.get(url)
+        self.assertEqual(r.status_code, 302)
+        self.assertEqual(len(r.content), 0)
```
rjsparks, line 962: It's not clear what the intent of this test block is? The view that would only edit the parent field should probably just refuse entry?

What about testing the form that allows editing the whole group (when you don't pass in field to the view). - I would expect a test that shows that the parent field isn't present on that form.

rjsparks (changes requested):

ietf/group/tests_info.py
pselkirk: For the chair, /edit/parent/ behaves exactly like /edit/xyzzy/ - it renders an edit page with no editable fields, just a Submit button and a Back button, which both lead back to the about page.

However, I bolluxed the user creation, and what's actually going on there is a redirection to the login page. 🤦 

But yeah, I can test the whole group form.

rjsparks (changes requested):
Thanks for adjusting the tests. This is close to ready but has one more change that should be made.

Right now, in this branch, /group/rswg/edit/parent presents the following:

image

There is risk of unexpected happening with submitting a null-form to the current code (which hunts for fields).

I suspect the current code would also let someone provide the parent field manually here (or on the full edit form) in a post and the value would be honored.

Please test those posts (attempting to change parent as a chair should result in no change).

I think the form view at /group/rswg/edit/parent should 403 unless you have the roles you've already identified for showing the field.

pselkirk:
It turns out that the form is accepted (form.is_valid==True), but form.cleaned_data only contains the fields that are shown to the user - in the case of editing just the parent field, it's an empty dict.

rjsparks:
I still think it's the right time to change the view behavior to 403 if there's no field to edit. (such as with your xyzzy example).

pselkirk:
Unit testing on github is failing with `'ietf/name/models.py:104: error: Need type annotation for "next_states"  [var-annotated]'`. Did something change in the settings?

rjsparks (approved):

PR #6509, opened 2023-10-20
refactor: remove unused secr/areas app

rpcross:


ietf/secr/urls.py
```diff
@@ -4,8 +4,6 @@
 urlpatterns = [
     re_path(r'^$', TemplateView.as_view(template_name='main.html')),
     re_path(r'^announcement/', include('ietf.secr.announcement.urls')),
-    re_path(r'^areas/', include('ietf.secr.areas.urls')),
-    re_path(r'^console/', include('ietf.secr.console.urls')),
```
rjsparks, line 8: suspect you didn't mean to catch console here, but leave it be.

rjsparks (approved):

PR #6520, opened 2023-10-23
fix: Add "numeric" option to localecompare

### LINKED ISSUE #6493: Numerical fields being sorted alphabetically
### Describe the issue

When looking at the IESG dashboard there will be a bunch of numerical fields that are sorted alphabetically rather than numerically. (when sorting from low -> high you get something like 1, 102, 11)

### Code of Conduct

larseggert:
Fixes #6493

### DIFF
--- a/ietf/static/js/list.js
+++ b/ietf/static/js/list.js
@@ -14,7 +14,8 @@ function text_sort(a, b, options) {
     // sort by text content
     return prep(a, options).localeCompare(prep(b, options), "en", {
         sensitivity: "base",
-        ignorePunctuation: true
+        ignorePunctuation: true,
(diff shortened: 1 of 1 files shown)


larseggert:
It would, because 2 < 100.

This _should_ restore the "natural sort" algorithm we had before.

rjsparks:
That ducks my question. Where is the standard that defines this behavior. A quick attempt at finding it goes into the weeds of the Unicode Technical Standard chasing `"kn"`, but I'm not finding the thing that says `foo 2 bar` comes before `foo 100 bar` even there.

rjsparks (approved):

PR #6527, opened 2023-10-24
feat: Render markdown agendas with bs5 styles 

larseggert:
So that the look is consistent with the rest of the datatracker.


rjsparks (approved):

ietf/meeting/views.py
```diff
@@ -279,15 +279,22 @@ def materials_document(request, document, num=None, ext=None):
...
-                    bytes = "<html>\n<head><base target=\"_blank\" /></head>\n<body>\n%s\n</body>\n</html>\n" % markdown.markdown(bytes.decode(encoding=chset))
-                    content_type = content_type.replace('plain', 'html', 1)
-                    break;
-                elif atype[0] == 'text/plain':
-                    break;
+                if atype[0] == "text/markdown":
+                    content_type = content_type.replace("plain", "markdown", 1)
+                    break
+                elif atype[0] == "text/html":
+                    bytes = render(
```
rjsparks, line 286: This is building an HttpResponse object, not a string (thanks to @kesara for spotting it).

rjsparks (changes requested):

ietf/static/css/ietf.scss
```diff
@@ -47,7 +47,7 @@ url("npm:bootstrap-icons/font/fonts/bootstrap-icons.woff") format("woff");
 @import "bootstrap-icons/font/bootstrap-icons";
 
 // Leave room for fixed-top navbar...
-body {
+body.navbar-offset {
```
rjsparks, line 50: This and the change to base seem out of scope for this PR subject. Please make things like this in their own PR going forward.

rjsparks (approved):

PR #6530, opened 2023-10-24
fix: BCPs can normatively cite all other standards levels

### LINKED ISSUE #6524: Logic to indicate downrefs from BCPs appears broken
### Describe the issue

The datatracker is indicating that references from BCPs to Proposed Standards and Draft Standards are downrefs.

For example, see these reference lists from three example BCPs or BCPs-to-be:
https://datatracker.ietf.org/doc/draft-ietf-tsvwg-ecn-encap-guidelines/references/
https://datatracker.ietf.org/doc/draft-ietf-dnsop-avoid-fragmentation/references/
https://datatracker.ietf.org/doc/rfc7141/references/

Other systems that depend on these incorrect downref categorizations are also affected, e.g. I first notice this issue in the last call announcement for draft-ietf-tsvwg-ecn-encap-guidelines

However, 1 of the 4 downrefs listed on the references page for ecn-encap-guidelines (rfc3168), was not listed in the last call announcement, even though it had the same maturity status as the others (PS). 
See https://datatracker.ietf.org/doc/draft-ietf-tsvwg-ecn-encap-guidelines/history/
Might this indicate another bug unmasked by the first bug?

### Code of Conduct

larseggert:
Fixes #6524

### DIFF
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -724,7 +724,7 @@ class RelatedDocument(models.Model):
         if source_lvl == "std" and target_lvl in ["ps", "ds"]:
             return pos_downref
 
-        if source_lvl in ["ps", "ds", "unkn"] and target_lvl == "unkn":
+        if source_lvl not in ["inf", "exp"] and target_lvl == "unkn":
             return "Possible Downref"
 
         if source_lvl == "unkn" and target_lvl in ["ps", "ds"]:
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -702,7 +702,7 @@ class RelatedDocument(models.Model):
             else:
                 target_lvl = self.target.document.intended_std_level.slug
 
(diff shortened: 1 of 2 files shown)


ietf/doc/models.py
```diff
@@ -702,7 +702,7 @@ def is_downref(self):
             else:
                 target_lvl = self.target.document.intended_std_level.slug
 
-        rank = { 'ps':1, 'ds':2, 'std':3, 'bcp':3 }
+        rank = { 'ps':1, 'ds':2, 'std':3, 'bcp':1 }
```
rjsparks, line 705: This is not doing what you intend.

Add tests to prove that the result is what is intended.

I'd look more closely at the conditional to see if we have a logic inversion there.

rjsparks (changes requested):

rjsparks (approved):

PR #6534, opened 2023-10-24
feat: Revamp AD dashboard

### LINKED ISSUE #6528: States in Conflict Review listing of AD Dashboard are not in the right order
### Describe the issue

In the "Conflict Review State Counts" table of the AD Dashboard (https://datatracker.ietf.org/doc/ad), the counts of conflict per state are listed.  The columns, each representing a state, should be in the progression by which the states are used.  A conflict starts with "Needs Shepherd" --> "AD Review" --> "IESG Evaluation", etc.  To that end, "Needs Shepherd" should come before "AD Review" (i.e., swap the order of these two columns).

### Code of Conduct

### LINKED ISSUE #6553: Documents are incorrectly listed in "Missing ballot positions" of the AD specific dashboard
### Describe the issue

The AD-specific dashboard (e.g. https://datatracker.ietf.org/doc/ad/roman.danyliw) has a section for documents "Missing ballot positions".  It appears to show documents which have not been scheduled for a telechat and/or even been to a telechat.  In my assessment that is not a "missing ballot".  Only document which have been reviewed at a telechat (i.e., been in "IESG Review" with a telechat date in the past) and I have not balloted on (but are not approved) should be shown.

This condition is likely temporal so I'm currently seeing https://datatracker.ietf.org/doc/draft-ietf-opsawg-rfc7125-update/ listed for me.  The responsible AD issued a ballot on 10/27/2023.  It it has not been to a telechat.  Has no date assigned.  Has not been "IESG Review".

### Code of Conduct

### LINKED ISSUE #6608: https://datatracker.ietf.org/doc/ad/ needs help text
### Description

The AD status page is probably useful to ADs.  It is opaque to those who haven't been on the IESG.  I suggest that each table title (I-D State Counts, etc) be an anchored link into a file explaining what the columns mean.

### Code of Conduct

larseggert:
Show state queue graphs to logged-in ADs instead of counts and trend arrows.

Fixes #6528
Fixes #6553
Fixes #6608


ietf/doc/views_search.py
```diff
@@ -345,44 +345,35 @@ def ad_dashboard_group(doc):
         if doc.get_state_slug('draft') == 'rfc':
             return 'RFC'
         elif doc.get_state_slug('draft') == 'active' and doc.get_state_slug('draft-iesg'):
-            return '%s Internet-Draft' % doc.get_state('draft-iesg').name
+            return str(doc.get_state('draft-iesg').name)
```
rjsparks, line 348: Why are you constructing a str from a str here (and throughout the PR)?

ietf/doc/views_search.py
```diff
@@ -345,44 +345,35 @@ def ad_dashboard_group(doc):
...
     elif doc.type.slug=='conflrev':
         if doc.get_state_slug('conflrev') in ('appr-reqnopub-sent','appr-noprob-sent'):
-            return 'Approved Conflict Review'
+            return 'Approved'
         elif doc.get_state_slug('conflrev') in ('appr-reqnopub-pend','appr-noprob-pend','appr-reqnopub-pr','appr-noprob-pr'):
-            return "%s Conflict Review" % State.objects.get(type__slug='draft-iesg',slug='approved')
+            return str(State.objects.get(type__slug='draft-iesg',slug='approved'))
         else:
-          return '%s Conflict Review' % doc.get_state('conflrev')
+          return str(doc.get_state('conflrev'))
```
rjsparks, line 357: here, why aren't you just getting the state name?
See https://github.com/ietf-tools/datatracker/blob/601ab53db9619c68b191b8d42902a1e59cb636bd/ietf/doc/models.py#L89C1-L90C25

ietf/doc/views_search.py
```diff
@@ -345,44 +345,35 @@ def ad_dashboard_group(doc):
...
         if doc.get_state_slug('statchg') in ('appr-pend','appr-pr'):
-            return '%s Status Change' % State.objects.get(type__slug='draft-iesg',slug='approved')
+            return str(State.objects.get(type__slug='draft-iesg',slug='approved'))
         else:
-            return '%s Status Change' % doc.get_state('statchg')
+            return str(doc.get_state('statchg'))
     elif doc.type.slug=='charter':
         if doc.get_state_slug('charter') == 'approved':
-            return "Approved Charter"
+            return "Approved"
```
rjsparks, line 367: Not sure why this is done differently than just getting the state name?
```
>>> State.objects.get(type='charter', slug='approved').name
'Approved'
```

rjsparks (changes requested):

ietf/doc/views_search.py
larseggert: Because one of these returns didn't return a string

ietf/doc/views_search.py
larseggert: Don't know either, I just kept the logic from before

larseggert:
Let me do the bigger revamp on this, rather than spending time on the current version.

larseggert:
I think it's ready, but it's not urgent

rjsparks (approved):
The help/states change really should have come in its own PR.

The dashboard also needs a deeper refactor to _not_ use the search to find things, but to go after the information directly from the models. There's too much having to lateral-translate one way and then back the other in the same view.

I'm worried about still more display (and token) transformations creeping in as utility functions, especially lying around in search. These are attractive nuisances and should be factored away so other things don't start using them.

All that said, this is the IESGs page and I won't slow it down.

PR #6549, opened 2023-10-26
fix: Don't include csrf_token in GET-only form

### LINKED ISSUE #6487: Better grooming of created queryset content for search results is needed.
Right now, if you type "dkim" into the main page searchbox, you are redirected to a search result page that has this URL:

`https://datatracker.ietf.org/doc/search?csrfmiddlewaretoken=<actual token elided>&name=dkim&sort=&rfcs=on&activedrafts=on&by=group&group=`

This is silly. The issue is that the view is blindly copying things from the submitted form into the queryset.
It should, instead, only copy the things that make sense. The csrfmiddlewaretoken should never be copied. Any thing that would result in key= with no value should be elided.

pselkirk:
Fixes #6487

### DIFF
--- a/ietf/templates/doc/search/search_form.html
+++ b/ietf/templates/doc/search/search_form.html
@@ -6,7 +6,6 @@
 <form id="search_form"
       class="form-horizontal"
       action="{% url 'ietf.doc.views_search.search' %}">
-    {% csrf_token %}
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
lgtm

PR #6567, opened 2023-10-30
fix: build proceedings attendee list from MeetingRegistration table. …

### LINKED ISSUE #6265: Some meeting participants listed twice on attendees list in proceedings
### Describe the issue

The attendees page in the proceedings page for 117 (and perhaps others) lists many, but not all, people twice. Scanning the list, it appears plausible that this is related to hackathon registration, but that's just a guess.

https://datatracker.ietf.org/meeting/117/proceedings/attendees/

Not exactly related, but in the same space as #5928

### Code of Conduct

rpcross:
…Fixes #6265


rjsparks:
This leaves some abandoned dbtemplates - it's reasonable to leave them in place in the short term should we need to fall back, but we'll need to track an action to remove them. Leaving this comment as a note to create an issue to track that task when this PR is merged.

rjsparks (approved):

jennifer-richards (commented):
I have a couple comments inline, but I think this is ok, assuming the duplicate `MeetingRegistration` data get cleaned up.

If the older data likely have similar issues, de-duplicating the records before displaying them would be something to consider. That would probably need to be paired with another method for detecting the problem, though, so that duplicates are not simply swept under the rug.

ietf/nomcom/utils.py
```diff
@@ -678,6 +678,16 @@ def three_of_five_eligible_8713(previous_five, queryset=None):
...
 
+def participants_for_meeting(meeting):
+    """ Return a tuple (checked_in, attended)
+        checked_in = set of onsite, checkedin participants (Person.pk)
+        attended = set of remote participants who attended a session (Person.pk)
+    """
+    checked_in = meeting.meetingregistration_set.filter(reg_type='onsite', checkedin=True).values_list('person', flat=True)
+    sessions = meeting.session_set.filter(Q(type='plenary') | Q(group__type__in=['wg', 'rg']))
+    attended = Attended.objects.filter(session__in=sessions).values_list('person', flat=True)
+    return (set(checked_in), set(attended))
```
rjsparks, line 689: evaluating these to sets here is not right - we should return querysets and let whatever called it chose when to evaluate. (I also agree this should move to meeting)

rjsparks (changes requested):
Jennifer's comments cause me to think about method signatures as well, and I do want some changes before we merge this.

ietf/nomcom/utils.py
rpcross: Moved function and changed to return querysets.

ietf/nomcom/tests.py
```diff
@@ -2775,6 +2776,24 @@ def test_elig_by_meetings(self):
...
+        MeetingRegistrationFactory(meeting=m, person=person_c, reg_type='remote')
+        MeetingRegistrationFactory(meeting=m, person=person_d, reg_type='remote')
+        AttendedFactory(session__meeting=m, session__type_id='plenary', person=person_c)
+        checked_in, attended = participants_for_meeting(m)
+        self.assertTrue(person_a.pk in checked_in)
+        self.assertTrue(person_b.pk not in checked_in)
+        self.assertTrue(person_c.pk in attended)
+        self.assertTrue(person_d.pk not in attended)
+
+
```
rjsparks, line 2796: This should move to the meeting tests.

rjsparks (changes requested):
Sorry - one more housekeeping change please.

rjsparks (approved):

PR #6569, opened 2023-10-31
fix: preserve RelatedDocument references to subseries

rjsparks:


ietf/sync/rfceditor.py
```diff
@@ -327,7 +328,7 @@ def extract_doc_list(parentNode, tagName):
             log("node: %s" % node)
             raise
     for d in data:
-        k = "RFC%04d" % d[0]
+        k = "RFC%d" % d[0]
```
rjsparks, line 331: This bug has been in production for years.

jennifer-richards (approved):
Looks good. The mention of "ghosted" in the same comment as "31 Oct" is a nice flourish. 🎃

PR #6570, opened 2023-11-01
fix: Don't redirect user to the login page when logging in

### LINKED ISSUE #5876: Sign in link issues
### Describe the issue

The "Sign in" link in the top right-hand corner keeps appending content for the next query parameter.

To reproduce:
* Go to Datatracker as an unlogged user.
* Click on "Sign in" button on the top right-hand corner.
* Click the "Sign in" button in the top right-hand corner again.
* New URL is `https://datatracker.ietf.org/accounts/login/?next=/accounts/login/%3Fnext%3D/`
* If you click again, the URL will be `https://datatracker.ietf.org/accounts/login/?next=/accounts/login/%3Fnext%3D/accounts/login/%253Fnext%253D/`.

Because clicking on the "Sign in" button from the sign-in page changes the `next` query parameter to `/account/login/<stuff>`, the user will be presented with a login screen after successful login.

### Code of Conduct

pselkirk:
Fixes #5876.

### DIFF
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -409,20 +409,16 @@ def startswith(x, y):
     return str(x).startswith(y)
 
 
-@register.filter(name='removesuffix', is_safe=False)
-def removesuffix(value, suffix):
-    """Remove an exact-match suffix
+@register.filter(name='removepath', is_safe=False)
+def removepath(value, suffix):
+    """Remove an exact-match path
     
     The is_safe flag is False because indiscriminate use of this could result in non-safe output.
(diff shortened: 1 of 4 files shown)


jennifer-richards (commented):
I think this'll solve the bug, but I'm a little worried about the heavy-handedness of `str.replace()` for the task. I'd feel better if it were restricted to a prefix. Will the original `/accounts/logout` removal still work if it's rephrased as a prefix?

pselkirk:
I'm not overly worried about abuse potential, but since we're really just going from stripping a suffix to stripping a prefix, I went back to your style of filter.

jennifer-richards (approved):

Issue #5432, opened 2023-03-28
Slide approvals should be marked to history


kivinen:
### Description

When WG chair approves a -00 document there is entry added to history saying:

  2022-12-07	00 	Yoav Nir 	WG -00 approved

but when the slides for the meeting are approved by someone there is no such entry in the history:

https://datatracker.ietf.org/doc/slides-116-ipsecme-ikev2-link-maximum-atomic-packet-and-packet-too-big-notification-extension/history/

I.e., there is only entry added when the slides where uploaded, but no entry when and by whom they were approved.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

kivinen2:
I test this now and it seems I do get both new slides and approvals in the history.

![image](https://github.com/user-attachments/assets/2135aca3-5cf0-4627-bda5-941418964c7f)

kivinen2:
So I think this can be closed.

PR #6588, opened 2023-11-04
fix: Add history entry when approving the slides.

kivinen:
Change os.rename to shutils.move as the submissions and proceedings are on the separate filesystems on the docker image, and this same thing might happen in the real environment in the future.


kivinen:
This fixes #5432

jennifer-richards (approved):
This looks ok to me, though should there also be an event for disapproval?

kivinen:
Jennifer Richards writes:
> @jennifer-richards approved this pull request.
> 
> This looks ok to me, though should there also be an event for disapproval?

For the initial presentation there is no way to log anything, as the
state is only created AFTER the document as been approved once.

Hmm... when the drafts are updated and then someone disapproves them,
then there would be possibility for making comment.
-- 
***@***.***

PR #6591, opened 2023-11-04
fix: Add Bing and DuckDuckGo search engines

### LINKED ISSUE #6590: Change search engine button on doc page
### Description

We should add more than just google to the "search for this doc" page.
I think Bing and DuckDuckGo.

### Code of Conduct

richsalz:
To the draft page

Fixes #6590


richsalz:
The issue has changed, and this PR is no longer the right thing to do. Closing.

Issue #3911, opened 2022-04-29
BoF deadline in important dates should rather have a link to the datatracker boF request page then the ietf.org page


rjsparks:
### Discussed in https://github.com/ietf-tools/datatracker/discussions/3910

<div type='discussions-op-text'>

<sup>Originally posted by **mirjak** April 29, 2022</sup>
The link to further instructions to the ietf.org should rather be at the point of the BoF request page, or at least also there!</div>

ajeanmahoney:
The template for important date _adds_ text to what the important date's description field contains for several importanddatename ids. This seems really odd - the information should just be part of the importantdatename description itself. This needs review.

The fix for this particular ticket is to change the description of the field in the admin.

evyncke:
I agree, it seems that `ietf/templates/meeting/important-dates.html` should be cleaned by removing most of of the specifics as they are not used.

evyncke:
OTOH (discovering DT piecemeal by piecemeal...), the template can use nice hyperlink with a descriptive text while any URL in the DB is simply transformed into a hyperlink (a little less nice).

rjsparks:
I think the right thing to do is to change the processing for the description of each important date to be markdown input, rendered through markdown at display.

All of the workarounds we're seeing in the template can then move into the description. @larseggert - do you have an opinion?

evyncke:
Nicer and more elegant solution indeed. BTW, is there a MD rendering library somewhere ?

rjsparks:
We are already using https://pypi.org/project/Markdown/.
Look at, for instance, https://github.com/ietf-tools/datatracker/blob/1ba87890ba9451bb696aef4c8d049bcefa85b71b/ietf/doc/views_doc.py#L560

PR #6594, opened 2023-11-04
Fix: meeting important dates can have markdown links

evyncke:
To address feat #3911 

Added a filter for the Django templates: 'markdown'

Data migration includes changing the important dates description to include the right markdown links. 

Still to do (another issue to be created?) as some links in important dates include the actual meeting number :-(


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -228,6 +229,26 @@ def link_other_doc_match(match):
...
+        python_markdown.markdown(
+            string,
+            extensions=[
+                "extra",
+                "nl2br",
+                "sane_lists",
+                "toc",
+            ],
+        )
+    )
```
larseggert, line 247: Why don't you call the code in ietf/utils/markdown.py?

evyncke: Because of a loop between the import statement (and no clue how to break this loop)

evyncke: In the same vein, I am unsure whether "toc" & "nl2br" extensions are useful or dangerous

rjsparks:
I pushed a commit that moves the new filter to avoid the circular import (and simplifies it to use utils.markdown).

Tests don't pass, but it's not this PRs fault.

ietf/doc/templatetags/ietf_filters.py
larseggert: https://stackoverflow.com/a/7199514/2240756

evyncke:
> I pushed a commit that moves the new filter to avoid the circular import (and simplifies it to use utils.markdown).

thanks, nicer indeed

rjsparks:
See the failed test run - the changes are producing html that doesn't validate.

For example
```
                                 <td>
                                    Cut-off date for revised BOF proposal requests to Area Directors at UTC 23:59. To request a BOF, please see instructions at <a href="https://www.ietf.org/how/bofs/bof-procedures"></a><a href="https://www.ietf.org/how/bofs/bof-procedures&lt;/a>">https://www.ietf.org/how/bofs/bof-procedures</a> on Requesting a BOF.
                                                                                                                                                                 ^
                                                                                                                                                                 wcag/h30: Anchor link must have a text describing its purpose on line 721:162
```

jennifer-richards:
> See the failed test run - the changes are producing html that doesn't validate.
> 

This was happening because the `linkify` filter was confusing the `markdown` filter. Reversing these avoids the problem and I think even does the right thing.

evyncke:
> 
> This was happening because the `linkify` filter was confusing the `markdown` filter. Reversing these avoids the problem and I think even does the right thing.

Indeed ! Thanks

PR #6600, opened 2023-11-04
fix: display document expiration notification in UTC

### LINKED ISSUE #1825: Show explicit expiration time and timezone in draft-expiry reminder emails
`keyword_sprint` `type_defect`   |    by lloyd.wood@yahoo.co.uk

___

(had a lot of trouble creating a login and getting ticket privileges to create this bug report. Never mind finding the bug link at datatracker - bottom of page.)

email reminders say when a draft expires, but not that you can't submit? 

From Ryan Cross via RT <ietf-info@ietf.org> 

To: Wood L Dr (Elec Electronic Eng); 

Thu 22/10/2015 4:45 AM

Hello Lloyd,
 
Thank you for your submission. This is an issue the IETF Tools Team should
 address. Please use the "Report a Bug" link at the bottom of this page:
 http://datatracker.ietf.org/ to raise this with the tools team. This way you
 will be notified of changes in status of the issue. If you are unable to do
 enter the bug for any reason let me know and I can enter it on your behalf.
 
Thank You,
 Ryan Cross
 AMS
 

On Mon Oct 19 22:58:53 2015, L.Wood@surrey.ac.uk wrote:
 > Well, this is mildly irritating.
 >
 > Had a bunch of warnings about drafts expiring:
 >
 > ***
 > IETF Secretariat <ietf-secretariat-reply@ietf.org>
 > To draft-wood-tsvwg-saratoga@ietf.org Oct 19 at 10:42 PM
 > The following draft will expire soon:
 >
 > Name: draft-wood-tsvwg-saratoga
 > Title: Saratoga: A Scalable Data Transfer Protocol
 > State: I-D Exists
 > Expires: 2015-10-20 (in 19 hours, 17 minutes)
 > ***
 >
 > Drafts were expiring sometime on the 20th of October (in, one
 > presumes, the US, because no timezone was given, and the IETF is
 > very much a US organisation. I live in Australia, no timezone for
 > expiry time given on these emails.)
 >
 > So I refreshed the draft ready to do the necessary six-month
 > accounting. Only to discover that I-D submission had already closed
 > at midnight UTC on the 19th because of the upcoming IETF meeting.
 > Which wasn't mentioned in the email.
 >
 > Why not state the earlier submission deadline as well as a hard limit?
 > Why aren't timezones stated?
 >
 > Bravo for https://datatracker.ietf.org/submit/ stating timezones, if
 > not consistently.
 >
 > Please file lack of timezones and lack of consideration of early
 > submission deadlines as a bug.
 >
 > thanks
 > Lloyd Wood
 > http://about.me/lloydwood
 

___
_Issue migrated from trac:1825 at 2022-03-04 04:47:37 +0000_

---

meadmaker:
The document expiration mail displays the date of the upcoming document expiration, but different timezones makes that date ambituous.  Instead, clearly show that as a timestamp in UTC.

Fixes https://github.com/ietf-tools/datatracker/issues/1825


rjsparks:
@meadmaker - please start using this for your commit messages: https://www.conventionalcommits.org/en/v1.0.0/

ietf/doc/expire.py
```diff
@@ -86,10 +87,15 @@ def send_expire_warning_for_draft(doc):
         (doc.get_state_slug("draft") != "active")):
         return # don't warn about dead or inactive documents
 
-    expiration = doc.expires.astimezone(DEADLINE_TZINFO).date()
+    expiration = doc.expires.astimezone(
+        DEADLINE_TZINFO
+    ).replace(
+        hour=0, minute=0, second=0, microsecond=0
+    ).astimezone(
```
jennifer-richards, line 94: Don't actually need to convert the timezone to UTC - the `|utc` in the template will do the right thing (also, just as a non-obvious fyi, we're trying to use `datetime.timezone.utc` as the UTC zone rather than `ZoneInfo("UTC")` or any of the various other synonyms)

jennifer-richards (changes requested):
Requesting a minor code cleanup as noted inline.

I also want to point out to @rjsparks that this will change the deadline from a tz-naive date to the UTC datetime describing midnight Pacific time. That was what I suggested as my best guess as the way to do things.

Thinking about it now, I think it'll be midnight at the _start_ of the `doc.expires` date, which might be incorrect. Maybe it needs to be `+datetime.timedelta(days=1)`.

rjsparks:
@larseggert - Can we by fiat declare draft expiration happens at UTC midnight, or do we need more process first?

larseggert:
I think IESG can declare

rjsparks:
Anywhere on earth essentially means "international date line" instead of "Greenwich Meridian". It is easy to do, but it runs counter to the previous expressed goal to do everything in UTC.

meadmaker:
My takeaways here are:
- Remove the extraneous timezone conversion, as identified by @jennifer-richards 
- Make the timestamp accurate
  - [Earlier in the file](https://github.com/ietf-tools/datatracker/blob/5dfbaabf5eb601f7a526e8837dc9f45bbbe08e2d/ietf/doc/expire.py#L64), the logic includes the "+1 day" calculation that @jennifer-richards suggested.  I'll adopt that one.

I'm going to go work on that, and submit it to this issue.  I'm going to look for opportunities to make the expiration of documents more explicit, though - we've already found at least six reasonable interpretations for when that could be (beginning of date and end of date, crossed with PDT, UTC, and International Dateline).  I wonder whether it would be better to convert the `date` into a `datestamp`, in order to make that explicit.

meadmaker:
> I wonder whether it would be better to convert the `date` into a `datestamp`, in order to make that explicit.

I should have said `timestamp` instead of `datestamp`.  Oops.

I just examined the `doc_document` table, and the `expires` column is already a `timestamp` data type.  It looks like that is set in `ietf/submit/utils.py` for regular submission and `ietf/doc/views_draft.py` for document resurrection by the Secretariat.  I expect that if I can direct those two code paths through a single function to compute the expiration timestamp correctly, then all the other calculations can be converted to simple value display instead.

rjsparks:
This lost attention during the transition period. Lets see if it can either be finished or turned back into a clarified issue.
cc: @jennifer-richards

jennifer-richards:
I believe the main hold-up is deciding what is the desired functionality.

Two parts:
1) what time of day is the expiration
2) where on Earth do you specify that time

I think 1) is "end of the expiration date" - at least, @cabo and I seem to agree.

For 2), UTC or "anywhere on earth" have been suggested.

The current implementation, if I understand it, will expire a document at the stroke of midnight UTC at the start of the expiration date. I think that leaves everyone unhappy, because we either want 12 hours later (for the end of day "anywhere on earth" standard, neglecting the Millenium Islands time zone) or 24 hours later (for end of day UTC)

rjsparks:
I agree with 1 and 2. We should build a reusable component to do that calculation. Is this PR a sufficient basis to work with to shift to that point, or should we approach the codebase again more generally?

jennifer-richards:
I'd say this is a good place to start.

PR #6605, opened 2023-11-05
feat: Capture volunteers from the registration system

rpcross:


ietf/api/views.py
```diff
@@ -202,6 +203,25 @@ def err(code, text):
             else:
                 send_account_creation_email(request, email)
                 response += ", Email sent"
+
+            # handle nomcom volunteer
+            if data['is_nomcom_volunteer'] and object.person:
+                now = datetime.datetime.now()
+                month = now.month
+                year = now.year
+                if month > 7:
```
richsalz, line 212: A comment explaining this if statement would be helpful.

rpcross: Added comment

ietf/api/views.py
```diff
@@ -202,6 +203,27 @@ def err(code, text):
             else:
                 send_account_creation_email(request, email)
                 response += ", Email sent"
+
+            # handle nomcom volunteer
+            if data['is_nomcom_volunteer'] and object.person:
+                # registrants for the fall meeting will be volunteering for the
+                # following year's nomcom
```
rjsparks, line 210: That's true for the summer meeting too (the nomcom selection process is run by then usually).

Instead I think we should discuss when the registration system should offer volunteering for nomcom - it may only want to do so only for the spring meeting. Or only if there is a nomcom configured that is accepting volunteers (and we can make the assumption that there would never be two nomcoms accepting volunteers at the same time.).

rjsparks (dismissed):

jennifer-richards (changes requested):
See inline - I want to confirm that it's ok to have bogus timestamps on existing `Volunteer` records and that the "time" field changing if the record is updated is what the field is meant to represent.

ietf/nomcom/models.py
```diff
@@ -327,7 +327,10 @@ class Volunteer(models.Model):
     nomcom = ForeignKey('NomCom')
     person = ForeignKey(Person)
     affiliation = models.CharField(blank=True, max_length=255)
-
+    time = models.DateTimeField(auto_now=True)
```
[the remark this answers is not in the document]
rjsparks: Good catch. auto_add_now is better once the train is running. For the initial data load, we _don't know_ what the time was, and we either reflect reality and allow None here, or _lie_ and say all the volunteer record sprang into existence the day the relevant nomcom was selected. I'd prefer that we harden any code we build to dealing with None times.

jennifer-richards:
Tests failing - migrations need an update

jennifer-richards (dismissed):

jennifer-richards (approved):

rjsparks (approved):

PR #6626, opened 2023-11-09
Install lsb-release to avoid docker container build failure

DavidSchinazi:
The Docker repository wasn't building for me, but this fix resolved the issue.


DavidSchinazi:
Oh looks like this was already fixed separately, closing

PR #6629, opened 2023-11-09
chore: Rip out reStructured text support, too

larseggert:


rjsparks:
closing this as asleep: we can reopen it if needed.
There are a lot of artifacts that are currently in rst that would need to be translated before we could drop rst support entirely.

PR #6632, opened 2023-11-10
chore: Replace markup_txt with rfc2html

larseggert:
This is a subset of #6629 that leaves the reStructured text stuff alone.


rjsparks:
This is a pretty major change, not just a library substitution for similar effect.

It essentially pulls a differently stylized htmlized view onto the main document page, replacing the plaintext view, getting use to even _more_ confusion about where people should look.

We need to push for less, and that's a bigger conversation than what we should have in a PR.

larseggert:
So I would be all in favor of completely removing the preview from this template, and force people to click the button for the rendering they want.

larseggert:
_Should_ I remove the preview?

rjsparks:
> We need to push for less, and that's a bigger conversation than what we should have in a PR.

I think we should have this discussion with the community (and really, this should be fodder for a more comprehensive UX-expert driven rework of the document display pages). Perhaps we should bring this up on tuesday's tools call.

rjsparks:
Based on the 12Dec tools call discussion, we should try to push this forward, but it's failing several tests.

rjsparks (changes requested):
Based on the 12Dec tools call discussion, we should try to push this forward, but it's failing several tests.

rjsparks:
It's now down to template validation failures - rfc2html may need tweaking?

larseggert:
I have some validation ignores queued up for next week.

docker/docker-compose.extend.yml
```diff
@@ -11,7 +11,7 @@ services:
             - /workspace/__pycache__
             - /workspace/.vite
             - /workspace/.yarn/unplugged
-            - app-assets:/assets
+            - /Users/lars/Documents/IETF/assets:/assets
```
rjsparks: Well, no. But I understand the desire to have some sort of local persistent store to use rather than suffering the resync every time.

larseggert: Sorry, that was an accidental inclusion.

rjsparks:
drafts for ads is still failing validation

larseggert:
Hm. It passed after rebasing.

rjsparks (approved):

jennifer-richards:
@larseggert - the cross-reference URLs generated by this were broken. At `/doc/draft-ietf-stir-certificates-ocsp/#page-11` (for example), the link for `[RFC5280]` is https://dt-main.dev.ietf.org/doc/draft-ietf-stir-certificates-ocsp/rfc5280 rather than https://dt-main.dev.ietf.org/doc/rfc5280

This will be reverted by #6909. Because we squash-merged this, you should be able to rebase this branch onto main after the revert PR is merged.

jennifer-richards:
I rebased the merged branch here: https://github.com/jennifer-richards/datatracker/tree/chore-remove-markup-txt-light

(You probably should double-check that I did that correctly as there were a few conflicts. Note that it does not include commits that did not leave any changes, so the commit count doesn't match this PR's 14 commits.)

PR #6646, opened 2023-11-16
fix: restore ancestors and decendants on document graph 

rjsparks:


ietf/doc/utils.py
```diff
@@ -957,49 +957,62 @@ def make_rev_history(doc):
...
-                    }
-                    if d.history_set.filter(rev=e.newrevisiondocevent.rev).exists():
-                        history[url]['pages'] = d.history_set.filter(rev=e.newrevisiondocevent.rev).first().pages
+            if d.type_id == "rfc":
+                url = urlreverse("ietf.doc.views_doc.document_main", kwargs=dict(name=d))
+                e = d.docevent_set.filter(type="published_rfc").order_by("-time").first()
+                history[url] = {
+                    "name": d.name,
+                    "rev": d.name,
+                    "published": e and e.time.isoformat(),
```
jennifer-richards, line 1001: Did you check whether the consumer of this data can handle a `null` in the JSON here?

ietf/doc/utils.py
```diff
@@ -957,49 +957,62 @@ def make_rev_history(doc):
...
+                    if hasattr(e, 'newrevisiondocevent'):
+                        url = urlreverse("ietf.doc.views_doc.document_main", kwargs=dict(name=d)) + e.newrevisiondocevent.rev + "/"
+                        history[url] = {
+                            'name': d.name,
+                            'rev': e.newrevisiondocevent.rev,
+                            'published': e.time.isoformat(),
+                            'url': url,
+                        }
+                        if d.history_set.filter(rev=e.newrevisiondocevent.rev).exists():
+                            history[url]['pages'] = d.history_set.filter(rev=e.newrevisiondocevent.rev).first().pages
```
jennifer-richards, line 1015: Should there be an analogous lookup of pages for the RFC case?

jennifer-richards (approved):
This looks fine to me - a couple small questions inline

ietf/doc/utils.py
rjsparks: I did not, but it would mean we have an RFC without a published event, which is an anomaly and the consequence would be that the graph doesn't display.

ietf/doc/utils.py
rjsparks: good question - I'm not sure where it's used. It isn't used for this particular graph.

jennifer-richards: I don't see that we use it, though we do describe using this endpoint as part of the API so outsiders might be using it.

PR #6686, opened 2023-11-22
fix: Remove Google from doc search action

richsalz:
Revised fix.

Fixes: #6590


jennifer-richards (approved):

rjsparks:
No, they are dupes, it happened as a side-effect of a question I asked. I'll close this one in favor of the other, but as I'll note there, I think we should address the bigger issue that Rich calls out before (or instead of merging even that one.

PR #6702, opened 2023-11-28
fix: Reject obvious bad encoding pastes into the Submitter field in submissions

### LINKED ISSUE #6455: Reject obvious bad encoding pastes into the Submitter field in submissions.
At the moment the Person record associated with cs.ru.nl@viguier.nl has a name of `=?utf-8?q?Beno=C3=AEt_Viguier?=`

The Person record (again at the moment) has no User - the information came only and exactly from the draft submission, and there it came from the `submitter` field. The extracted author names look fine.

improve the form validation to look for pastes with this encoding and reject them.

pselkirk:
Fixes #6455

### DIFF
--- a/ietf/submit/forms.py
+++ b/ietf/submit/forms.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2011-2022, All Rights Reserved
+# Copyright The IETF Trust 2011-2023, All Rights Reserved
 # -*- coding: utf-8 -*-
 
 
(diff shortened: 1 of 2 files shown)


ietf/submit/forms.py
```diff
@@ -758,6 +758,20 @@ def cleaned_line(self):
...
 
+    def clean_name(self):
+        name = super(SubmitterForm, self).clean_name()
+        if name.startswith('=?'):
+            msg = f'"{name}" appears to be a MIME-encoded string.'
+            try:
+                import email.header
+                text, encoding = email.header.decode_header(name)[0]
+                decoded_name = text.decode(encoding)
+                msg += f' Did you mean "{decoded_name}"?'
```
jennifer-richards, line 769: Do you know whether the user pasting in the MIME-encoded string sees their original with the `=?` characters, or does their browser render it as decoded? (If it's the latter, I'm wondering if their experience will be a field containing "Jane Doe" with a warning "Did you mean 'Jane Doe'?" leading to confusion)

jennifer-richards (commented):
See question inline

ietf/submit/forms.py
pselkirk: It appears that the literal string "=?utf-8?q?Beno=C3=AEt_Viguier?=" was pasted into the submitter field. In any case, that's what I'm checking for, and the error string would look like `"=?utf-8?q?Jane_Doe?=" appears to be a MIME-encoded string. Did you mean "Jane Doe"?`

jennifer-richards (approved):

PR #6703, opened 2023-11-28
fix: Show requested rev if it is current, even w/o dochistory

### LINKED ISSUE #6699: feat/rfc - older rfcs with incomplete/incorrect DocHistory won't show the last draft before the rfc is published where main does
https://datatracker.ietf.org/doc/draft-klensin-rfc2821bis/11/ shows a draft

https://dt-rfc.dev.ietf.org/doc/draft-klensin-rfc2821bis/11/
redirects to RFC5321

The existing datatracker behavior should be preserved until improvements related to #6059 and the tree of issues mentioned there are made.

jennifer-richards:
This fixes #6699 by restoring the old datatracker behavior when the viewing the latest rev of a draft that has no `DocHistory`.

### DIFF
--- a/ietf/doc/views_doc.py
+++ b/ietf/doc/views_doc.py
@@ -210,8 +210,8 @@ def document_main(request, name, rev=None, document_html=False):
     snapshot = False
 
     gh = None
-    if rev:
-        # find the entry in the history
+    if rev and rev != doc.rev:
+        # find the entry in the history if the rev requested is not the current rev
         for h in doc.history_set.order_by("-time"):
             if rev == h.rev:
                 snapshot = True


jennifer-richards:
Note that for the more common case, where we _do_ have `DocHistory`, this change means we'll be using the `Document` instead of a `DocHistory` when explicitly requesting the latest rev. I think this is somewhere between ok and better, but we could fix the immediate bug with a lighter touch if that change is worrisome.

rjsparks (approved):

rjsparks:
I agree that using the Document in this case is actually better.

PR #6759, opened 2023-12-11
feat!: RFCs and subseries as Document types

rjsparks:


rjsparks:
obviated by a push to main.

PR #6773, opened 2023-12-13
fix: Update group dependencies for new rfc model

### LINKED ISSUE #6768: Group dependencies view (deps.json) causes 500 error
### Describe the issue

After deploying v12.0.0, visiting `/wg/<acronym>/deps.json` results in a 500 error. The error is

```
'NoneType' object has no attribute 'slug'
```
at line 775 of `ietf/group/views.py` (in the `dependencies()` view)

### Code of Conduct

jennifer-richards:
This eliminates the attempts to get `state.slug` when `state is None`, which superficially fixes the error in ticket #6768.

PR is put up as a draft because I think more attention may be needed - some of the logic identifies RFCs by using the `draft` state. I think that should look at the type instead but I'm still sorting out what's meant to be going on.

Fixes #6768

### DIFF
--- a/ietf/group/views.py
+++ b/ietf/group/views.py
@@ -742,11 +742,17 @@ def dependencies(request, acronym, group_type=None):
     )
     rfc_or_subseries = {"rfc", "bcp", "fyi", "std"}
     both_rfcs = Q(source__type_id="rfc", target__type_id__in=rfc_or_subseries)
-    pre_rfc_draft_to_rfc = Q(source__type_id="draft", source__states__slug="rfc", target__type_id__in=rfc_or_subseries)
-    rfc_to_pre_rfc_draft = Q(source__type_id__in=rfc_or_subseries, target__type_id="draft", target__states__slug="rfc")
+    pre_rfc_draft_to_rfc = Q(
+        source__states__type="draft",
+        source__states__slug="rfc",
+        target__type_id__in=rfc_or_subseries,
+    )
     both_pre_rfcs = Q(
-        source__type_id="draft", source__states__slug="rfc",
-        target__type_id="draft", target__states__slug="rfc",
+        source__states__type="draft",
+        source__states__slug="rfc",
+        target__type_id="draft",
+        target__states__type="draft",
+        target__states__slug="rfc",
     )
     inactive = Q(source__states__slug__in=["expired", "repl"])
     attractor = Q(target__name__in=["rfc5000", "rfc5741"])
@@ -755,7 +761,6 @@ def dependencies(request, acronym, group_type=None):
         RelatedDocument.objects.filter(references)
         .exclude(both_rfcs)
         .exclude(pre_rfc_draft_to_rfc)
-        .exclude(rfc_to_pre_rfc_draft)
         .exclude(both_pre_rfcs)
         .exclude(inactive)
         .exclude(attractor)
--- a/ietf/group/views.py
+++ b/ietf/group/views.py
@@ -754,9 +754,12 @@ def dependencies(request, acronym, group_type=None):
         target__states__type="draft",
         target__states__slug="rfc",
     )
-    inactive = Q(source__states__slug__in=["expired", "repl"])
+    inactive = Q(
+        source__states__type="draft",
+        source__states__slug__in=["expired", "repl"],
+    )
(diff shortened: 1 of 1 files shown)


ietf/group/views.py
```diff
@@ -737,26 +737,35 @@ def dependencies(request, acronym, group_type=None):
 
     references = Q(
         Q(source__group=group) | Q(source__in=cl_docs),
-        source__type="draft",
+        source__type="draft",  # does this need to be type__in=["draft", "rfc"]? Check for informative ref rfc->draft
```
rjsparks, line 740: I think likely not, but it _might_ be a change from previous behavior without it.

Since this is only used for groups, we don't have to worry about rfcs that didn't come from drafts here.

The only use such data might have to the consumer of this function is knowing that some RFC inf-reffed this group's active draft. But that's a done deal, so the group won't really be motivated to _do_ anything with the draft given this knowledge. So it's probably noise and shouldn't be on the graph in the first place.

ietf/group/views.py
```diff
@@ -737,26 +737,35 @@ def dependencies(request, acronym, group_type=None):
...
-
-    both_rfcs = Q(source__type_id="rfc", target__type_id="rfc")
+    rfc_or_subseries = {"rfc", "bcp", "fyi", "std"}
+    both_rfcs = Q(source__type_id="rfc", target__type_id__in=rfc_or_subseries)
+    pre_rfc_draft_to_rfc = Q(source__type_id="draft", source__states__slug="rfc", target__type_id__in=rfc_or_subseries)
+    rfc_to_pre_rfc_draft = Q(source__type_id__in=rfc_or_subseries, target__type_id="draft", target__states__slug="rfc")
+    both_pre_rfcs = Q(
+        source__type_id="draft", source__states__slug="rfc",
+        target__type_id="draft", target__states__slug="rfc",
+    )
```
rjsparks, line 750: it's probably worth adding that you're looking for `states__type="draft"` to all of the places you are looking for `states__slug=="rfc"` in case some other state machine ever gets an `rfc` node.

ietf/group/views.py
```diff
@@ -771,13 +780,12 @@ def dependencies(request, acronym, group_type=None):
...
         "nodes": [
             {
-                "id": x.name,
-                "rfc": x.get_state("draft").slug == "rfc",
-                "post-wg": not x.get_state("draft-iesg").slug
-                in ["idexists", "watching", "dead"],
-                "expired": x.get_state("draft").slug == "expired",
-                "replaced": x.get_state("draft").slug == "repl",
-                "group": x.group.acronym if x.group.acronym != "none" else "",
+                "id": x.became_rfc().name if x.became_rfc() else x.name,
```
rjsparks, line 783: Check this on a busy group and if the time is noticeable, I think we can get the db to grab this with the primary query instead of n+1 -ing it here.

rjsparks: We might do that in a way the puts the conditional into the query as well, so that we don't see this pattern repeating.

ietf/group/views.py
jennifer-richards: I looked at a handful of cases comparing against the 11.16.0 release (i.e., for posterity, the release before the one with the "new" rfc modeling) and did not see any where nodes were missing. That included some selected specifically because they had a `refinfo` relationship pointing from an RFC to a group draft.

Also, I think the only way this could miss a relationship is if there was a relationship from an RFC to a group draft without there also having been a relationship from the draft that _became_ the RFC to the group draft. Does that happen (aside from April 1 RFCs, which I feel pretty comfortable leaving off these plots except perhaps for amusement value)?

ietf/group/views.py
jennifer-richards: On the idr group (see below to gauge whether this is "busy", but it looks it to me), this brute force approach takes the request from 2-2.4 seconds up to 2.2-2.5 seconds on my dev machine. The code here is taking advantage of the caching in `became_rfc()`, so I believe it's at least only paying the N+1 penalty and not hammering the db any worse than that.

image

ietf/group/views.py
jennifer-richards: done

rjsparks (approved):

PR #6777, opened 2023-12-13
fix: consider subseries when calculating downref

### LINKED ISSUE #6774: BCP references mislabeled as "possible downref"
### Describe the issue

The references for RFC 9514 (and others - no particularly relevant reason to single this one out) includes two normative references to BCPs, both labeled as "Possible Downref". Prior to the feat/rfc (v12.0.0) release, these had no downref noted at all.

Wild guess is that the "Possible Downref" is noticing that the target of the ref is not of `type_id='rfc'`

### Code of Conduct

rjsparks:
fixes #6774


jennifer-richards (approved):
Lgtm

PR #6784, opened 2023-12-14
fix: Include refs to pre-RFC drafts as refs to the RFC

jennifer-richards:
On the person profile page, fixes ref counts to agree with what was showed pre-RFC model refactoring.

On the doc referencedBy page:
* includes references to the draft that became the RFC
* fixes the "As draft-..." badge which was broken during the feat/rfc work due to a confusing shadowed variable name


ietf/doc/views_doc.py
```diff
@@ -1439,6 +1439,10 @@ def document_references(request, name):
 def document_referenced_by(request, name):
     doc = get_object_or_404(Document,name=name)
     refs = doc.referenced_by()
+    if doc.came_from_draft():
+        refs |= doc.came_from_draft().referenced_by()
+    if doc.became_rfc():
```
rjsparks, line 1444: Still not sure about this.
If you have a `draft` Document, and you want to know what referenced _it_, reporting what referenced the RFC it might have become doesn't seem to make sense.

jennifer-richards: My motivation to add this was to preserve behavior from before feat/rfc came along. I think it's the right(ish) thing to do.

A reference that comes in this way shows up with a badge saying "As RFCxyz" so it's annotated as being to a related entity rather than the draft itself. Including it this way lets someone looking at the draft see who is using the work, even if it's not literally referred to, while still being clear about the caveat that it's an indirect reference.

rjsparks: Maybe we should add that motivation to the docstring?
Or rename the function?

jennifer-richards: Docstring makes sense. I think it's a reasonable name for this view (vs in the model methods, where someone might later use it for other purposes).

rjsparks (approved):

PR #6789, opened 2023-12-16
chore: Bump version to 12 in a couple places

jennifer-richards:
I think we missed these spots in the v12.0.0 release. I don't think we need to mark these as breaking, but note that the equivalent changes were part of a breaking change when v11.0.0 came out.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Get Next Version
+    - name: Get Next Version (Prod)
       if: ${{ github.ref_name == 'release' }}
       id: semver
       uses: ietf-tools/semver-action@v1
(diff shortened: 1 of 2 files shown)


rjsparks:
Let's talk about what to do with these things rather than just have an epoch token in them.

@NGPixel Does any software made decisions based on the content of these settings? 

We show them in some templates and it would be nicer to show something more useful to the viewer.

NGPixel:
They are used in the dev builds. Production builds are using the version generated from the conventional commits.

No "decision" is made based on them though.

jennifer-richards:
It's perhaps possible to extract the version from the tag using something like
```
dev ➜ /workspace $ git describe --tags --always $(git rev-parse head)
12.0.0-13-g7d447802d

dev ➜ /workspace $ git checkout 6083205cc8068234bc9a92b3ff66a8a5f36a9d74

dev ➜ /workspace $ git describe --tags $(git rev-parse head)
11.16.0-281-g6083205cc
```
(the output format is `<most recent tag>-<commit count since tag>-g<abbreviated SHA>`)

Could do this at run-time if the `__version__` is empty (or some placeholder value) so that local dev will say something sensible in the page footer. For sandbox deploys it could be done as part of the deploy action (if there's not a better way to grab the info)

NGPixel (dismissed):

jennifer-richards:
The commit just now uses git inside a dev container to extract the most recent release tag, then adds the branch name and commit hash in the same format as shows up on dev releases. E.g., on my system it looks like `Version 12.0.0-dev (v12-bump - cfea732)`. (It's "12.0.0-dev" because this PR is outdated - if I switch to current main, I get `Version 12.6.1-dev (...` instead.)

This should only affect work in dev containers, and I think will work under VS Code but I have not tried. Happy to adjust the format (maybe something that more clearly labels it as a non-release version number?) or add better error checking, but wanted to get feedback on the general approach first.

jennifer-richards:
Just realized I'm being silly about the branch / commit, cleanup incoming...

rjsparks (approved):

PR #6797, opened 2023-12-18
fix: Fix disclaimer logic

larseggert:
As discussed on Slack


rjsparks:
We _really_ need a test (or more) that verifies this logic. If some well-meaning seemingly unrelated change breaks this and the wrong disclaimer starts showing on a subsequent release, the damage would be significant.

rjsparks:
I'll merge this so we can see the result in the staging dev area, but I don't think we should merge the branch to main without a solid set of tests.

PR #6809, opened 2023-12-18
fix: don't create a volunteer object when a person has already volunt…

rjsparks:


ietf/api/views.py
```diff
@@ -209,7 +209,7 @@ def err(code, text):
                     nomcom = NomCom.objects.get(is_accepting_volunteers=True)
                 except (NomCom.DoesNotExist, NomCom.MultipleObjectsReturned):
                     nomcom = None
-                if nomcom:
+                if nomcom and not Volunteer.objects.filter(nomcom=nomcom,person=object.person).exists():
```
jennifer-richards, line 212: `Volunteer.objects.get_or_create()` instead?

jennifer-richards (dismissed):

jennifer-richards (approved):

PR #6821, opened 2023-12-20
fix: Only insert <wbr> when there isn't a space before or after

larseggert:
Fixes the issue @wkumari raised in email.

### DIFF
--- a/ietf/static/js/ietf.js
+++ b/ietf/static/js/ietf.js
@@ -57,7 +57,7 @@ $(document)
                 var text = $(this)
                     .text();
                 // insert some <wbr> at strategic places
-                var newtext = text.replace(/([@._+])/g, "$1<wbr>");
+                var newtext = text.replace(/\S([@._+])\S/g, "$1<wbr>");
                 if (newtext === text) {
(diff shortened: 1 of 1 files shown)


ietf/static/js/ietf.js
```diff
@@ -57,7 +57,7 @@ $(document)
                 var text = $(this)
                     .text();
                 // insert some <wbr> at strategic places
-                var newtext = text.replace(/([@._+])/g, "$1<wbr>");
+                var newtext = text.replace(/\S([@._+])\S/g, "$1<wbr>");
```
cabo, line 60:
```
"123.456".replace(/([@._+])/g, "$1<wbr>");
'123.<wbr>456'
"123.456".replace(/\S([@._+])\S/g, "$1<wbr>");
'12.<wbr>56'
```

larseggert: Doh. Thanks.

wkumari:
Oh, ta! 
Yeah, I would never have found that -- thanks.

rjsparks (approved):

PR #6835, opened 2023-12-21
fix: better archive search links for group documents

### LINKED ISSUE #6823: the "Mailing list discussion" link for rfcs is incorrect
The generated link into the mailarchive uses `&q=` where it should use `/q=`. The query should also look for both references to the rfc name and the name of the draft the rfc came from.

rjsparks:
fixes #6823

### DIFF
--- a/ietf/templates/doc/document_info.html
+++ b/ietf/templates/doc/document_info.html
@@ -431,7 +431,7 @@
                         {% endfor %}
                         {% if doc.group and doc.group.list_archive %}
                             {% if doc.group.list_archive|startswith:settings.MAILING_LIST_ARCHIVE_URL %}
-                                <a href="{{ doc.group.list_archive }}?q={{ doc.name }}">
+                                <a href="{{ doc.group.list_archive }}?q={{ doc.name }}{% if doc.came_from_draft %} OR %22{{ doc.came_from_draft.name }}%22{% endif%}">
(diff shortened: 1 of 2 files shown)


ietf/templates/doc/document_info.html
```diff
@@ -431,7 +431,7 @@
                         {% endfor %}
                         {% if doc.group and doc.group.list_archive %}
                             {% if doc.group.list_archive|startswith:settings.MAILING_LIST_ARCHIVE_URL %}
-                                <a href="{{ doc.group.list_archive }}?q={{ doc.name }}">
+                                <a href="{{ doc.group.list_archive }}?q={{ doc.name }}{% if doc.came_from_draft %} OR %22{{ doc.came_from_draft.name }}%22{% endif%}">
```
kesara, line 434: Why not URL encode spaces around `OR` as well?

rjsparks: Can you find an ecosystem where it's actually necessary? (escaping the double-quotes was mostly a quick way to construct a string that django wouldn't mangle on the way out).

kesara: I requested a change just to avoid confusion. It's not apparent that spaces around `OR` are important.

kesara (approved):

PR #6836, opened 2023-12-21
fix: show state help for the editorial stream

### LINKED ISSUE #6755: /doc/help/state/draft-stream-editorial/ fails with a 404
The page is looking for a state it cannot find.

rjsparks:
fixes #6755

### DIFF
--- a/ietf/doc/views_help.py
+++ b/ietf/doc/views_help.py
@@ -1,5 +1,7 @@
 # Copyright The IETF Trust 2013-2023, All Rights Reserved
 
+import debug # pyflakes: ignore
+
 from django.shortcuts import render, get_object_or_404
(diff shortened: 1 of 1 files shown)


ietf/doc/views_help.py
```diff
@@ -1,5 +1,7 @@
 # Copyright The IETF Trust 2013-2023, All Rights Reserved
 
+import debug # pyflakes: ignore
```
kesara, line 3: Did you forget to remove this?

rjsparks: No - it's convention to leave it once we add it to a module to assist with future development. That's why it has the the `ignore` pyflakes directive.

kesara (approved):

PR #6851, opened 2023-12-26
fix: sorting RFCs by date in personal profiles sorts

aj-stein-nist:


ietf/templates/person/profile.html
```diff
@@ -106,7 +106,7 @@ <h2 class="mt-5" id="rfcs-{{ forloop.counter }}">
                             <td>
                                 <a class="text-nowrap" href="{{ doc.get_absolute_url }}">RFC {{ doc.rfc_number }}</a>
                             </td>
-                            <td>{{ doc.pub_date|date:"b Y"|title }}</td>
+                            <td>{{ doc.pub_date|date:"Y-m-d"|title }}</td>
```
rjsparks, line 109: If you look at the first page of RFCs, they only show Month and Year for when they were published (except for Apr 1 RFCs which show the date. We do have a published date in the database, but the RFC Editor wants this page to reflect what the RFC editor website would show.

Rather than this particular fix, we may want to add a separate undisplayed sort key (I think we have this in other sortable tables) for the column.

rjsparks (changes requested):
Thank you for the PR!
As noted inline, I think you should take a slightly different approach.

ietf/templates/person/profile.html
rjsparks: @larseggert Can you point to an example where we've provided an alternate sort function or sort values?

aj-stein-nist: Sounds good. I will research this and try to get back at it on Friday. So just to be clear, if I switch to the approach (thanks so much for the context, I suspected my fix was too easy if you labelled it "medium" difficulty 😆), do I need frontend tests or does that not cover the Django server-side generated content? Just want to make my PR easy to review and meet your standards. I am a first-timer on ietf-tools.

Thanks again for the quick reply!

rjsparks: Very happy to have you start in on contributing!

Our front-end test framework for older views is not well, and will be replaced.. Please don't spend time trying to figure what's there out.  Thanks for considering tests, but for this one we will use human visual inspection.

Several people that could help on this one are on holiday breaks, and we might not get feedback from them until next week.

nguyenpham00 (commented):
# # A first-level heading
## A second-level heading
### A third-level heading

PR #6854, opened 2023-12-28
fix: Properly set AD as action holder when submitting to IESG for publication

### LINKED ISSUE #5227: Action Holder is empty when draft is in IESG State "Publication Requested"
### Describe the issue

First off, thank you so much for the Action Holder feature, it's amazing. I think I might have found a bug in it though. When a WG chair requests publication on a document, the Action Holder field appears but it starts off empty. Would it be possible to have it start filled in with the name from the "Responsible AD" field instead? I noticed a [draft](https://datatracker.ietf.org/doc/draft-ietf-babel-mac-relaxed/) stuck in this state and a few of us were confused as we didn't know who to ask.

### Code of Conduct

pselkirk:
Fixes #5227

The clear intent of `to_iesg` is that
a) the document AD should be the group AD, if not already set, and b) the document Action Holder should be the document AD; but there was an order-of-operation error,
such that the Action Holder remained empty.

### DIFF
--- a/ietf/doc/views_draft.py
+++ b/ietf/doc/views_draft.py
@@ -564,9 +564,7 @@ def to_iesg(request,name):
 
             events = []
             def doc_event(type, by, doc, desc):
-                e = DocEvent(type=type, by=by, doc=doc, rev=doc.rev, desc=desc)
-                e.save()
-                return e
+                return DocEvent.objects.create(type=type, by=by, doc=doc, rev=doc.rev, desc=desc)
 
             if doc.get_state_slug("draft-iesg") == "idexists":
                 events.append(doc_event("started_iesg_process", by, doc, f"Document is now in IESG state <b>{target_state['iesg'].name}</b>"))
--- a/ietf/doc/views_draft.py
+++ b/ietf/doc/views_draft.py
@@ -563,18 +563,18 @@ def to_iesg(request,name):
             by = request.user.person
 
             events = []
-            def append_event(type, desc):
+            def doc_event(type, by, doc, desc):
                 e = DocEvent(type=type, by=by, doc=doc, rev=doc.rev, desc=desc)
                 e.save()
-                events.append(e)
+                return e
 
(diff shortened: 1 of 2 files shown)


ietf/doc/views_draft.py
```diff
@@ -560,23 +560,22 @@ def to_iesg(request,name):
...
-
-            if doc.get_state_slug("draft-iesg") == "idexists":
-                e = DocEvent()
-                e.type = "started_iesg_process"
-                e.by = by
-                e.doc = doc
-                e.rev = doc.rev
-                e.desc = "Document is now in IESG state <b>%s</b>" % target_state['iesg'].name
+            def append_event(type, desc):
+                e = DocEvent(type=type, by=by, doc=doc, rev=doc.rev, desc=desc)
```
rjsparks, line 567: I worry a bit at taking values out of the parent scope this way. Lets either make all of these things arguments, or change this to a utility class where an instance gets its own copy of these things in its init.

rjsparks (changes requested):
Small adjustment requested.

ietf/doc/views_draft.py
```diff
@@ -560,22 +560,21 @@ def to_iesg(request,name):
...
-
             by = request.user.person
 
             events = []
-
-            changes = []
+            def doc_event(type, by, doc, desc):
+                e = DocEvent(type=type, by=by, doc=doc, rev=doc.rev, desc=desc)
+                e.save()
+                return e
```
rjsparks, line 569:
```suggestion
                return DocEvent.objects.create(type=type, by=by, doc=doc, rev=doc.rev, desc=desc)
```

rjsparks: Or 
`return doc.docevent_set.create(type=type, by=by, rev=doc.rev, desc=desc)`

rjsparks (changes requested):
One more nit:

rjsparks (approved):

PR #6899, opened 2024-01-08
fix: Fix ordering of draft documents in AD dashboard

### LINKED ISSUE #6829: AD's documents are not listed by life cycle on AD dashboard
### Describe the issue

On the AD Dashboard (e.g., https://datatracker.ietf.org/doc/ad/roman.danyliw), documents are clustered by their state (PubReq, AD Evaluation, etc).  The current order is:

(1) PubReq
(2) AD Eval
(3) IETF LC
(4) IESG Eval
(5) Waiting for AD Go Ahead
(6) Approved Announcement to be sent
(7) I-D Exists
(8) Expired I-D
(9) RFC Ed

In a previous version of the DT, the order of the categories reflected their ordering in the document processing pipeline.  The order should be (using the previous numbering)

(7) I-D Exists <-- swapped order
(1) PubReq
(2) AD Eval
(3) IETF LC
(5) Waiting for AD Go Ahead <-- swapped order
(4) IESG Eval
(6) Approved Announcement to be sent
(9) RFC Ed
(8) Expired I-D <-- swapped order
 

### Code of Conduct

pselkirk:
Fixes #6829

### DIFF
--- a/ietf/doc/utils_search.py
+++ b/ietf/doc/utils_search.py
@@ -303,7 +303,6 @@ AD_WORKLOAD = {
         # "defer",  # probably not a useful state to show, since it's rare
         "approved",
         "ann",
-        "expired",
     ],
     "rfc": [
         "rfcqueue",
--- a/ietf/doc/utils_search.py
+++ b/ietf/doc/utils_search.py
@@ -292,16 +292,18 @@ def prepare_document_table(request, docs, query=None, max_results=200):
 #
 AD_WORKLOAD = {
     "draft": [
+        "idexists",
         "pub-req",
         "ad-eval",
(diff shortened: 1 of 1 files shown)


pselkirk:
One take-away is that removing a state from `AD_WORKLOAD` doesn't ipso facto suppress it - it just moves it to the end of the section, due to how `sort_key` works. So we should explicitly list all the states we might see, and may want to rethink the commented-out "probably not a useful state to show" states.

Also, this doesn't address Roman's expectation that `rfcqueue` should precede `expired`, since `rfcqueue` was grouped with `rfc`.

rjsparks:
closing this as OBE

PR #6913, opened 2024-01-11
feat: Web and Internet Transport Area

rjsparks:


rjsparks:
closing this as abandoned - the idea make it easy to test that the schedule editor worked, but completing automation will require a bit of a refactor of the group utilities and the group edit view so that a management command (and preferably a migration) could make these kinds of changes and still preserve, .e.g.,  RoleHistory

PR #6926, opened 2024-01-16
refactor: Move cron jobs to celery tasks

jennifer-richards:
This creates a first batch of celery tasks to replace some of the jobs done by cron. Adds a simple management command to actually schedule the tasks, which is done with them disabled. This should be suitable for a first test deployment where we manually disable the cron task and enable its celery equivalent to see how things go.

The management command `periodic_tasks --create-default` will set up the schedules. Running it without any options will summarize the cron-like tasks. It also accepts `--enable` or `--disable` to enable or disable a task by its PK, although the admin interface is probably a nicer way to do this.


rjsparks (approved):

rjsparks:
cronch

PR #6936, opened 2024-01-18
refactor: use statics.ietf.org for all statics

rjsparks:


ietf/settings.py
```diff
@@ -169,8 +169,8 @@
     STATIC_URL = "/static/"
     STATIC_ROOT = os.path.abspath(BASE_DIR + "/../static/")
 else:
-    STATIC_URL = "https://www.ietf.org/lib/dt/%s/"%__version__
-    STATIC_ROOT = "/a/www/www6s/lib/dt/%s/"%__version__
+    STATIC_URL = "https://static.ietf.org/lib/%s/"%__version__
+    STATIC_ROOT = None
```
rjsparks, line 173: Check this carefully - I've searched through the uses at django/djano and I _think_ this is legitimate, but another careful look is warranted

jennifer-richards (dismissed):
This looks good to me. Assuming you'll be collecting static files in an environment where the `STATIC_ROOT is not None` it should be ok. (I don't think it's _necessary_ to clear it out, but it's a nice hint and I don't see that it will cause any problems.)

rjsparks:
@NGPixel - does the build environment already override STATIC_ROOT?

NGPixel:
> @NGPixel - does the build environment already override STATIC_ROOT?

You mean dev? I don't see it being overridden in https://github.com/ietf-tools/datatracker/blob/main/docker/configs/settings_local.py

rjsparks:
No - I meant build - as in "create a production release" - in what context are you running `collectstatic` to gather the statics to put at R2?

rjsparks:
Answer is at https://github.com/ietf-tools/datatracker/blob/0ba7eaf37e0858d4c882348140888e9df40417a7/dev/deploy/settings_local_collectstatics.py#L8

NGPixel (approved):

PR #6957, opened 2024-01-22
fix: Set authors when publishing an RFC

### LINKED ISSUE #6949: Authors field missing in RFC 9521
### Describe the issue

While reviewing https://datatracker.ietf.org/doc/rfc9521/, I found that Authors field missing, none of the authors displayed.

### Code of Conduct

jennifer-richards:
Fixes #6949 

Restores pre-v12 behavior- authors are assumed to be the same as the last rev of the draft.

Adds the `reset_rfc_authors` management command, which can be used to correct RFCs published from drafts. By default, this will not touch an RFC with at least one author, but a `--force` option will override this. Adds a `DocEvent` describing the change.

### DIFF
--- a/ietf/doc/management/commands/reset_rfc_authors.py
+++ b/ietf/doc/management/commands/reset_rfc_authors.py
@@ -23,16 +23,25 @@ class Command(BaseCommand):
             raise CommandError(
                 f"rfc{options['rfcnum']} does not exist in the Datatracker."
             )
-        n_authors = rfc.documentauthor_set.count()
-        if n_authors > 0:
+        orig_authors = rfc.documentauthor_set.all()
+        if orig_authors.exists():
             # Potentially dangerous, so refuse unless "--force" is specified
             if not options["force"]:
                 raise CommandError(
                     f"{rfc.name} already has authors. Not resetting. Use '--force' to reset anyway."
                 )
+            removed_auth_names = list(orig_authors.values_list("person__name", flat=True))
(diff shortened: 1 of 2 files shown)


rjsparks (approved):

ietf/doc/management/commands/reset_rfc_authors.py
```diff
@@ -0,0 +1,67 @@
...
+        parser.add_argument("rfcnum", type=int, help="RFC number to modify")
+        parser.add_argument(
+            "--force",
+            action="store_true",
+            help="reset even if RFC already has authors",
+        )
+
+    def handle(self, *args, **options):
+        try:
+            rfc = Document.objects.get(type="rfc", rfc_number=options["rfcnum"])
```
rodneymorgan225, line 21: -

ietf/sync/rfceditor.py
```diff
@@ -462,6 +462,14 @@ def update_docs_from_rfc_index(
...
             if draft:
                 doc.formal_languages.set(draft.formal_languages.all())
+                for author in draft.documentauthor_set.all():
+                    # Copy the author but point at the new doc. 
+                    # See https://docs.djangoproject.com/en/4.2/topics/db/queries/#copying-model-instances
+                    author.pk = None
+                    author.id = None
+                    author._state.adding = True
+                    author.document = doc
+                    author.save()
```
rodneymorgan225, line 472: #Hood

PR #7004, opened 2024-01-31
feat: Provide an API to generate draft email aliases

jennifer-richards:
This is a piece of the work to break up `Cron-runner` to decouple the mail services from the datatracker. This provides an API endpoint that returns JSON mapping alias name -> email address lists.

Authentication is token-based, pulling in some work from `feat/rpc-api` for continuity with the future.

Adds a script, `ietf/bin/draft-aliases-from-json.py`, that converts the output from the API call into the `draft-virtual` and `draft-aliases` files produced by the original `generate_draft_aliases` management command. This uses only the Python standard library, so can be broken out into a better place when we split mail and datatracker to separate hosts.

I'm expecting the script to be refactored and reused for group aliases using a similar scheme.


jennifer-richards:
I think we need to add the group alias refactoring as well before it makes sense to merge this. Closing.

PR #7038, opened 2024-02-08
fix: Raise 404 when multiple person found in /community/personal/

kesara:


ietf/community/views.py
```diff
@@ -54,7 +54,7 @@ def view_list(request, email_or_name=None):
     try:
         clist = lookup_community_list(request, email_or_name)
     except MultiplePersonError as err:
-        return HttpResponse(str(err), status=300)
+        raise Http404(str(err))
```
rjsparks, line 57: We should either 404 with no decoration or use some other response. I'm not quickly coming up with a use-case for telling someone that the string entered matches multiple people is helpful.

kesara: Yeah! 404 page doesn't show any decorations. I'll remove error strings.

jennifer-richards (commented):
Would it make sense to push the `MultiplePersonError` -> 404 logic down into `lookup_community_list()`? Is it used anywhere that a 404 isn't the right thing to do?

(I'm not actually a fan of helper directly 404ing, but that's a bigger set of changes...)

rjsparks:
> Would it make sense to push the `MultiplePersonError` -> 404 logic down into `lookup_community_list()`? Is it used anywhere that a 404 isn't the right thing to do?
> 
> (I'm not actually a fan of helper directly 404ing, but that's a bigger set of changes...)

Yes, I think this would make sense, and it would obviate the need for the MultiplePerssonError exception in the first place - the utility could simply raise the 404 instead.

rjsparks:
@kesara @jennifer-richards : Is this still something we wish to pursue?

kesara:
> > @kesara @jennifer-richards : Is this still something we wish to pursue?
> 
> I think so, though I don't recall in detail the circumstances leading to it. I guess it's replacing 500s with 404s?
> 
> [edit] or I guess 300s with 404s, based on the test changes

@jennifer-richards, This was discussed on Slack.
Then, @rjsparks suggested returning 404 such cases.

kesara:
OBE

PR #7042, opened 2024-02-11
feat: Links to chatlogs and session recordings on proceedings page

### LINKED ISSUE #6791: Add links to chat recordings and session recordings in proceedings page
### Description

>From Lixia Zhang:
>for example, on this page:
> https://datatracker.ietf.org/meeting/118/proceedings/
>
>one can find all meeting materials, except the chatroom recording.

Currently, there are no links to the chat log or the session recordings from the proceedings view.

### Code of Conduct

pselkirk:
Fixes #6791

### DIFF
--- a/ietf/meeting/models.py
+++ b/ietf/meeting/models.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2007-2022, All Rights Reserved
+# Copyright The IETF Trust 2007-2024, All Rights Reserved
 # -*- coding: utf-8 -*-
 
 
@@ -1069,6 +1069,9 @@ class Session(models.Model):
 
     def bluesheets(self):
         return list(self.get_material("bluesheets", only_one=False))
+
+    def chatlogs(self):
+        return list(self.get_material("chatlog", only_one=False))
 
     def slides(self):
         if not hasattr(self, "_slides_cache"):
--- a/ietf/meeting/models.py
+++ b/ietf/meeting/models.py
@@ -1270,38 +1270,26 @@ class Session(models.Model):
             return self.meeting.group_at_the_time(self.group_at_the_time().parent)
 
     def audio_stream_url(self):
-        if (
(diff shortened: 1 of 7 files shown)


ietf/templates/meeting/group_proceedings.html
```diff
@@ -64,6 +76,12 @@ <h1>{{ entry.group }}</h1>
               </a>
               <br>
           {% endfor %}
+            {% if entry.session.video_stream_url %}
+                <a href="https://www.meetecho.com/ietf{{ meeting.number }}/recordings#{{ entry.group.acronym.upper }}">
```
jennifer-richards, line 80: This probably needs a helper along the lines of `session_recording_url` that hands the template the appropriate URL if there is one. The appropriate URL will almost surely vary over time if it doesn't already.

rjsparks: This is using the same pattern that several other places in the codebase use. If it's getting it wrong for the URL varying over time, it's doing that in production already.

It _would_ be good, however, to have all of them use a single helper function so the meetecho url isn't repeated through the codebase. It would be even better to move the url pattern into settings.

ietf/meeting/models.py
```diff
@@ -1296,6 +1296,16 @@ def onsite_tool_url(self):
             return url.format(session=self)
         return None
 
+    def session_recording_url(self):
+        if (
+            self.meeting.type.slug == "ietf"
+            and self.has_onsite_tool
+            and (url := getattr(settings, "MEETECHO_SESSION_RECORDING_URL", ""))
```
rjsparks, line 1303: We've been avoiding the walrus to date to make it easier for newcomers to the codebase to contribute.
Not using it here would make the condition easier to read/debug anyhow

rjsparks (changes requested):
One more change request to make future maintenance smoother

rjsparks (approved):

PR #7069, opened 2024-02-16
fix: Expand and enforce Cancel Interim Meeting Request comments length to 512

### LINKED ISSUE #6998: Add warning when character limit is exceeded on Cancel Interim Meeting Request page
### Describe the issue

The comments field on the page to cancel an interim meeting (e.g. https://datatracker.ietf.org/meeting/interim/request/interim-2024-nfsv4-02/cancel) apparently has 255-character limit, and exceeding that makes the form return a 500 internal server error. 

It would be helpful if the form could warn that the character limit was being exceeded, or if the error message that is returned was more descriptive than just a generic "internal server error."

### Code of Conduct

pselkirk:
Fixes #6998

### DIFF
--- a/ietf/meeting/migrations/0005_alter_session_agenda_note.py
+++ b/ietf/meeting/migrations/0005_alter_session_agenda_note.py
@@ -1,0 +1,18 @@
+# Copyright The IETF Trust 2024, All Rights Reserved
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("meeting", "0004_session_chat_room"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="session",
+            name="agenda_note",
+            field=models.CharField(blank=True, max_length=512),
+        ),
+    ]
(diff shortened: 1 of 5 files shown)


ietf/meeting/migrations/0005_alter_session_agenda_note.py
```diff
@@ -0,0 +1,18 @@
...
+        ("meeting", "0004_session_chat_room"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="session",
+            name="agenda_note",
+            field=models.CharField(blank=True, max_length=512),
+        ),
+    ]
```
pselkirk, line 18: Note: #7003 also has a 0005 migration. One of these will have to be updated to 0006, depending on merge order.

ietf/templates/meeting/interim_meeting_cancellation_notice.txt
```diff
@@ -3,6 +3,6 @@ The {{ group.name }} ({{ group.acronym }}) {% if not meeting.city %}virtual {% e
 interim meeting for {{ meeting.date|date:"Y-m-d" }} from {{ start_time|time:"H:i" }} to {{ end_time|time:"H:i" }} {{ meeting.time_zone }}
 has been cancelled.
 
-{{ meeting.session_set.0.agenda_note }}
+{{ meeting.session_set.first.agenda_note }}
```
pselkirk, line 6: Note: I'm deeply uneasy about this change, since I have to assume it worked before, but in my test environment, generated email had the line: ```** No value found for &#x27;meeting.session_set.0.agenda_note&#x27; **'```

rjsparks: Why assume it worked before? There's no test that looks to see if the note actually made it into the email i think.
Please add one.

I verified that the cancellation that led to this issue probably had a agenda_note added, but the email that went out did not contain it.

rjsparks (approved):

PR #7077, opened 2024-02-20
fix: Revert "Email authors" link for an RFC to the draft that it came from

### LINKED ISSUE #6856: "Email authors" button provides invalid email for https://datatracker.ietf.org/doc/html/rfc4035
### Describe the issue

The "Email authors" gives the email address "rfc4035@ietf.org" which is not found. The automated email response is:

> Address not found
> Your message wasn't delivered to rfc4035@ietf.org because the address couldn't be found, or is unable to receive mail.
> The response from the remote server was:
> 550 5.1.1 <rfc4035@ietf.org>: Recipient address rejected: User unknown in virtual alias table 

Would it be possible to fix the email address in order to communicate with the authors? Thank you!

### Code of Conduct

pselkirk:
Fixes #6856

### DIFF
--- a/ietf/templates/doc/document_draft.html
+++ b/ietf/templates/doc/document_draft.html
@@ -611,9 +611,9 @@
         </table>
         <div class="buttonlist">
             <a class="btn btn-primary btn-sm"
-               href="mailto:{{ doc_name }}@ietf.org?subject=Mail%20regarding%20{{ doc.name }}">
-               <i class="bi bi-envelope">
-               </i>
+               href="mailto:{{ doc.name }}@ietf.org?subject=Mail%20regarding%20{{ doc.name }}">
+                <i class="bi bi-envelope">
+                </i>
                 Email authors
             </a>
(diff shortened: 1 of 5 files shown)


ietf/templates/doc/document_draft.html
```diff
@@ -611,9 +611,9 @@
         </table>
         <div class="buttonlist">
             <a class="btn btn-primary btn-sm"
-               href="mailto:{{ doc.name }}@ietf.org?subject=Mail%20regarding%20{{ doc.name }}">
-                <i class="bi bi-envelope">
-                </i>
+               href="mailto:{{ doc_name }}@ietf.org?subject=Mail%20regarding%20{{ doc.name }}">
+               <i class="bi bi-envelope">
+               </i>
```
rjsparks, line 616: This should be a full revert - there should be no changes to document_draft.html for this PR.

rjsparks (changes requested):
The tests aren't looking at the link generated for the button. Maybe they should - if we left the changes to document_draft.html in place, the button would have generated a bad URL I think.

rjsparks (approved):

PR #7117, opened 2024-02-28
refactor: Clarify which mail archive is being searched

### LINKED ISSUE #6993: mail archive search URL (sometimes?) wrong
### Describe the issue

While doing a SECDIR review, the URL for "search the mail archives" is this:
```
No results found ('No results: https://mailarchive.ietf.org/arch/export/mbox/?qdr=c&start_date=2023-08-02&email_list=secdir&q=subject%3A%28draft-ietf-lamps-rfc8398bis%29&as=1 -> No results found') (searched for ["draft-ietf-lamps-rfc8398bis"](https://mailarchive.ietf.org/arch/search/?qdr=c&start_date=2023-08-02&email_list=secdir&q=subject%3A%28draft-ietf-lamps-rfc8398bis%29&as=1), corresponding [export](https://mailarchive.ietf.org/arch/export/mbox/?qdr=c&start_date=2023-08-02&email_list=secdir&q=subject%3A%28draft-ietf-lamps-rfc8398bis%29&as=1)). You have to fill in link and content yourself.
```

It shouldn't be *secdir* it should be the WG where the document is, in this case SPASM.

It also wasn't clear to me that what I would get would be a TGZ file of mbox archives.

### Code of Conduct

pselkirk:
Fixes #6993

### DIFF
--- a/ietf/doc/views_review.py
+++ b/ietf/doc/views_review.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2016-2020, All Rights Reserved
+# Copyright The IETF Trust 2016-2024, All Rights Reserved
 # -*- coding: utf-8 -*-
 
 
@@ -955,6 +955,8 @@ def complete_review(request, name, assignment_id=None, acronym=None):
 
     mail_archive_query_urls = mailarch.construct_query_urls(doc, team)
 
+    team_list_email = team.list_email or "directorate"
+
     return render(request, 'doc/review/complete_review.html', {
         'doc': doc,
         'team': team,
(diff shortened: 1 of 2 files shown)


ietf/doc/views_review.py
```diff
@@ -955,6 +955,8 @@ def complete_review(request, name, assignment_id=None, acronym=None):
 
     mail_archive_query_urls = mailarch.construct_query_urls(doc, team)
 
+    team_list_email = team.list_email or "directorate"
```
rjsparks, line 958: This conditional isn't making sense to me. Isn't it the case that if team.list_email is not Truthy, that mail_archive_query_urls will be None? (see the first return in `construct_query_urls).

I also don't think you _need_ this extra variable - if mail_archive_query_urls exists, see if you need out of it in the view?

rjsparks (changes requested):
I'm not following part of the suggested change:

rjsparks (approved):

Issue #7104, opened 2024-02-26
User can be created with an invalid username, leading server errors after login


webknjaz:
### Describe the issue

I just registered @ datatracker.ietf.org. I tend to use sub-addressing when registering on different websites. This time, I registered with an address having a format of `twochars+~accounts/datatracker.ietf.org@domain.name.com`. I got a verification email, clicked on the link, filled out the form with the name and a password generated using Bitwarden (128 random chars with special symbols). This worked too.

I'm trying to log in, the browser sends my login creds to https://datatracker.ietf.org/accounts/login/?next=/ and gets a normal 302 Found in response, with `Location: /`. Following the redirect, it shows me this banner:
> #### Internal Server Error
> ---------------------------
> The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there was an error in a script used to generate the requested page. In case this was an overload problem or other temporary error, you may want to try again in a little while.
>
> A failure report with details about what happened has been sent to the server administrators. It would be helpful if you would create an issue providing additional information at [GitHub](https://github.com/ietf-tools/datatracker/issues/new/choose), too.

I can click on various links in the top menu but neither works — each of them shows this exact banner. When I clean the cookies, the content shows up on all the pages, but as soon as I log in again, it's broken the same way.

This makes me thing there's some crash/traceback that is happening post-login in the web app, after retrieving the user session and doing something with the data. The HTTP session itself probably works on the django framework level. Perhaps, it doesn't like my e-mail format (the most likely reason!).

P.S. The only website pages that seem to work are direct RFC pages like https://datatracker.ietf.org/doc/html/rfc6874

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

jennifer-richards:
I believe you are correct that it is related to the format of your username. Will take some investigation to see whether we can fix this to accept the username you used or whether we need to adjust the rules on username format. 

If it's acceptable to use a different username, that will probably be the fastest way to get online would be to change it (this will probably require assistance, I'll contact you directly to discuss).

jennifer-richards:
The issue is caused by the "/" in the email address used for the Django username.

Django's User model allows only alphanumerics and the `@.+-_` characters in usernames. The _easiest_ thing to do here is to disallow characters other than those in email addresses, but that will disallow many valid emails.

[...]

PR #7124, opened 2024-02-29
feat: validate usernames with Django's rules

jennifer-richards:
This addresses #7104 - but note that it adds significant constraints to the email addresses allowed for new user registration.

### DIFF
--- a/ietf/ietfauth/forms.py
+++ b/ietf/ietfauth/forms.py
@@ -23,8 +23,8 @@ from .widgets import PasswordStrengthInput, PasswordConfirmationInput
 
 validate_username_email = ASCIIUsernameValidator(
     message=(
-        "This value may contain only unaccented lowercase a-z "
-        "and uppercase A-Z letters, numbers, and @/./+/-/_ characters."
+        'This value may contain only unaccented lowercase letters (a-z), '
+        'digits (0-9), and the special characters "@", ".", "+", "-", and "_".'
     )
 )
 
--- a/ietf/ietfauth/forms.py
+++ b/ietf/ietfauth/forms.py
@@ -244,9 +244,18 @@ class ChangeUsernameForm(forms.Form):
         assert isinstance(user, User)
         super(ChangeUsernameForm, self).__init__(*args, **kwargs)
         self.user = user
-        emails = user.person.email_set.filter(active=True)
-        choices = [ (email.address, email.address) for email in emails ]
+        choices = [(email.address, email.address) for email in self._allowed_emails(user)]
         self.fields['username'] = forms.ChoiceField(choices=choices)
+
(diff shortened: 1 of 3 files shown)


ietf/ietfauth/forms.py
```diff
@@ -20,8 +21,21 @@
...
+
+
+class UsernameEmailField(forms.EmailField):
+    """Email that is suitable for a username"""
+    default_validators = forms.EmailField.default_validators + [validate_username_email]
+
+
 class RegistrationForm(forms.Form):
-    email = forms.EmailField(label="Your email (lowercase)")
+    email = UsernameEmailField(label="Your email (lowercase)")
```
rjsparks, line 38: should reconcile (lowercase) here with upper and lowercase in vallidate.

rjsparks: will the emailfield validators disallow `foo@bar@baz`? (particularly `foo@@baz`)

jennifer-richards: They implement the checks in [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#autoid-24) accurately, I believe. I'll double-check.

This change should be giving us all the validation we had before plus the standard Django username validation so it should leave us with a strict subset of the bugs we had before.

I'll update the message thrown by the new username validator and, if it doesn't already give a sensible error for non-lowercase input, I'll validate that as well.

jennifer-richards: Checked - it rejects double-@ (both arrangements you listed) both via in-browser validation and, if you bypass that, in Django's validation. It also already complains if you use upper-case letters so it's just the error message on the new validator that needs to change.

jennifer-richards: Updated the message and refactored the case validation so it also gets enforced when selecting a username from active addresses.

rjsparks:
@jennifer-richards is this still something to pursue?

jennifer-richards:
> @jennifer-richards is this still something to pursue?

I think so - the original bug is still present. IIRC the next thing to do is check whether there are existing Users who have these invalid usernames set.

rjsparks:
Maybe we do that check after removing Users with no Persons?

Michealinc (commented):
Explict

jennifer-richards:
> Maybe we do that check after removing Users with no Persons?

@rjsparks, this might be ripe now? After last month's personless-user purge, we still have zero. OTOH, with the move to an external OIDC provider we might want to leave this alone. It might be that we do away with email-based usernames entirely.

jennifer-richards:
Closing: issue does not come up often, changes to auth may obviate it entirely, and we can reopen if it proves needed.

Issue #7187, opened 2024-03-15
IETF datatracker: new "testing" tag for "additional resources"


bclaise:
### Description

Case 1: individual drafts
datatracker.ietf.org/doc/draft-ahuang-netconf-notif-yang/ as an example
I can EDIT the additional resources
If I click EDIT, I see the following tags: faq, github_org, github_repo, github_username, gitlab_repo, gitlab_username, jabber_log, jabber_room, mailing_list, mailing_list_archive, related_implementations, repo, slack, tracker, webpage, wiki, yc_entry, yc_impact, zulip
I have not tested with the live datatracker, but it seems that I can select among the different existing tags, and add a pointer.

case 2: a working group document, I am an author, and not the WG chair
I can't edit the "additional resources"

case 3: working document, and I am the WG chair
I can't test this one.

Conclusion:
It seems that we "just" need to add a new tag (such as "testing") and follow the same datatracker id authorization logic (the 3 use cases above)

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

evyncke:
Will try to address this one during Code Sprint

evyncke:
@bclaise would "related_test_cases" be OK ?

rjsparks:
What's being pointed to? Documentation of how to test? Code? Results? Something else?

evyncke:
It seems indeed to be just a value in a model/table (to be handled by the secretariat I think) to add 'related_test_cases'.

Based on oral discussion with @bclaise it is more about wiki page or any external repository.

evyncke:
It seems that adding a single row in name_extresourcename will do the trick indeed ;-)

rjsparks:
>it is more about wiki page or any external repository.

Then why not use the existing `wiki` or `repository` external names and give them descriptive text like "Related Test Cases"?

What value is having the extra _tag_ bringing?

bclaise:
> > it is more about wiki page or any external repository.
> 
> Then why not use the existing `wiki` or `repository` external names and give them descriptive text like "Related Test Cases"?
> 
> What value is having the extra _tag_ bringing?

The wiki or repository tags don't bring the necessary that this is testing code, with for example PCAP.

graf3net:
I suggest to add the following tags: related_test_cases, related_test_results
[...]

PR #7196, opened 2024-03-16
Fix: new "testing" tag for "additional resources"

evyncke:
Fixing #7187


rjsparks:
I'm going to close this one. It is well formed, but the issue discussion looks to be taking us in a different direction, and when it concludes, I can add the new descriptors directly with the admin

PR #7197, opened 2024-03-16
fix: Show recordings for interims

### LINKED ISSUE #6543: Youtube link is not shown in Notes and Recordings section of past meeting materials
See https://datatracker.ietf.org/meeting/interim-2023-tools-11/session/tools

There is an icon in the top right that links to the youtube recording.

There should also be a row for it in the Notes and Recordings section of that view.

richsalz:
List the recordings if the "meeting number" starts with "interim"

Fixes: #6543


richsalz:
Should the same fix be made for the buttons, at https://github.com/ietf-tools/datatracker/blob/main/ietf/templates/meeting/session_buttons_include.html#L139 ?

ietf/templates/meeting/session_details_panel.html
```diff
@@ -320,7 +320,7 @@ <h3 class="mt-4">Notes and recordings</h3>
                 </tr>
             {% endif %}
             {# Recordings #}
-            {% if meeting.number|add:"0" >= 80 %}
+            {% if meeting.number|add:"0" >= 80  or "interim" in meeting.number %}
```
rjsparks, line 323:
```suggestion
            {% if meeting.type_id == 'ietf' and meeting.number|add:"0" >= 80  or meeting.type_id != 'ietf' %}
```

rjsparks: The above suggestion is a bit awkward - it requires knowing the precedence of operators as documented at https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#boolean-operators. It might be better to move this into the view code to get away from the `|add:"0"` type-coersion-by-hack-trick anyhow.

richsalz: I thought of something like that.  In addition to the one mentioned above, the other [tests against meeting 60](https://github.com/ietf-tools/datatracker/blob/main/ietf/templates/meeting/session_buttons_include.html#L129).  What's the significance of meetings 60 and 80?  I'd like to remove the magic numbers, replacing with a reasonable function name.

rjsparks:
>Should the same fix be made for the buttons, at https://github.com/ietf-tools/datatracker/blob/main/ietf/templates/meeting/session_buttons_include.html#L139 ?

Looks like it.

This really needs a new (or updated) test.

ietf/templates/meeting/session_details_panel.html
rjsparks: they aren't magic (though they aren't complete) - they are places where the _shape_ of proceedings and what was captured at meetings fundamentally changes.  60 was when we first started capturing chatlogs (jabber at the time). 80 was a point at where recording tech changed significantly.

That said, we don't have any Documents of type "recording" before ietf-100, so the code being touched here could could (for meetings of type ietf) not show the recordings block at all for meeting numbers < 100.

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -881,3 +881,19 @@ def badgeify(blob):
             )
 
     return text
+
+@register.filter
+def has_chat_logs(meeting):
```
rjsparks, line 886: Should this get the same "Return true for interims" treatment?

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -881,3 +881,19 @@ def badgeify(blob):
...
+@register.filter
+def has_recordings(meeting):
+    if meeting.type_id != 'ietf':
+        return True
+    num = meeting.get_number()
+    return num != None and num >= 80
+
+#register.filter
+#define has_notes(meeting):
+    return meeting.uses_notes()
```
rjsparks, line 899: This looks incomplete - if the intent was to remove the filter L899 should also go.

ietf/doc/templatetags/ietf_filters.py
rjsparks: It's not clear that these should be filters - why aren't they methods on Meeting instead? (those would be visible to the templates). I think you were on the way to discovering that with the partial removal of the has_notes filter, and the redundant argument passed to the templates in the views later.

rjsparks (changes requested):
This is shaping up nicely. It should go a litter further as noted inline.

ietf/doc/templatetags/ietf_filters.py
rjsparks: (and I suspect all of these should just be removed once the functionality is moved to Meeting.

ietf/doc/templatetags/ietf_filters.py
richsalz: Happy to make them methods on the meeting if you prefer. I didn't know templates could invoke methods directly :)

ietf/doc/templatetags/ietf_filters.py
richsalz: yes, this was just checking in a snapshot before i got on the plane.  will move to templates and fix shortly.

ietf/doc/templatetags/ietf_filters.py
richsalz: > Should this get the same "Return true for interims" treatment?

I guess?  I changed it in the commit 466a6306a

richsalz:
Rebased, squashed, fixed the tests. This is ready for review now.

ietf/meeting/views.py
```diff
@@ -1691,7 +1690,7 @@ def api_get_agenda_data (request, num=None):
         },
         "categories": filter_organizer.get_filter_categories(),
         "isCurrentMeeting": is_current_meeting,
-        "useNotes": meeting.uses_notes(),
+        "uses_notes": meeting.uses_notes(),
```
jennifer-richards, line 1693: Should also stay as a camelCase key (`usesNotes`) to match the rest of the structure

jennifer-richards (changes requested):
Overall looks good - thank you for getting rid of the kludgy template code! There are a few nits to adjust that I've noted inline but I don't see any big issues.

richsalz:
@jennifer-richards I believe I addressed all your feedback, PTAL.

ietf/meeting/models.py
```diff
@@ -385,17 +385,19 @@ def uses_notes(self):
...
     def has_recordings(self):
         if self.type_id != 'ietf':
             return True
         num = self.get_number()
-        return num != None and num >= 80
+        return num is not None and num >= 80
 
     def has_chat_logs(self):
+        if self.type_id != 'ietf':
+            return True;
```
jennifer-richards, line 398: I think this should be `False` to preserve the behavior? For a non-ietf meeting, `self.get_number()` returns `None`, so it used to return `False`. (If it should be `True` or is a don't care, just say so, I'm just pointing out the change.)

jennifer-richards (commented):
One question inline

ietf/meeting/models.py
rjsparks: I think what's here now matches current and desired behavior. What existed earlier in this branch may have been off base about whether interims had chat logs, but the do (and always have).

rjsparks (changes requested):
Let me know if I misread the request for keeping the js variables in camelCase.

ietf/meeting/tests_views.py
```diff
@@ -259,7 +259,7 @@ def test_meeting_agenda(self):
                 },
                 "categories": rjson.get("categories"), # Just expect the value to exist
                 "isCurrentMeeting": True,
-                "useNotes": True,
+                "usesNotes": False,
```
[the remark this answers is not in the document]
richsalz: I added the archeological comment.  And fixed the other JS case.

client/agenda/AgendaScheduleList.vue
```diff
@@ -296,7 +296,7 @@ const meetingEvents = computed(() => {
           color: 'red'
         })
       }
-      if (agendaStore.useNotes) {
+      if (agendaStore.uses_notes) {
```
jennifer-richards, line 299: I think this one needs `usesNotes` also?

jennifer-richards (changes requested):
I think there's still one variable name change needed

client/agenda/AgendaScheduleList.vue
richsalz: dang.  but no change in test result, so I missed it.  fix pushed.

jennifer-richards:
I think there was still disagreement about `useNotes` vs `usesNotes`. Easier to push a commit than to point at them, but please review that I'm not mixing up similarly named variables in 4e8ce54cdbaeb35c77c370ba86e90a29ab629376

jennifer-richards (dismissed):

rjsparks (dismissed):

rjsparks (approved):

TohaAbir:
I need some time. I lost my phone. I am need to recover information how
check my activity

PR #7199, opened 2024-03-16
refactor: Remove decorator dependency from utils

### LINKED ISSUE #5654: Refactor to drop dependency on decorator package
### Description

We have a few decorators defined in `ietf/utils/decorator.py` that use the "decorator" package. This provides the `@decorator` decorator and a `decorate` method. The built-in Python `functools.wraps()` method can fulfill the needs here without the additional dependency. As far as I can tell we're not making use of any of the features provided by the "decorator" package.

The `@decorator` mechanism also seems to interfere with Django's `@method_decorator`, which led to refactoring the `requires_api_key` decorator already.

### Code of Conduct

microamp:
Fixes #5654 

I have noticed that `decorator` is also being used in `debug.py`, so it will remain in `requirements.txt` for now.

### DIFF
--- a/ietf/utils/decorators.py
+++ b/ietf/utils/decorators.py
@@ -4,7 +4,6 @@
 
 import datetime
 
-from decorator import decorator, decorate
 from functools import wraps
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
This looks ok - would be good to give the same treatment to `debug.py`, but that could go into a separate PR.

PR #7202, opened 2024-03-16
fix: Set develop URL correctly

### LINKED ISSUE #4507: TRACKER_BASE_URL should be overriden to the dev instance when in dev mode

richsalz:
feat: Change texts to say "./ietf/manage.py" throughout; add the "./" prefix.

feat: Change manage.py to have a default argument of 8001, and docs to not reference an explicit port so they pick up the default.

feat: Change manage.py so that if you give an argument other than 8001 print a warning message that the URLs may be wrong

fix: Override the IDTRACKER_BASE_URL in the settings_local file to point to "http://localhost:8000" under the assumption that the docker default port mapping is being used.

Fixes: #4507

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ If VS Code is not available to you, in your clone, type `cd docker; ./run`
 
 Once the containers are started, run the tests to make sure your checkout is a good place to start from (all tests should pass - if any fail, ask for help at tools-develop@). Inside the app container's shell type:
 ```sh
-ietf/manage.py test --settings=settings_test
+./ietf/manage.py test --settings=settings_test
 ```
 
 Note that we recently moved the datatracker onto PostgreSQL - you may still find older documentation that suggests testing with settings_sqlitetest. That will no longer work.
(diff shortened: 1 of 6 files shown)


jennifer-richards (changes requested):
See comment [here](https://github.com/ietf-tools/datatracker/issues/4507#issuecomment-2002154829) - I think this is paying attention to the wrong port for adjusting the URL. I like adding the default 8001 to runserver, but additional (and different) work is needed to make the host interface port available to `settings_local.py`

Also, I'm not sure the `./` prefix is needed? I'm in the habit of using `ietf/manage.py` without it. I don't object to changing the docs, though.

richsalz:
Easy question first: I don't have "." in my $PATH so I need the dot-slash. It's safe to always put it there.

As for getting CUSTOM_PORT setting, I tried to put into the environment by adding this line to docker-compose.extend.yml:
```
+        environment:
+            - CUSTOM_PORT:$CUSTOM_PORT
 ```
and I got a "no vaue; using empty string" So if you have hints on how to plumb the CUSTOM_PORT env var from the outside-docker down to the docker image running the app, please reply.  Or push a commit to this branch. :)

rjsparks:
This is a more complicated problem now than it was when it got the contrib-easy tag. The port may not be _knowable_ at  build time (using vscode, for example, the external port is something the app hunts for to find the next port available from the one that was asked for), so the work will really need to be runtime.

jennifer-richards:
> and I got a "no vaue; using empty string" So if you have hints on how to plumb the CUSTOM_PORT env var from the outside-docker down to the docker image running the app, please reply. Or push a commit to this branch. :)

How are you running the container? I think the `CUSTOM_PORT` only applies to starting from a command line with the `docker/run` script. if you're using VS Code and the devcontainer, that works differently.

richsalz:
> How are you running the  container?

Command line `./docker/run`

rjsparks:
Rich - thanks for this run at the problem, but I think we should close this PR and take a more comprehensive look at the different requirements for the different development run environments.. If you're ok with that, please close it.

richsalz:
Sure, never hurts to step back and think again.

PR #7203, opened 2024-03-16
fix: Use email or name when building community list view

### LINKED ISSUE #7036: Personal community list view crashes if email addresses are all inactive
### Describe the issue

If all email addresses for a logged-in user are deactivated (`active=False`), the `ietf.community.views.view_list()` view causes a 500 error if they view their own list.

This is because the view tries to look up the `manage_list` view with `clist.person.email_address`, which does not return an address in this case.

### Code of Conduct

microamp:
Fixes #7036 

An additional argument has been added to pass the `email_or_name` variable through.


rjsparks:
Since this is fixing a crash bug, we should add, or modify an existing test to include, a test that exercises the condition that led to the crash.

That test should also cover the view for person Y after having constructed a person X with name N and email E, and a person Y with name N (note same name) but no email.

rjsparks (changes requested):
See the earlier note re: tests

jennifer-richards (commented):
Looks good to me, agree that it should have a test.

jennifer-richards:
I added a test - which fails because it turns out this goes deeper. There's a similar issue in `list_menu.html` with the added wrinkle that it's included both in the view we're already touching and in `group_documents.html`.

microamp:
> I added a test - which fails because it turns out this goes deeper. There's a similar issue in `list_menu.html` with the added wrinkle that it's included both in the view we're already touching and in `group_documents.html`.

Thanks for having a go at it! (It was on my list at the code sprint, but I ran out of time.)

I'll have a look at the files you mentioned when I can.

rjsparks:
@jennifer-richards @microamp Is this still something to pursue?

jennifer-richards (approved):
lgtm

In my previous comments I was worried because the `email_or_name` variable is not set when the `list_menu.html` is included via `group_documents.html` in the `group_documents()` view. On closer review, there are guards in place so the `email_or_name` is not needed in the group context.

rjsparks (approved):

PR #7204, opened 2024-03-16
fix: add celery tasks for idnits2 and bibxml file generation

### LINKED ISSUE #7192: Create tasks to replace generate_* management commands
### Description

Need to create Celery tasks to replace the following management commands
* `ietf/doc/management/commands/generate_idnits2_rfc_status.py`
* `ietf/doc/management/commands/generate_idnits2_rfcs_obsoleted.py`
* `ietf/doc/management/commands/generate_draft_bibxml_files.py`

The last of these has options that I believe are not used, so it would be fine if it assumes the default for all options (though it wouldn't hurt to keep them as arguments to the task).

We should keep the management commands around for now, we will remove them once we've switched to the task-based implementations.

### Code of Conduct

russhousley:
Add celery tasks for idnits2 and bibxml file generation. Fixes #7192


ietf/doc/tests_tasks.py
```diff
@@ -61,3 +69,66 @@ def test_notify_expirations_task(self, get_drafts_mock, send_warning_mock):
...
+        os.mkdir(bibxmldir)
+        filepath = os.path.join(bibxmldir, filename)
+        self.assertFalse(os.path.exists(filepath))
+        generate_bibxml_files_for_recent_drafts_task(days=7)
+        self.assertTrue(os.path.exists(filepath))
+        with io.open(filepath, encoding='utf-8') as f:
+            content = f.read()
+        self.assertIn(draft.title, content)
+
+    def test_generate_bibxml_files_for_recent_drafts_task_with_bad_vakue(self):
```
jennifer-richards, line 130: typo: `vakue` -> `value`

jennifer-richards (changes requested):
One typo, one small change I think we should make.

ietf/doc/utils.py
```diff
@@ -1259,6 +1272,33 @@ def bibxml_for_draft(doc, rev=None):
...
+def bibxml_for_all_drafts(bibxmldir):
+    doc_events = NewRevisionDocEvent.objects.filter(type='new_revision', doc__type_id='draft').order_by('time')
+    for e in doc_events:
+        try:
+            bibxml = bibxml_for_draft(e.doc, e.rev)
+            ref_rev_file_name = os.path.join(bibxmldir, 'reference.I-D.%s-%s.xml' % (e.doc.name, e.rev))
+            write_bibxml_file_if_changed(ref_rev_file_name, bibxml)
+        except Exception as ee:
+            log.log('\n%s-%s: %s\n' % (e.doc.name, e.doc.rev, str(ee)))
+            raise ee
```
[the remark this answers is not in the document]
russhousley: Sounds fine to me.

ietf/doc/tests_tasks.py
russhousley: Goof catch.  I read past it many times.

jennifer-richards:
In the scramble to cut over to the new infrastructure, I lost track of this PR and wound up reimplementing it slightly differently. I'm merging that over so we can merge this.

jennifer-richards (dismissed):

jennifer-richards (approved):

PR #7206, opened 2024-03-16
refactor: Move document state help to /doc

jimfenton:


ietf/help/views.py
```diff
@@ -1,30 +1,20 @@
...
-                
-    return render(request, 'help/state_index.html', {"types": types})
+    return redirect('/doc/help/state/', permanent = True)
 
 def state(request, doc, type=None):
     if type:
         streams = [ s.slug for s in StreamName.objects.all() ]
         if type in streams:
             type = "stream-%s" % type
     slug = "%s-%s" % (doc,type) if type else doc
```
rjsparks, line 18: else doc.type_id maybe?

ietf/doc/views_help.py
```diff
@@ -9,6 +9,18 @@
...
+    types = StateType.objects.all()
+    names = [ type.slug for type in types ]
+    for type in types:
+        if "-" in type.slug and type.slug.split('-',1)[0] in names:
+            type.stategroups = None
+        else:
+            groups = StateType.objects.filter(slug__startswith=type.slug)
+            type.stategroups =  [ g.slug[len(type.slug)+1:] for g in groups if not g == type ] or ""
+                
+    return render(request, 'doc/state_index.html', {"types": types})
```
rjsparks, line 22: the concept of `used` on State probably needs to be rendered here somewhere - maybe that can be done entirely in the templates.

ietf/help/views.py
```diff
@@ -1,30 +1,20 @@
...
-    names = [ type.slug for type in types ]
-    for type in types:
-        if "-" in type.slug and type.slug.split('-',1)[0] in names:
-            type.stategroups = None
-        else:
-            groups = StateType.objects.filter(slug__startswith=type.slug)
-            type.stategroups =  [ g.slug[len(type.slug)+1:] for g in groups if not g == type ] or ""
-                
-    return render(request, 'help/state_index.html', {"types": types})
+    return redirect('/doc/help/state/', permanent = True)
```
rjsparks, line 11: This redirect could be moved directly into the URLs file, and the view code and test for it can be removed.

rjsparks (changes requested):
Jim - this is great and gets most of the way there, but we need to touch a few things before bringing this into main.

ietf/help/views.py
jimfenton: No, in this context `doc` is a simple string variable that comes from a portion of the URL. Maybe that's a poor choice of variable name (preexisting).

jimfenton:
Also noticed that doc/help/state/draft-iana/ now returns a 404. /help/state/draft/iana/ didn't produce a very useful page before, but at least it didn't generate a 404.

rjsparks:
> Also noticed that doc/help/state/draft-iana/ now returns a 404. /help/state/draft/iana/ didn't produce a very useful page before, but at least it didn't generate a 404.

Interesting catch - `draft-iana` isn't a document state type. I think that it passing through the URLs to date was a bug.

rjsparks:
I've deleted the unused StateType in production using the admin.

ietf/doc/views_help.py
rjsparks: nevermind - the bit is badly used in the State table.

rjsparks (approved):

Issue #6902, opened 2024-01-09
Redundant word in banner for Legacy stream documents


rjsparks:
![image](https://github.com/ietf-tools/datatracker/assets/10996692/6476aea7-3140-4435-a1bd-cdee1d9f1ff3)

Note `Legacy stream stream`

rjsparks:
The fix for this might be to fix the data rather than the code:
![image](https://github.com/ietf-tools/datatracker/assets/10996692/0059c2c2-baf0-4357-bbc3-3a092e26760f)
but we'd have to look to see where `desc` is otherwise used to make sure it's not pushing the problem around inappropriately

rjsparks:
On quick inspection, I don't find places that would suffer loss of fidelity if we changed the `desc` value for the `legacy` `StreamName` object

rjsparks:
The Last call announcement text would probably benefit from having "stream" added where the `desc` is used:
https://github.com/ietf-tools/datatracker/blob/e24fb601db9e81c467aadf28b82fdc2d259b0f16/ietf/templates/doc/mail/last_call_announcement.txt#L34-L37
Similarly:
https://github.com/ietf-tools/datatracker/blob/e24fb601db9e81c467aadf28b82fdc2d259b0f16/ietf/secr/templates/telechat/doc.html#L88-L90
and
https://github.com/ietf-tools/datatracker/blob/e24fb601db9e81c467aadf28b82fdc2d259b0f16/ietf/templates/iesg/moderator_doc.html#L167-L169

PR #7207, opened 2024-03-16
fix: redundant word in banner for Legacy stream documents

richsalz:
This does all the changes suggested by @rjsparks in #6902 and a little bit more cleanup

### DIFF
--- a/ietf/doc/views_doc.py
+++ b/ietf/doc/views_doc.py
@@ -603,12 +603,7 @@ def document_main(request, name, rev=None, document_html=False):
         additional_urls = doc.documenturl_set.exclude(tag_id='auth48')
 
         # Stream description and name passing test
-        if doc.stream != None:
-            stream_desc = doc.stream.desc
-            stream = "draft-stream-" + doc.stream.slug
-        else:
-            stream_desc = "(None)"
-            stream = "(None)"
+        stream = ("draft-stream-" + doc.stream.slug) if doc.stream != None else "(None)"
 
         html = None
         js = None
@@ -647,7 +642,6 @@ def document_main(request, name, rev=None, document_html=False):
                                        revisions=simple_diff_revisions if document_html else revisions,
                                        snapshot=snapshot,
                                        stream=stream,
-                                       stream_desc=stream_desc,
                                        latest_revision=latest_revision,
                                        latest_rev=latest_rev,
                                        can_edit=can_edit,
--- a/ietf/name/fixtures/names.json
+++ b/ietf/name/fixtures/names.json
@@ -14058,7 +14058,7 @@
   },
   {
     "fields": {
-      "desc": "Legacy stream",
+      "desc": "Legacy",
       "name": "Legacy",
       "order": 6,
       "used": true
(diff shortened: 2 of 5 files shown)


ietf/name/fixtures/names.json
```diff
@@ -14042,7 +14042,7 @@
   },
   {
     "fields": {
-      "desc": "Legacy stream",
+      "desc": "Legacy",
```
rjsparks, line 14045: This won't actually change production. it will show up here as a side-effect of the change in the database instead. The PR will need a data migration to change the name.

rjsparks (changes requested):
See inline comments

ietf/name/fixtures/names.json
richsalz: Got it.  It will take me a few days to re-understand the migration stuff. If you want to add a commit or hints to me, feel free.  (Not asking, just not wanting to be a blocker if I don't have to be)

rjsparks: No worries, this can wait awhile

rjsparks (dismissed):

ietf/doc/views_doc.py
```diff
@@ -603,12 +603,7 @@ def document_main(request, name, rev=None, document_html=False):
         additional_urls = doc.documenturl_set.exclude(tag_id='auth48')
 
         # Stream description and name passing test
-        if doc.stream != None:
-            stream_desc = doc.stream.desc
-            stream = "draft-stream-" + doc.stream.slug
-        else:
-            stream_desc = "(None)"
-            stream = "(None)"
+        stream = ("draft-stream-" + doc.stream.slug) if doc.stream != None else "(None)"
```
jennifer-richards, line 606: nit: "is not None" instead of "!= None"

jennifer-richards (dismissed):
Just a minor nit - I'm ok with committing as is or making the change.

ietf/doc/views_doc.py
richsalz: There are seven "!= None" and four "is not None" in the file.  Let me know if you want to standardize on one.

jennifer-richards: Should definitely be `is not None` (citation: https://peps.python.org/pep-0008/#programming-recommendations)

We're not obsessively zealous about it, but I try to rephrase things as I encounter them and it'd be great if you don't mind cleaning these up!

jennifer-richards (changes requested):
Having a little trouble following the history because of the force pushing, but it looks like at some point there had been a migration to make the change on production that Robert had suggested and it was taken out?

I think it needs that unless we're planning to make the change through the admin, which would not be unreasonable since it's a single value.

Aside from that, lgtm.

rjsparks:
I built and pushed a migration - I suspect Rich missed that and didn't merge back into his branch.
Rich - can you cherry pick it from the history, or do you want me to do that and push that commit back in here again?

richsalz:
@rjsparks if you push the commit, then I'll probably leave this untouched as @jennifer-richards basically approved and I'll just keep my hands off :)

rjsparks:
for other PRs - if someone else commits to richsalz/datatracker `<PR branch>`, remember to pull those before rebasing. Also consider never rebasing unless something has gone horribly wrong - rather, just merge main into your work.

rjsparks:
migration cherry-picked back into the PR

richsalz:
> consider never rebasing

sorry for that.  Old habit I picked up with OpenSSL, the first place I learned git.

jennifer-richards (approved):

PR #7209, opened 2024-03-17
fix: Add vim temp files to .gitignore

richsalz:


rjsparks (approved):

rjsparks:
This made me look at https://github.com/github/gitignore/blob/main/Global/Vim.gitignore, which seems overly aggressive.

PR #7215, opened 2024-03-17
feat: use gunicorn

rjsparks:


dev/build/datatracker-start.sh
```diff
@@ -7,4 +7,14 @@ echo "Running Datatracker migrations..."
...
+gunicorn \
+          --workers 53 \
+          --max-requests 32768 \
+          --timeout 180 \
+          --bind :8000 \
+          --error-logfile gunicorn_error.log \
+          --log-level info \
+          ietf.wsgi:application
+
+          #--env SCOUT_NAME=Datatracker \
```
richsalz, line 20: ooh, does this allow environment settings?  Could we add something like
```
--env CUSTOM_PORT:$CUSTOM_PORT
```
and then filter that through the sed command?

jennifer-richards: (note that this PR is targeting the feat/helm branch, so won't be directly relevant to the dev containers - at least not immediately)

This script is running inside the container - it's what runs when you start that up. The `CUSTOM_PORT` isn't set here

jennifer-richards (approved):

jennifer-richards (approved):
Looks good to me. My only worry is whether the python-json-logger project will continue to be maintained (current release is a year old, little maintainer activity since - also see https://github.com/madzak/python-json-logger/issues/187). I think it's adequate for now, at least, just want to keep an eye on it.

PR #7219, opened 2024-03-18
refactor: Remove decorator dependency from debug.py

microamp:
The PR removes the `decorator` dependency from `debug.py`. This should complete #5654.

See also: #7199

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,8 +8,6 @@ bleach>=6
 types-bleach>=6
 celery>=5.2.6
 coverage>=4.5.4,<5.0    # Coverage 5.x moves from a json database to SQLite.  Moving to 5.x will require substantial rewrites in ietf.utils.test_runner and ietf.release.views
-decorator>=5.1.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -56,7 +54,7 @@ pyquery>=1.4.3
 python-dateutil>=2.8.2
 types-python-dateutil>=2.8.2
 python-magic==0.4.18    # Versions beyond the yanked .19 and .20 introduce form failures
-pymemcache>=4.0.0  # for django.core.cache.backends.memcached.PyMemcacheCache 
+pymemcache>=4.0.0  # for django.core.cache.backends.memcached.PyMemcacheCache
```
rjsparks, line 57: nit - try not to bring in unrelated whitespace changes - put those in their own PR. (Otherwise time gets wasted later chasing back through revisions to find why the lines really changed.

rjsparks (approved):

jennifer-richards (approved):

PR #7221, opened 2024-03-18
fix: Disable send-review-reminders from bin/daily

jennifer-richards:
This celery task was enabled a while ago, so the call here was redundant (and causing doubled reminders to be sent)

### DIFF
--- a/bin/daily
+++ b/bin/daily
@@ -47,5 +47,5 @@ $DTDIR/ietf/manage.py run_yang_model_checks -v0
 # Enable when removed from /a/www/ietf-datatracker/scripts/Cron-runner:
 $DTDIR/ietf/bin/expire-last-calls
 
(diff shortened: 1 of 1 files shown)


bin/daily
```diff
@@ -47,8 +47,5 @@ $DTDIR/ietf/manage.py run_yang_model_checks -v0
 # Enable when removed from /a/www/ietf-datatracker/scripts/Cron-runner:
 $DTDIR/ietf/bin/expire-last-calls
 
-# Send reminders originating from the review app
-$DTDIR/ietf/bin/send-review-reminders
-
-# Purge older PersonApiKeyEvents
+I# Purge older PersonApiKeyEvents
```
rjsparks, line 50:
```suggestion
# Purge older PersonApiKeyEvents
```

rjsparks (approved):

PR #7248, opened 2024-03-22
ci: parameterize / update settings

jennifer-richards:
~Creating as a draft because this is still a WIP.~ In `helm/settings_local.py`, I've made a stab at pulling settings out to the environment. Major question is whether this is the right balance.

Notably not resolved is how the API keys dict gets populated - it's currently filled in with "redacted" values.

### DIFF
--- a/helm/settings_local.py
+++ b/helm/settings_local.py
@@ -8,10 +8,10 @@ from ietf import __release_hash__
 from ietf.settings import *                                          # pyflakes:ignore
 
 
-SERVER_MODE = os.environ.get("DATATRACKER_SERVER_MODE", "development")  # todo decide if we need a "staging" mode
+# Default to "development". Production _must_ set DATATRACKER_SERVER_MODE="production" in the env!
+SERVER_MODE = os.environ.get("DATATRACKER_SERVER_MODE", "development")
 
 # Secrets
-
 if "DATATRACKER_DJANGO_SECRET_KEY" in os.environ:
     SECRET_KEY = os.environ.get("DATATRACKER_DJANGO_SECRET_KEY")
 elif SERVER_MODE == "production":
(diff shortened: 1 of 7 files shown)


NGPixel (approved):

helm/settings_local.py
```diff
@@ -1,124 +1,231 @@
...
+
+from ietf import __release_hash__
 from ietf.settings import *                                          # pyflakes:ignore
 
-ALLOWED_HOSTS = ['*']
+USE_TZ = True
+
+# Secrets
+
+SECRET_KEY = os.environ.get("DATATRACKER_DJANGO_SECRET_KEY")
```
richsalz, line 14: One possibility is to wrap the `os.environ.get` calls here so that if the value is "file:/foo/bar/baz" you make the value returned read the specified file.

jennifer-richards: Interesting thought. At the moment the environment is the method we expect to be most convenient for production deployment, but something like this might help if/when we try to move back to a more unified setup. Will keep it in mind. Thanks!

jennifer-richards:
I believe this is ready for review - I have also adopted the changes suggested in #7250 (but check whether I missed anything because I did it manually)

jennifer-richards:
Also, note in particular [8537e48](https://github.com/ietf-tools/datatracker/pull/7248/commits/8537e4875122d39d2ca15d3bd216c35bc337b146). I think we want to do this (and I think it'll work without issue on ietfa while making things slightly easier on the new infrastructure)

Additionally [7132fcb](https://github.com/ietf-tools/datatracker/pull/7248/commits/7132fcb11ddb804f13b7988e38b8cb94e80ac7cc), [882993e](https://github.com/ietf-tools/datatracker/pull/7248/commits/882993ea4ee36ac661674b5119999e4ec6d7e6a1), and [586093c](https://github.com/ietf-tools/datatracker/pull/7248/commits/586093c358b0bc9b2cda599a0b80c38ec9a2cd07) remove some unused things from `settings.py`. This is far enough afield here that I wouldn't push back if you suggested extracting those to a separate PR.

rjsparks (approved):

PR #7250, opened 2024-03-22
chore: expect artifacts on /a

rjsparks:
Some more potential settings adjustments to be considered along with https://github.com/ietf-tools/datatracker/pull/7248


jennifer-richards:
I think I've rolled the suggestions here into #7248

PR #7252, opened 2024-03-23
ci: collectstatic --no-input

jennifer-richards:
Needed to prevent errors when starting up a container if there are updated timestamps.

I think we should eventually get rid of the collectstatic call for production, though?


rjsparks:
Production needs the statics collected locally as well as referenceable in another location (some of the apps use them). We could have a build step make a local thing that holds them instead of building them as part of the deploy.

rjsparks (approved):

PR #7258, opened 2024-03-25
fix: Exclude replaced documents from IESG discusses

### LINKED ISSUE #7179: Replaced documents show on https://datatracker.ietf.org/iesg/discusses/ - they shouldn't.
At the moment draft-faltstrom-unicode-11 shows on the discusses page.

pselkirk:
Fixes #7179


rjsparks (changes requested):
This is good (though it could be simplified to chain the exclude onto the existing statement, and the whole statement could benefit from being processed by black.)

But there should be a test that proves it. I suspect it will be a small adjustment to an existing test.

rjsparks:
OBE by #7712

PR #7283, opened 2024-04-02
ci: shared PersistentVolumeClaim on /a

jennifer-richards:
Tested in minikube by setting `accessModes` to `ReadWriteOnce` and confirming that datatracker / celery / beat all saw the same filesystem.

### DIFF
--- a/helm/templates/persistentvolumeclaims.yaml
+++ b/helm/templates/persistentvolumeclaims.yaml
@@ -1,6 +1,17 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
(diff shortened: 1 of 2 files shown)


NGPixel (approved):

rjsparks (approved):
400Gi?

jennifer-richards:
> 400Gi?

That was my memory of the rough size of the snapshot we mean to mount on `/a` - this value will probably need to change in ways that depend on how the infrastructure sets up that volume, so I didn't worry much about it. (For minikube quick-and-dirty testing, I changed this to 1 Gi)

PR #7291, opened 2024-04-04
feat: API to list role-holder addresses

jennifer-richards:
Adds API endpoint needed to decouple the `makewhitelist` script from the datatracker.

The code is refactored compared to the old `list-role-holder-emails` script ([final version here](https://github.com/ietf-tools/datatracker/blob/b5bf6789da27e5077fa6708dafc8f77167aed989/ietf/bin/list-role-holder-emails)), but I confirmed that at least for a recent dev database, the refactored and original code agree.

Also: I've reused the `ietf.api.views.email_aliases` API token here because it's what the rest of the API calls used by the postconfirm scripts use. I'm a little uneasy about the now misleading name, but have kept it for operational simplicity. I'd be open to either renaming the token or using a dedicated token for this endpoint.


rjsparks:
I think we should use a separate api token named for this view. We may have additional clients of the APIs in the future, and we should be able to make the decision to limit one of the other of these endpoints from those additional clients.

rjsparks:
The functionality looks fine. The list of groups to scrape from should get leadership review (I expect we will add groups and group types to it), but that does not have to happen now.

jennifer-richards:
I think I'm done now. In addition to straightening out the API token reuse, I realized I'd deviated from the pattern used for the last few API endpoints. I've refactored to do the Group-specific work in `ietf.group.utils.role_holder_emails()` and put the API-specific interface in `ietf.api.views.role_holder_addresses()`. 

May not be perfect, but it's at least consistent.

rjsparks (dismissed):

rjsparks (approved):

PR #7336, opened 2024-04-17
feat: obviate ghostlinkd

rjsparks:


dev/deploy-to-container/settings_local.py
```diff
@@ -60,10 +60,11 @@
 BOFREQ_PATH = '/assets/ietf-ftp/bofreq/'
 CONFLICT_REVIEW_PATH = '/assets/ietf-ftp/conflict-reviews/'
 STATUS_CHANGE_PATH = '/assets/ietf-ftp/status-changes/'
-INTERNET_DRAFT_ARCHIVE_DIR = '/assets/archive/id'
+INTERNET_DRAFT_ARCHIVE_DIR = '/assets/collection/draft-archive'
 INTERNET_ALL_DRAFTS_ARCHIVE_DIR = '/assets/archive/id'
```
jennifer-richards, line 64: Should `INTERNET_ALL_DRAFTS_ARCHIVE_DIR` change to `/assets/ietf-ftp/internet-drafts` to match other places?

docker/configs/settings_local.py
```diff
@@ -46,10 +46,11 @@
 BOFREQ_PATH = '/assets/ietf-ftp/bofreq/'
 CONFLICT_REVIEW_PATH = '/assets/ietf-ftp/conflict-reviews/'
 STATUS_CHANGE_PATH = '/assets/ietf-ftp/status-changes/'
-INTERNET_DRAFT_ARCHIVE_DIR = '/assets/archive/id'
+INTERNET_DRAFT_ARCHIVE_DIR = '/assets/collection/draft-archive'
 INTERNET_ALL_DRAFTS_ARCHIVE_DIR = '/assets/archive/id'
```
jennifer-richards, line 50: Similarly - where should `INTERNET_ALL_DRAFTS_ARCHIVE_DIR` point or is it really different for different contexts?

jennifer-richards (commented):
Reviewed what's here. Agree that you've caught all the places in your comments in the first commit. A couple minor questions that might just reflect the wip-ness of this PR.

Also, I think you can remove your `TODO` in `views_charter.py` now?

dev/deploy-to-container/settings_local.py
rjsparks: Yes, but not as part of this PR I think. (That should be part of a full rework of the various environments idea of the structure of /assets).

rjsparks:
fwiw, for each unit, I left “bury the errors” vs “crash if there are errors” strategies as they are. Both of them bug me immensely. But to do a better thing all these places (where the errors are really _bad_, and crashing is really _bad_) will require a bigger refactor.

rjsparks:
Also (for future us looking back at why we did this likely-to-look-foolish thing) - this moves a baked-in assumption into the datatracker from ghostlinkd (that the linked things are on the same filesystem and hardlinks will work).

Hopefully in the near future, we will keep all the things touched here in a blobstore instead of the filesystem and the above assumption (and the things this PR adds) will be simplified/rationalized into other applications feeding from that blobstore.

ietf/idindex/tasks.py
```diff
@@ -43,9 +49,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 @shared_task
 def idindex_update_task():
     """Update I-D indexes"""
+    # Why are these not using values from django.conf.settings?
```
jennifer-richards, line 52: 😵‍💫 

They should - probably leftover from when this was shell code in `bin/hourly` and copied over too-faithfully

ietf/idindex/tasks.py
```diff
@@ -69,17 +78,18 @@ def idindex_update_task():
         derived_all_id2_tmpfile = tmp_mgr.make_temp_file(all_id2_content)
 
         # Move temp files as-atomically-as-possible into place
-        tmp_mgr.move_into_place(all_id_tmpfile, id_path / "all_id.txt")
+        # TODO: multiwrite - anything being written into id_path also needs to be written to the all archive path
```
jennifer-richards, line 81: todo: remove this todo

jennifer-richards (dismissed):
Looks good, just a leftover "todo" comment to remove that I see.

WRT my comment about wrapping the `link()` operation - I think we'll end up wanting to do that, but I don't think it's necessary now. The main advantage is that I think it'd make the tests more independent, which might be slightly easier to change now. (Instead of separately testing that each place makes the right hard links, it'd just be a test that the hypothetical method was called, then a single test to update when that changes to a new mirroring strategy).

I don't think it'll be _that_ much easier to do this now vs later, though.

yasar456uhyhluih:
update_in_archives()

ietf/settings.py
```diff
@@ -685,6 +685,7 @@ def skip_unreadable_post(record):
 MEETING_RECORDINGS_DIR = '/a/www/audio'
 DERIVED_DIR = '/a/ietfdata/derived'
 FTP_DIR = '/a/ftp'
+ALLID_DOWNLOAD_DIR = '/a/www/www6s/download'
```
jennifer-richards, line 688: Is `ALLID` a term in use already? If not, maybe `ALL_ID`? I struggle to parse the former.

jennifer-richards (approved):

PR #7366, opened 2024-04-25
ci: run datatracker pod as non-root user

### LINKED ISSUE #4820: Move "apply patches" check to its own management command
### Description

One of the checks in `checks.py`, run by `ietf/manage.py check`, applies patches from the `patch/` path to the Python libraries. In some cases, such as in the celery beat Docker container, we need to be sure these patches have been applied but do not need to run any of the other checks. Would be nice to break the patching out to a separate management command.

The check command could continue to apply the patches, but should at least fail if the patches haven't been applied.

### Code of Conduct

jennifer-richards:
Fixes #4820

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -67,9 +67,6 @@ datatracker:
   podAnnotations: {}
   podLabels: {}
 
-  podSecurityContext: {}
-    # fsGroup: 2000
-
 #readinessProbe:
 #  httpGet:
 #    # /submit/tool-instructions/ just happens to be cheap until we get a real health endpoint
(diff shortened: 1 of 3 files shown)


rjsparks (approved):

helm/values.yaml
```diff
@@ -90,13 +87,17 @@ datatracker:
     #   cpu: 100m
     #   memory: 128Mi
```
NGPixel, line 89: I don't think it's defined in the deployment template?

jennifer-richards: Isn't that https://github.com/ietf-tools/datatracker/blob/145c9917252218e83639e2acbb2b444425b89a7a/helm/templates/deployments/datatracker.yaml#L27-L28 ?

NGPixel: Ah indeed. I missed the different naming.

NGPixel (approved):

PR #7376, opened 2024-04-29
fix: improve file investigation results

rjsparks:


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -906,13 +906,17 @@ def mtime(path):
...
+    return path.stat().st_mtime == 0
+
 @register.filter
 def url_for_path(path):
     """Consructs a 'best' URL for web access to the given pathlib Path object.
 
     Assumes that the path is into the Internet-Draft archive or the proceedings.
     """
-    if path.match(f"{settings.AGENDA_PATH}/**/*"):
+    if Path(settings.AGENDA_PATH) in path.parents:
```
jennifer-richards, line 919: These are not exactly synonymous: 
```python
>>> p = Path("/a/b/c")
>>> Path("/a/b/") in p.parents
True
>>> p.match("/a/b/**/*")
False
```

jennifer-richards (changes requested):
Only "Requesting changes" to ask whether the observation in my inline comment is worrying.

ietf/doc/templatetags/ietf_filters.py
rjsparks: Right - this is exactly one of the issues this PR is addressing.

jennifer-richards: Ah- great!

jennifer-richards (approved):

PR #7418, opened 2024-05-14
feat: get tool versions without VersionInfo model

jennifer-richards:
* chore: remove update_external_command_info call

* feat: get tool version without VersionInfo

* chore: Remove VersionInfo model

* chore: Migration to remove VersionInfo

* fix: handle errors better; ignore stderr


ietf/utils/__init__.py
```diff
@@ -1 +1,29 @@
...
+        "yanglint",
+    ]
+    _versions: dict[str, str] = dict()
+
+    def __getitem__(self, item):
+        if item not in self._known:
+            return "Unknown"
+        elif item not in self._versions:
+            try:
+                self._versions[item] = subprocess.run(
```
rjsparks, line 19: One thought that occurs to me is that this puts the time to invoke these 4 things (which _should_ be fast) in the warmup path of every worker thread.

I'm willing to give this a shot, but if we're not already planning to do so in feat/k8s, maybe this should be replaced with something that happens at build time?

rjsparks (approved):

ietf/utils/__init__.py
jennifer-richards: Yes, moving to a build time mechanism would be better. I expect that since we generally (or maybe always) actually do work with the tool we ask for the version of, any overhead will be really small by comparison. I guess that's not guaranteed though. 

This has the virtue of working both on ietfa and in a container, so it'll bridge the move to the new infrastructure.

PR #7435, opened 2024-05-19
fix: validate form in login()

jennifer-richards:
This prevents 500s if username/password contain null characters. It refactors our custom `login()` view as a custom `LoginView` subclass and an associated custom `AuthenticationForm` subclass. 

The following applied to the first commit only (which broke the ability to log in using any email address):
~~The error is happening in some checks we do prior to the actual authentication check - it's looking to see if the User's password was cleared and whether it has a Person. The actual auth happens in the `LoginView.as_view()` call, which already checks and rejects attempts involving null characters. If that were to change, this PR would potentially make it possible someone to bypass the checks for a cleared password or missing Person. I don't think that's either likely or harmful, but wanted to point at it.~~

### DIFF
--- a/ietf/ietfauth/tests.py
+++ b/ietf/ietfauth/tests.py
@@ -50,8 +50,6 @@ from ietf.utils.test_utils import TestCase, login_testing_unauthorized
 from ietf.utils.timezone import date_today
 
 
-import ietf.ietfauth.views
-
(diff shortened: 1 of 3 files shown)


rjsparks (approved):

Hamidzai456hzj (commented):
آخرین اپدیتش را میخواهم

PR #7462, opened 2024-05-23
feat: IESG table column 'Pre pubreq'

### LINKED ISSUE #4577: Add Total Number of I-Ds to the I-D State Count table in the AD Dashboard
### Description

The I-D State Count table of the AD Dashboard (https://datatracker.ietf.org/doc/ad) currently lists the documents an AD is shepherding from PubReq to Approval.  To provide a sense of the load prior to PubReq, recommend adding a column as follows:

OLD I-D State Count
* Column 1 = AD Name
* Column = Pub Req

NEW I-D State Count
* Column 1 = AD Name
* Column 2 = Total I-Ds = a count of adopted I-Ds in the WGs overseen by the AD + any independent documents the AD is assigned to (in cases of AD sponsored documents) 
* Column 3 = Pub Req

### Code of Conduct

holloway:
Screenshot from 2024-05-24 08-42-39

Fixes https://github.com/ietf-tools/datatracker/issues/4577

### DIFF
--- a/ietf/templates/doc/ad_list.html
+++ b/ietf/templates/doc/ad_list.html
@@ -39,8 +39,8 @@
                         <th scope="col" data-sort="pre-pubreq">Pre pubreq</th>
                     {% endif %}                    
                     {% for state, state_name in dt.states %}
-                        <th scope="col" class="col-1" title=""
-                            data-sort="{{ state }}-num">
+                        <th scope="col" class="col-1" data-sort="{{ state }}-num"
+                        >
                             <a href="{% url 'ietf.doc.views_help.state_help' type='draft-iesg' %}#{{ state }}">
                                 {{ state_name|split:'/'|join:'/<wbr>' }}
                             </a>
@@ -58,6 +58,7 @@
                             <td
                                 class="col-1 align-bottom"
                                 data-sum="pre-pubreq"
+                                data-series-graph
                             >
                                 {{ ad.pre_pubreq }}
                             </td>
@@ -76,7 +77,7 @@
                     <th scope="row">Sum</th>
                     {% if dt.type.1 == "Internet-Draft" %}
                         <td class="align-bottom">
-                            <div data-sum-result="pre-pubreq"></div>
+                            <div data-sum-result="pre-pubreq" data-series-graph></div>
                         </td>
                     {% endif %}
                     {% for state, state_name in dt.states %}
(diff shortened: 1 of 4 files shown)


ietf/templates/doc/ad_list.html
```diff
@@ -87,6 +103,37 @@ <h2 class="mt-5" id="{{ dt.type.0 }}">{{ dt.type.1 }} State Counts</h2>
     </script>
     <script src="{% static "ietf/js/highcharts.js" %}"></script>
     {{ data|json_script:"data" }}
+
+    <script>
+         Array.from(document.querySelectorAll("table"))
+            .filter(table => {
+                const hasDataSum = Boolean(table.querySelector("[data-sum]"))
+                return hasDataSum
```
NGPixel, line 111: Nitpicking here: avoid creating a variable that's never used afterward. return the result directly.

ietf/templates/doc/ad_list.html
```diff
@@ -87,6 +103,37 @@ <h2 class="mt-5" id="{{ dt.type.0 }}">{{ dt.type.1 }} State Counts</h2>
...
+            })
+            .forEach(table => {
+                const sums = Array.from(table.querySelectorAll("[data-sum]")).reduce((
+                    sums,
```
NGPixel, line 115: There are 2 variables named "sums" in 2 different scopes, which can cause confusion when reading.

NGPixel (changes requested):

jennifer-richards (commented):
One inline comment. Also, as a general matter, it'd be nice to keep chipping away at moving javascript out of inline `<script>` tags and into their own js files. It's not exactly on-topic for this PR, though.

ietf/templates/doc/ad_list.html
```diff
@@ -87,6 +103,37 @@ <h2 class="mt-5" id="{{ dt.type.0 }}">{{ dt.type.1 }} State Counts</h2>
...
+                    cell
+                ) => {                    
+                    const key = cell.dataset.sum
+                    const value = parseFloat(cell.textContent)
```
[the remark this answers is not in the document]
hollowaysmith: So something like this?

```javascript
function safeParseFloat(text) {
        const trimNumber = text.trim()
        if(!trimNumber.match(/^[0-9.]+$/)) {
            console.warn(`Unable to parse "${trimNumber}" as a number.`)
            return Number.NaN
        }
        return parseFloat(text)
}
```

I think having tests for this is going to be hard with inline scripts but I guess I could put it into Playwright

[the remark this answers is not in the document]
holloway: I'll try refactoring this JS into a separate script in the next PR.

playwright/tests-legacy/docs/ad.spec.js
```diff
@@ -0,0 +1,26 @@
+const { test, expect } = require('@playwright/test')
+const viewports = require('../../helpers/viewports')
+
+// ====================================================================
+// NOMCOM - EXPERTISE
```
NGPixel, line 5: Wrong title

ietf/templates/doc/ad_list.html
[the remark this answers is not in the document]
jennifer-richards: That would do it, though we could also do it on the Django side just looking at the rendered template. That might be easier to turn into a test failure without having to add Playwright coverage or resort to `console.warn`.

ietf/templates/doc/ad_list.html
holloway: Fixed

ietf/templates/doc/ad_list.html
holloway: Fixed

NGPixel (approved):

PR #7492, opened 2024-05-31
fix: correct links to early meeting proceedings.

rjsparks:


jennifer-richards (approved):
Looks fine, though I wonder if there'd be any value in teaching the `agenda` view to deal with these? (I guess it still needs a special case to know to give back a PDF instead of defaulting to HTML so maybe not worth it)

Issue #7390, opened 2024-05-01
Cannot scroll metadata sidebar


becarpenter:
### Describe the issue

In pages like https://datatracker.ietf.org/doc/html/draft-carpenter-6man-zone-ui-04, it's nice that the metadata sidebar can be switched on or off, but when it's present, it can longer be scrolled up or down and most of it is inaccessible without zooming way out.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

cabo:
Interesting.  WFM (tried current Safari and Chromium).
What was your Browser?

becarpenter:
Firefox 117 on Windows 10. Same problem with Chrome and Edge on Windows. I can't test on Linux as I'm on travel.

On an Android phone things are different, as the metadata sidebar isn't presented as a sidebar, and things work as intended.

jennifer-richards:
Firefox 125 on macOS allows the metadata to be scrolled using a scroll wheel or trackpad gesture when the cursor is over the sidebar, but there's no associated scroll bar widget for the sidebar.

becarpenter:
I don't have a mouse with me this week, but you're right, a very non-intuitive trackpad gesture does work (hold down one finger and scroll with another finger). It seems to me that this is new, but without a time machine I can't be sure.

becarpenter:
I meant to add that when I got home from travel, I realised that I've been scrolling that sidebar with the mouse wheel (without ever thinking about it), so this is nothing new. But with just the trackpad, it's a _very_ non-intuitive two-finger operation.

Isn't it possible to add a vertical scroll bar within the sidebar?

rjsparks:
I think @larseggert had tried and ran into issues, but we'll revisit.

larseggert:
The main issue is that the main view scrollbar and the sidebar scrollbar overlap. There may be a solution through adding padding, but that obviously wastes a bit of horizontal space.

becarpenter:
Can you make the sidebar float? 
There's an ancient stackoverflow conversation about something similar: https://stackoverflow.com/questions/19967352

becarpenter:
Thanks! I just noticed that this has been fixed with an additional scrollbar.

PR #7499, opened 2024-06-03
fix: Scroll metadata sidebar tab content

holloway:
An attempt at addressing https://github.com/ietf-tools/datatracker/issues/7390 which looks like,

**Chrome**
Screenshot from 2024-06-05 14-24-13

**Firefox**
Screenshot from 2024-06-05 14-24-50

For UX reasons just the tab content in the sidebar has been made scrollable. Is that ok? The UX reason are that if the sidebar had a full scrollbar it would look like the page itself had 2 scrollbars, whereas this is more clearly just making the tab content scrollable. This does mean that very short screens might not have enough visible scrolling area but such a small screen is probably mobile/tablet which needs a different responsive approach anyway.

Thoughts?

@larseggert @martinthomson

### DIFF
--- a/ietf/templates/doc/document_html.html
+++ b/ietf/templates/doc/document_html.html
@@ -205,7 +205,7 @@
                             HTML is available for this document.
                         </div>
                         {% endif %}
-                        <ul class="nav nav-tabs nav-fill small" role="tablist">
+                        <ul class="nav nav-tabs nav-fill small me-2" role="tablist">
                             <li class="nav-item" role="presentation" title="Document information">
                                 <button class="nav-link px-2"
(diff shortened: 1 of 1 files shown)


holloway:
About #e3cba5e ...some browsers/platforms don't allocate space for scrollbars (ie they have zero-width scrollbars that only float above the page) which means two scrollbars could render ontop of each other. This adds some additional spacing.

rjsparks:
This looks good to me - Martin and Lars have had time to comment. I'll merge this in and it will be part of the next release (which might be tomorrow).

rjsparks (approved):

PR #7510, opened 2024-06-06
feat: use surname/initials for author name

jennifer-richards:
This adds logic similar to what is used in xml2rfc's `TextWriter` to render an author name from the `initials` and/or `surname` fields when `fullname` is not present.

I've also updated places where we display the name to account for the possibility that the name is absent. If possible, I show the affiliation instead. This again matches what `TextWriter` does when rendering an RFC.

@russhousley

### DIFF
--- a/ietf/templates/submit/submitter_form.html
+++ b/ietf/templates/submit/submitter_form.html
@@ -11,12 +11,14 @@
 </p>
 {% load ietf_filters %}
 {% for author in submission.authors %}
-    <button type="button"
-           class="author btn btn-primary mb-3"
-           data-name="{{ author.name }}"
-           data-email="{% if author.email %}{{ author.email }}{% endif %}">
-        {{ author.name }}
-    </button>
+    {% if author.name %}
+        <button type="button"
+                class="author btn btn-primary mb-3"
+                data-name="{{ author.name }}"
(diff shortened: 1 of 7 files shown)


russhousley:
Looks good to me.

ietf/templates/submit/submitter_form.html
```diff
@@ -11,12 +11,14 @@ <h3 class="mt-5">Submitter</h3>
 </p>
 {% load ietf_filters %}
 {% for author in submission.authors %}
-    <button type="button"
-           class="author btn btn-primary mb-3"
-           data-name="{{ author.name }}"
-           data-email="{% if author.email %}{{ author.email }}{% endif %}">
-        {{ author.name }}
-    </button>
+    {% if author.name %}
```
rjsparks, line 14: if author.name is None for all authors in submission.authors, does the form fail?

rjsparks (commented):
One question inline before approving

ietf/templates/submit/submitter_form.html
jennifer-richards: No, these buttons are just present to quick-fill the actual form field via javascript. If there are no buttons, then the submitter has to fill in the fields manually.

rjsparks (approved):

PR #7523, opened 2024-06-11
refactor: adjust mail ingestion api

jennifer-richards:


ietf/api/views.py
```diff
@@ -662,9 +647,8 @@ def _err(code, text):
             iana_ingest_review_email(message)
         elif dest == "ipr-response":
             ipr_ingest_response_email(message)
-        elif dest == "nomcom-feedback":
-            year = payload["year"]
-            nomcom_ingest_feedback_email(message, year)
+        elif dest == "nomcom-feedback-2024":
```
rjsparks, line 650: why not parse the year out of the dest at this point?

rjsparks (approved):

PR #7573, opened 2024-06-21
feat: Status message

holloway:
Requirements:
* show a status message on all pages to users
* doesn't require JavaScript
* doesn't pollute Cloudflare cache with status messages (ie status messages can't be inline in the template, the browser has to compose the page)

This approach uses a non-JS fallback via `<noscript><iframe src=></iframe></noscript>`, and for JS users there's JS + API to show a message.

There isn't any design applied yet to either approach. The UX of this has not been decided.


rjsparks:
suggest this be closed now, and more work be against the feature branch at ietf-tools/datatracker.

PR #7585, opened 2024-06-21
fix: prune away unsused model and view

rjsparks:


rjsparks:
I ran black on the resource file.

jennifer-richards (approved):

PR #7628, opened 2024-07-03
feat: dev mode admin + refactor api init

jennifer-richards:
When in dev mode, applies styling to the Django admin site that resembles main site's dev mode. It also changes the admin title from "Django administration" to "Datatracker administration".

This introduces an `admin` app with an `apps.py`. This lets us override the title / header for the admin site without customizing more templates. In theory we could put the contents of the `ietf/admin/` directory into an `ietf/admin.py` instead, but this does not play well with TastyPie's api autodiscovery. 😡 

The CSS for this is a bit clunky - importing `ietf.css` to the admin does nasty things to its formatting. Instead, I've copied the explicit values for a few colors to try to keep the admin consistent with the overall scheme of the site. This is fragile and we could do it better, but I don't think it's worth much more effort.


jennifer-richards:
Converting to draft to figure out what to do about tests failing. They don't like an app with no `models.py` file. I'm going to revisit whether it's worth adding the customization this way.

jennifer-richards:
N.b., tests are now passing aside from the URL coverage

rjsparks (dismissed):

jennifer-richards:
Sorry for waffling about whether to leave this as a draft or not. I think that some of the pain I blamed TastyPie for might be self-induced. We have an `autodiscover()` method in `ietf.api.__init__` that is based on a now long stale version of Django's admin module autodiscovery method. I want to try bringing that up to date with modern Django's method and see if it lets us avoid having to promote the admin customization to its own app.

jennifer-richards:
Ok - I think this is ready. The last commit ([95ce3f8](https://github.com/ietf-tools/datatracker/pull/7628/commits/95ce3f843fe073617801a7a763fa5ba4d7396d1c)) brought the api init up to Django's spec for initializing things that pay attention to the app registry.

I've left the admin app as a fully fleshed out app. With the changes here, I _could_ combine the files in `ietf/admin/` into a single `admin.py` but I think we might as well follow the standard layout even if it's a minimal app.

rjsparks (approved):

PR #7659, opened 2024-07-10
feat: Site status message

holloway:
Features

* show a status message on all public pages
* doesn't require JavaScript
  * non-JS browsers use an `<iframe>` pointing to a route showing the latest status message. If no site message is active it shows "No site status message.". The iframe is a `<noscript><iframe src=></iframe></noscript>` so only browser without JavaScript have this fallback. 
  * JS-enabled browsers use [Vue NaiveUI: Notifications](https://www.naiveui.com/en-US/os-theme/components/notification) with a default design. Dismissed statuses are noted in browser `localStorage` (ie per device and browser) not per user, so dismissed statuses aren't remembered on a different device and will need to be dismissed again.
* doesn't pollute Cloudflare cache with status messages (ie status messages aren't rendered inline in each Django page as Cloudflare would cache outdated statuses... instead the browser composes the page with JavaScript+APIs or IFrames)
* Has an endpoint to redirect users to the latest status: `/status`
* Django and Playwright tests


ietf/status/views.py
```diff
@@ -35,6 +35,7 @@ def status_latest_html(request):
 
 def status_page(request, slug):
     status = get_object_or_404(Status, slug=slug)
+    print("status", status)
```
rjsparks: debug leak?

holloway:
Hey @rjsparks this is ready for another review

client/components/Status.vue
```diff
@@ -0,0 +1,83 @@
...
+
+            const isSameStatusPage = Boolean(document.querySelector(`[data-status-id="${status.id}"]`))
+
+            if(isSameStatusPage) {
+              console.info(`Not showing site status ${status.id} because we're on the target page`)
+              return
+            }
+
+            if(notificationInstances[status.id]) {
+              console.info(`Not showing site status ${status.id} because it's already been displayed`)
```
NGPixel, line 48: Change to console.debug?

client/components/Status.vue
```diff
@@ -0,0 +1,83 @@
+<script>
```
NGPixel, line 1: Use `<script setup>` instead of manually defining the component further down.

client/components/Status.vue
NGPixel, line 1: Various uses of semi-colons. Remove all.

client/shared/json-wrapper.js
```diff
@@ -0,0 +1,18 @@
+export const JSONWrapper = {
+  parse(jsonString, defaultValue) {
+    if(typeof jsonString !== "string") return defaultValue
```
NGPixel, line 3: Avoid single line if / return

NGPixel (dismissed):

ietf/status/models.py
```diff
@@ -0,0 +1,22 @@
+# Copyright The IETF Trust 2017-2020, All Rights Reserved
```
rjsparks, line 1: this is a new model copyright should be 2024 - same with all other new files. Modified files should have the date range extended to 2024

ietf/status/migrations/0001_initial.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 4.2.13 on 2024-06-24 04:24
```
rjsparks, line 1: Replace the Generated by with the trust copyright line

richsalz:
I wrote a tool for OpenSSL to do it automatically. Do you want me to open an issue to add something like that to the release process? https://github.com/openssl/tools/blob/master/release-tools/do-copyright-year

rjsparks:
Sure - split that into a separate issue so we don't distract from this PR too much.

ietf/status/migrations/0001_initial.py
rjsparks: Also, since this is a new app coming in, delete the migrations and regenerate them so that there's only a 0001_initial.

ietf/status/views.py
```diff
@@ -0,0 +1,50 @@
...
+from django.core.exceptions import FieldError
+from django.shortcuts import render, get_object_or_404
+from ietf.status.models import Status
+
+import debug                            # pyflakes:ignore
+
+def get_context_data():
+    status = Status.objects.order_by("-date").first()
+    if status is None or status.active == False:
+        return { "hasMessage": False }
```
rjsparks, line 16: This prevents us from staging the next message.
Consider instead:
```
    status = Status.objects.filter(active=True).order_by("-date").first()
```

ietf/status/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if status.slug is None:
+        raise FieldError("No slug generated. This shouldn't happen.")
```
rjsparks, line 19: Not sure this is a valid use of FieldError. In any case there's nothing to catch this, so it would result in a 500 being served.

It would be better to make the slug a required and unique field in the model with `null=False, blank-False, unique=True`, with  and let the ORM ensure it will never be empty,

ietf/status/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+    context = {
+        "hasMessage": True,
+        "id": status.id,
+        "slug": status.slug,
+        "title": status.title,
+        "body": status.body,
+        "url": f"/status/{status.slug}",
+        "date": status.date.isoformat(),
+        "by": status.by.name,
```
rjsparks, line 29: `by` seems to be inconsistently used? The admin could default it to the logged in user, but maybe we should remove it from the model?

ietf/templates/status/status.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
+{% load origin %}
+{% load ietf_filters static %}
+{% block content %}
+    {% origin %}
+    <h1 data-status-id="{{ status.id }}">
+        {% block title %} {{ status.title }} {% endblock %}
+    </h1>
+    <div>
```
rjsparks, line 9: Since the view that uses this will let you look at inactive slugs, it would be good to put something in the template that warns if the status is no longer active.

rjsparks (changes requested):
Very close, but I've marked a few things I think need to change.

ietf/status/views.py
holloway: Changed. This was intentional (ie only the latest status which must also be active is used) but happy to change it.

ietf/status/views.py
holloway: I've removed it from the templates / API response, but kept it in the model for auditing purposes. Happy to remove it if you'd prefer though.

ietf/templates/status/status.html
holloway: Good idea. I've added a Bootstrap 'badge' to the `<h1>` saying "inactive"... thoughts?

Screenshot from 2024-07-16 12-36-17

rjsparks: wfm

holloway:
@rjsparks @NGPixel ready for review again

holloway:
@NGPixel for context, Lars pointed out that routes using `document_html.html` (eg `/doc/html/rfc9547`) use a different base and so wouldn't get the status message. So this adds the status code to that template too. This is the first time that Vite's `client/embedded.js` has been used on that route. I hope that's ok to do.

jennifer-richards:
I think the failing URL test is caused by not covering the `api/v1/(?P<api_name>status)/` endpoint

ietf/status/views.py
```diff
@@ -0,0 +1,46 @@
...
+def status_page(request, slug):
+    sanitised_slug = slug.rstrip("/")
+    status = get_object_or_404(Status, slug=sanitised_slug)
+    return render(request, "status/status.html", context={
+        'status': status,
+        'status_page_html': markdown.markdown(status.page or ""),
+    })
+
+def status_latest_json(request):
+    return HttpResponse(json.dumps(get_context_data()), status=200, content_type='application/json')
```
jennifer-richards, line 40: There's a `JsonResponse` that might be more appropriate here

ietf/status/tests.py
```diff
@@ -0,0 +1,122 @@
...
+            title = "my title 1",
+            body = "my body 1",
+            active = True,
+            by = Person.objects.get(user__username='ad'),
+            slug = slug
+        )
+        status.save()
+
+        r = self.client.get(f'/status/{slug}/')
+        # with a Status it should redirect
```
jennifer-richards, line 101: Stale comment?

ietf/status/tests.py
```diff
@@ -0,0 +1,122 @@
...
+            body = "my body 1",
+            active = True,
+            by = Person.objects.get(user__username='ad'),
+            slug = slug,
+            page = f"# {test_string}"
+        )
+        status.save()
+
+        r = self.client.get(f'/status/{slug}/')
+        # with a Status it should redirect
```
jennifer-richards, line 118: Stale comment?

ietf/status/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+import json
+from django.http import HttpResponse, HttpResponseRedirect, HttpResponseNotFound
+from ietf.utils import markdown
+from django.shortcuts import render, get_object_or_404
+from ietf.status.models import Status
+
+import debug                            # pyflakes:ignore
+
+def get_context_data():
```
jennifer-richards, line 12: I think this needs a name more descriptive of its purpose

ietf/status/views.py
```diff
@@ -0,0 +1,46 @@
...
+    if status is None:
+        return { "hasMessage": False }
+
+    context = {
+        "hasMessage": True,
+        "id": status.id,
+        "slug": status.slug,
+        "title": status.title,
+        "body": status.body,
+        "url": f"/status/{status.slug}",
```
jennifer-richards, line 23: This ought to use the `reverse()` method (that we usually alias to `urlreverse()` in tests) to get the path to the correct view so that we're not baking paths into the code.

jennifer-richards (changes requested):
This is looking good. I have a few minor comments / change requests.

ietf/status/views.py
```diff
@@ -20,13 +20,13 @@ def get_context_data():
         "slug": status.slug,
         "title": status.title,
         "body": status.body,
-        "url": f"/status/{status.slug}",
+        "url": urlreverse('ietf.status.views.status_page', kwargs={ 'slug': status.slug }),
```
jennifer-richards, line 23: nit: double quotes instead of single

ietf/utils/test_runner.py
```diff
@@ -566,6 +566,14 @@ def ignore_pattern(regex, pattern):
                         covered.add(regex)
                         break
 
+            uncovered = set()
+            for regex, compiled, obj in patterns:
+                if regex not in covered:
+                    uncovered.add(regex)
+
+            if(len(uncovered) > 0):
+                print("Uncovered URLs", uncovered)
```
jennifer-richards, line 575: Did you mean to leave this in the commit? Maybe have it write to a file instead of stdout?

holloway: Yes, just to see some debug, although I will remove it again soon

holloway: Removed.

jennifer-richards (dismissed):
lgtm. Only nit I'm noticing (which I should have noticed before) is that we have copyright year ranges going back to 2016 or 2017 on newly added files. In some cases maybe this is appropriate if it's copied code, but I have generally started with only the current year in a new file if it's only standard boilerplate carried over from other places.

Don't know if we're consistent enough about this for it to be worthwhile to change (or if it's even appropriate).

holloway:
@jennifer-richards i've updated the ©

jennifer-richards (dismissed):

rjsparks (dismissed):

rjsparks (approved):

PR #7663, opened 2024-07-10
ci: run migrations via init container

jennifer-richards:
This splits migrations out of the datatracker pod into an init container and uses this to ensure that only datatracker pods (and not auth pods) run the migrations.

Alternatives would be to give the auth pod its own startup script that skips the migrations or to add conditionals to `datatracker-start.sh` that skip the migrations for the auth pods. I don't like introducing the conditionals down into the scripts, hence the approach here. (Though I don't like that we have to reproduce the container config for the datatracker container _again_ in the init container, either.)

### DIFF
--- a/dev/build/datatracker-start.sh
+++ b/dev/build/datatracker-start.sh
@@ -7,6 +7,7 @@ if ! ietf/manage.py migrate --skip-checks --check ; then
     echo "Unapplied migrations found, waiting to start..."
     sleep 5
     while ! ietf/manage.py migrate --skip-checks --check ; do 
+        echo "... still waiting for migrations..."
         sleep 5
     done
 fi
--- a/dev/build/datatracker-start.sh
+++ b/dev/build/datatracker-start.sh
@@ -3,8 +3,13 @@
 echo "Running Datatracker checks..."
 ./ietf/manage.py check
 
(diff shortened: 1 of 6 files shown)


dev/build/datatracker-start.sh
```diff
@@ -3,8 +3,13 @@
 echo "Running Datatracker checks..."
 ./ietf/manage.py check
 
-echo "Running Datatracker migrations..."
-./ietf/manage.py migrate --skip-checks --settings=settings_local
+if ! ietf/manage.py migrate --skip-checks --check ; then
+    echo "Unapplied migrations found, waiting to start..."
+    sleep 5
+    while ! ietf/manage.py migrate --skip-checks --check ; do
```
rjsparks, line 9: maybe echo something here so we know if were just stuck waiting for migrations?

rjsparks (dismissed):

dev/build/datatracker-start.sh
jennifer-richards: Actually, right after the `if` we `echo Starting datatracker...` so I think that's ok? I'll add a similar one to Celery since it doesn't have one

rjsparks (approved):

PR #7684, opened 2024-07-12
fix: check correct state machine when clearing status change ballots

### LINKED ISSUE #7335: Attempts to clear a status change ballots crash
```
AttributeError at /doc/status-change-mpls-rfc7506-to-historic/edit/clearballot/statchg/

'NoneType' object has no attribute 'slug'

Exception Location: | .../12.10.0/ietf/doc/views_ballot.py, line 405, in clear_ballot

```

cc: @cindymorgan

rjsparks:
Fixes #7335

### DIFF
--- a/ietf/doc/views_ballot.py
+++ b/ietf/doc/views_ballot.py
@@ -403,7 +403,7 @@ def clear_ballot(request, name, ballot_type_slug):
         by = request.user.person
         if close_ballot(doc, by, ballot_type_slug):
             create_ballot_if_not_open(request, doc, by, ballot_type_slug)
-        if doc.get_state('draft-iesg').slug == 'defer':
+        if doc.get_state("statchg").slug == "defer":
             do_undefer_ballot(request,doc)
         return redirect("ietf.doc.views_doc.document_main", name=doc.name)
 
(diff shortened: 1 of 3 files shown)


jennifer-richards (dismissed):

ietf/doc/views_ballot.py
```diff
@@ -403,7 +403,7 @@ def clear_ballot(request, name, ballot_type_slug):
         by = request.user.person
         if close_ballot(doc, by, ballot_type_slug):
             create_ballot_if_not_open(request, doc, by, ballot_type_slug)
-        if doc.get_state('draft-iesg').slug == 'defer':
+        if doc.get_state("statchg").slug == "defer":
```
jennifer-richards, line 406: This shoul have a guard against `doc.get_state()` returning `None`. (I think this is the cause of the failing test)

jennifer-richards (changes requested):

ietf/doc/views_ballot.py
rjsparks: hmm - that just exposes another subtle bug - the check should be against the ballot_type_slug, and if there is no state, the view had best do nothing (404 would be right I think).

jennifer-richards (approved):

PR #7685, opened 2024-07-13
ci: migrate to using secret instead of configmap

NGPixel:
- Use secret instead of configmap
- Move non-primary containers further down, so that the correct image is displayed in Portainer

### DIFF
--- a/k8s/secrets.yaml
+++ b/k8s/secrets.yaml
@@ -1,9 +1,9 @@
 apiVersion: v1
-kind: ConfigMap
+kind: Secret
 metadata:
-  name: django-config
-data:
-    # n.b., these are debug values / non-secret secrets
+  name: secrets-env
+type: Opaque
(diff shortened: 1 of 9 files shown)


k8s/secrets.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v1
-kind: ConfigMap
+kind: Secret
 metadata:
-  name: django-config
-data:
-    # n.b., these are debug values / non-secret secrets
+  name: secrets-env
```
jennifer-richards, line 4: I think this should be `dt-secrets-env`? (If I understand correctly, it's just for illustration)

NGPixel: No, as Kustomize will prefix it with `dt` upon creation.

jennifer-richards (approved):

PR #7689, opened 2024-07-15
fix: Ensure called from the docker directory

richsalz:
Found while naively trying to clean things per an email from Robert. :)


jennifer-richards (approved):
lgtm

PR #7702, opened 2024-07-17
feat: Groups menu overflow shadows

### LINKED ISSUE #7528: Menus mistakenly look complete when actually further content can be scrolled to
### Describe the issue

Selecting the Group menu and then hovering over Security shows a menu as per the attached screenshot.  This menu only shows some of the groups and can be scrolled to see more but due to two factors, most people will look at that and think this is the complete list of groups.  Those two factors are a) that the menu does not extend down as far as it can given the window size; and b) it does not display a scrollbar or some form of 'more below' indicator.

Recommend that the scrollbar is forced to display when an incomplete list is shown, or the bottom line of the menu has a down arrow or some other indicator that there are further entries below.

Screenshot 2024-06-13 at 10 32 31

### Code of Conduct

holloway:
To resolve #7528 this adds a `box-shadow` to indicate that more scrolling is available. [This is a common UX pattern](https://stevebauman.ca/adding-scroll-shadows-to-overflow-containers/), and it requires JavaScript.

Here's how it looks,
* Light mode: Screenshot from 2024-07-17 12-36-56
* Dark mode: Screenshot from 2024-07-17 12-37-16

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -1188,3 +1188,20 @@ blockquote {
     padding-left: 1rem;
     border-left: solid 1px var(--bs-body-color);
 }
+
+.overflow-shadows {
+    transition: box-shadow 0.5s;
+}
+
+.overflow-shadows--both {
+    box-shadow: inset 0px 21px 18px -20px var(--bs-body-color),
+                inset 0px -21px 18px -20px var(--bs-body-color);
+}
+
+.overflow-shadows--top-only {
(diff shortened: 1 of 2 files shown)


NGPixel (approved):

rjsparks:
There are a few other places where we have UI that has a scrollbar that could use this treatment - hopefully it generalizes and will be easy to apply in those spots (example, the various dropdowns on the search page).

PR #7715, opened 2024-07-20
fix: Explicitly set `executable_path` for Selenium

larseggert:
So it finds `geckodriver` again.

### DIFF
--- a/ietf/utils/jstest.py
+++ b/ietf/utils/jstest.py
@@ -30,7 +30,7 @@ if skip_selenium:
     print("     "+skip_message)
 
 def start_web_driver():
-    service = Service(log_output=f"{executable_name}.log", service_args=['--log-no-truncate'])
+    service = Service(executable_path=f"/usr/bin/geckodriver", log_output=f"{executable_name}.log", service_args=['--log-no-truncate'])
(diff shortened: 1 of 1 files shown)


ietf/utils/jstest.py
```diff
@@ -30,7 +30,7 @@
     print("     "+skip_message)
 
 def start_web_driver():
-    service = Service(log_output=f"{executable_name}.log", service_args=['--log-no-truncate'])
+    service = Service(executable_path=f"/usr/bin/geckodriver", log_output=f"{executable_name}.log", service_args=['--log-no-truncate'])
```
rjsparks, line 33: Since we already pulled out `executable_name`, we should use it here. Maybe pull out `executable_path` too.

rjsparks (changes requested):

ietf/utils/jstest.py
larseggert:
```suggestion
    service = Service(executable_path=f"/usr/bin/{executable_name}", log_output=f"{executable_name}.log", service_args=['--log-no-truncate'])
```

rjsparks (approved):

PR #7716, opened 2024-07-20
chore: Add additional log messages to directauth()

### LINKED ISSUE #7709: Add logging to directauth()
### Description

The `/directauth` API is being used by the IMAP server for authentication. It would be helpful if more messages were logged.

### Code of Conduct

microamp:
Resolves #7709

### DIFF
--- a/ietf/api/views.py
+++ b/ietf/api/views.py
@@ -429,6 +429,7 @@ def directauth(request):
                 data = None
 
         if raw_data is None or data is None:
+            log.log("Request body is either missing or invalid")
             return HttpResponse(json.dumps(dict(result="failure",reason="invalid post")), content_type='application/json')
 
         authtoken = data.get('authtoken', None)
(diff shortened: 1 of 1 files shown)


jennifer-richards (dismissed):

ietf/api/views.py
```diff
@@ -449,18 +452,21 @@ def directauth(request):
...
         # Note well that we are using user.username, not what was passed to the API.
-        if user_query.count() == 1 and authenticate(username = user_query.first().username, password = password):
+        user_count = user_query.count()
+        if user_count == 1 and authenticate(username = user_query.first().username, password = password):
             user = user_query.get()
             if user_query.filter(person__isnull=True).count() == 1: # Can't inspect user.person direclty here
                 log.log(f"Direct auth of personless user {user.pk}:{user.username}")
             else:
                 log.log(f"Direct auth: {user.pk}:{user.person.plain_name()}")
+            log.log(f"Direct auth success: {username}")
```
rjsparks, line 462: This is overly chatty - can we tighten it up so that we don't get two log lines as we go through this code?

rjsparks (changes requested):

ietf/api/views.py
microamp: Good point. I added the `Direct auth success` prefix to each successful response above.

rjsparks (approved):

PR #7723, opened 2024-07-21
fix: use area-acronym-at-the-time in proceedings

### LINKED ISSUE #7706: Area names are not correct for renamed areas in proceedings
### Describe the issue

I was looking back at older proceedings and looked at IETF 107. The areas listed here includes WIT, and do not list Transport (TSV). There appear the datatracker fails maintain/present correctly what area WGs belonged to at the particular point in time and meeting. This leads to strange presentation like for 107 where MASQUE WG was TSV, but WebTrans was ART at that time. And ART do exist as category due to other WGs. This appears to be common across all proceedings pages that are generated based on datatracker database. 

Example: https://datatracker.ietf.org/meeting/107/proceedings

### Code of Conduct

russhousley:
When generating IETF meeting proceedings, use the Area Acronym for each WG at the time of the meeting. Fixes #7706.

### DIFF
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -4130,6 +4130,13 @@ def organize_proceedings_sessions(sessions):
 
 def proceedings(request, num=None):
 
+    def area_and_group_acronyms_from_session(s):
+        area = s.group_parent_at_the_time()
+        if area == None:
+            area = s.group.parent
+        group = s.group_at_the_time()
+        return (area.acronym, group.acronym)
+
     meeting = get_meeting(num)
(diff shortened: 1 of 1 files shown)


rjsparks (approved):

jennifer-richards (approved):
Looks to me like it's working. It's adding 15-20% more queries to the proceedings calculation (for 119, it adds about 500 to the already very large query count of 2840 when rendering the page).

I suspect we can improve the cost of this change, but better to look at it in the big picture than to dig deep on it now.

PR #7724, opened 2024-07-21
fix: refactor api_new_meeting_registration. Fixes #7608

### LINKED ISSUE #7608: Registration API Bug: switching reg type results in multiple MeetingRegistration records
### Describe the issue

If a registrant switches registration type onsite -> remote, or vice-versa, a new MeetingRegistration record is created, resulting in two records for the person. The question is, how do we differentiate this scenario, switching registrations, from the, albeit rare, situation where someone does actually register for both online and remote.

### Code of Conduct

rpcross:
- change to use JSON payload
- handle multiple records in one request


rpcross:
I am working on tests for this. Using a JSON payload means the existing require_api_key decorator won't work. Assuming JSON payload is the way to go, are there other precedents for checking API key? Function can do it's own validation, or can update decorator.

jennifer-richards:
> are there other precedents for checking API key?

https://github.com/ietf-tools/datatracker/blob/b5ab4b66110ed0777dae170e87db9343876b2741/ietf/api/ietf_utils.py#L27-L46 is what we've started using for datatracker. This uses the X-Api-Key header to store a shared secret and keeps the auth concerns out of the API itself.

rjsparks:
@rpcross This may have fallen into the cracks because of 120?

rjsparks:
@rpcross - lets discuss this soon?

jennifer-richards (changes requested):
There are a few comments inline, but I have reservations about the strategy used for removing/updating/replacing `MeetingRegistration` records. As it's done, I'm pretty sure there are race conditions that can lead to corrupt registrations. E.g., if two identical requests to register a user are at  nearly the same time, two identical registrations may be created. This will happen if the later-running request handler executes line 334 (`regs = MeetingRegistration.objects.filter(meeting__number=number, email=email)`) before the earlier-running handler executes line 357 (`new_reg.save()`).

At minimum, we ought to put a uniqueness constraint on `MeetingRegistration` to guard against this. Not sure, but I think it'd need to be over `email`, `meeting`, and `reg_type`.

I think that constraint should be enough to avoid data corruption, but it would result in some API errors that were unnecessary. I.e., the two simultaneous identical requests would both "succeed" in that their work would be done as requested, but the later one would get back an error. It'd be very cool to rework the API using `get_or_create` (plus probably some additional subtlety handling) to avoid this, but I'm not sure it's necessary.

ietf/api/views.py
```diff
@@ -232,6 +233,153 @@ def err(code, text):
...
+    # handle cancelled. there will be only one record
+    if payload[0]['cancelled']:
+        reg = MeetingRegistration.objects.filter(
+            meeting__number=number,
+            email=email,
+            reg_type=payload[0]['reg_type'],
+            ticket_type=payload[0]['ticket_type']).first()
+        if reg:
+            reg.delete()
+        return HttpResponse('Success', status=202, content_type='text/plain')
```
rjsparks, line 326: I read this to mean that if you are notifying datatracker of cancelled registrations you will send one request per cancelled record - that is, if someone cancels both hackathon and regular meeting, you'll have to POST two different requests each with one record rather than send both cancellations in the same POST.

If that's correct, it would be good to error if the first payload indicates cancelled and there are more than one records in the payload.

ietf/api/views.py
```diff
@@ -232,6 +233,153 @@ def err(code, text):
...
+
+    # get person
+    person = Person.objects.filter(email__address=email).first()
+    if not person:
+        log.log(f"api_new_meeting_registration_v2 no Person found for {email}")
```
rjsparks, line 331: Do you want the registration system to _know_ that the registration it just sent doesn't match a person? Think about whether telling the person that just registered "Oh no - something went wrong with your registration" might move some of the person-fixup work we do at meetings to the period before the meeting.

ietf/api/views.py
```diff
@@ -232,6 +233,153 @@ def err(code, text):
...
+            ticket_type=registration['ticket_type'],
+            checkedin=registration['checkedin'])
+
+        # update any existing records if there are any
+        new_reg.pk = _safe_pop(pks)
+        if new_reg.pk:
+            log.log(f"Updating MeetingRegistration record for meeting:{meeting} email:{email}")
+        else:
+            log.log(f"New MeetingRegistration record for meeting:{meeting} email:{email}")
+        new_reg.save()
```
rjsparks, line 356: This confuses me, and looks error-prone - it could effectively swap pks on records depending on the order L334 returns objects. Why isn't this using get_or_create with defaults and using the created flag instead?

ietf/api/views.py
```diff
@@ -232,6 +233,153 @@ def err(code, text):
...
+                    person=person,
+                    defaults={
+                        "affiliation": registration["affiliation"],
+                        "origin": "registration"
+                    }
+                )
+
+    # delete any remaining records
+    if pks:
+        MeetingRegistration.objects.filter(pk__in=pks).delete()
```
rjsparks, line 378: If L333-L356 change, this will have to be adjusted to match

ietf/api/views.py
```diff
@@ -232,6 +233,153 @@ def err(code, text):
...
+        else:
+            log.log(f"New MeetingRegistration record for meeting:{meeting} email:{email}")
+        new_reg.save()
+
+        # removed account creation email. Registration requires Datatracker account
+
+        # handle nomcom volunteer
+        if registration['is_nomcom_volunteer'] and person:
+            try:
+                nomcom = NomCom.objects.get(is_accepting_volunteers=True)
```
rjsparks, line 363: Does the registration system check to see if a nomcom is accepting volunteers before offering the choice to someone?

rjsparks (changes requested):

ietf/api/views.py
jennifer-richards: Is there a reason to prefer the current model where a single person's registration is represented as several `MeetingRegistration` instances? The gymnastics in checking that the payload pertains only to a single email seems to be begging for a model that represents that as a single registration per email with fields representing the different options.

I think that'll greatly simplify the API implementation. It'll mean a migration, but I don't think that'll be terribly difficult.

jennifer-richards: Also, as far as I can tell there are currently no models with foreign keys into `MeetingRegistration` so it's not an going to break things now if PKs get swapped, but it'll do something in between cramping our style and causing serious bugs later.

ietf/api/tests.py
```diff
@@ -880,9 +881,112 @@ def test_api_new_meeting_registration(self):
...
+        r = self.client.post(url, data=json.dumps(bad_regs), content_type='application/json', headers={"X-Api-Key": "valid-token"})
+        self.assertEqual(r.status_code, 400)
+        #
+        # Test valid POST
+        r = self.client.post(url, data=json.dumps(regs), content_type='application/json', headers={"X-Api-Key": "valid-token"})
+        self.assertContains(r, "Success", status_code=202)
+        #
+        # Check record
+        reg = regs[0]
+        obj = MeetingRegistration.objects.get(email=reg['email'], meeting__number=reg['meeting'])
```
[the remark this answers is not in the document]
rpcross: Addressed in commit 06619ec675ee85a030e97ed9917bc205810fe9e0

ietf/api/views.py
rpcross: Notifications to Datatracker are triggered by the post_save signal on Registration. Regular onsite meeting registration and hackathon registration are each a Registration record, meaning a notification gets sent for each record.

Addressed in commit 4ae730675d89454bd9a0d7b86257f54a661fa808

ietf/api/views.py
rpcross: I left this in for completeness, but do we expect this to happen? Since Registration now requires the participant to be logged in via the Datatracker, the existence of Person should be largely assured. My feeling is that all auth issues at the last few meetings have been either 1) someone is trying to access meetecho without having registered, 2) duplicate person issues.

ietf/api/views.py
rpcross: > This confuses me, and looks error-prone - it could effectively swap pks on records depending on the order L334 returns objects. Why isn't this using get_or_create with defaults and using the created flag instead?

It certainly could swap pks. It is a brute force approach I chose for it's definitiveness. We can’t use get_or_create because there may be more than one existing registration record even for the same reg_type (two onsite one-day regs for example) AND the reg_type can change, switch from onsite to remote, so we’d be updating a record with a different reg_type. The simplest approach is to delete and re-create, or fully overwrite as I have done. Other approaches require more complexity. The following might work. Assuming that any notification will include all existing registrations for a given participant, and that a single registration save triggers the notification than it should be true that only one registration is added or was changed. If payload == 1 we can use get_or_create. (Except if payload == 1 and 2 already exist it means we missed a cancel somewhere). If payload > 1 we know that only one is new or updated and the rest should match existing records. We can do the work to determine which record doesn't match existing. This is a more complex procedure but would result in preserving the record PKs.

rpcross: > Is there a reason to prefer the current model where a single person's registration is represented as several `MeetingRegistration` instances? The gymnastics in checking that the payload pertains only to a single email seems to be begging for a model that represents that as a single registration per email with fields representing the different options.
> 
> I think that'll greatly simplify the API implementation. It'll mean a migration, but I don't think that'll be terribly difficult.

We actually started with one MeetingRegistration record per person and had a reg_type field that was text and contained comma separated types, "onsite, hackathon_onsite", which proved problematic to deal with so we change the model to match registration, which could be multiple registration records per person. The current model seems appropriate.

                  <- meeting 120 onsite week pass
person A   <- meeting 121 onsite one day pass
                  <- meeting 121 remote week pass

Or maybe breakout another table?

person A <- meeting 121 <- onsite one-day
                                           <- remote week pass

ietf/api/views.py
rpcross: It does not. It's another level of integration that could be added. My understanding was that AMS would coordinate with the nomcom chair. When prepping to open registration if nomcom is accepting volunteers I would ensure the volunteer question appears on the reg form.

rpcross:
Update:
The new proposal is to return to having a single MeetingRegistration record per person per meeting. The various registrations a single person can have for one meeting will be moved to a related object like this:

class MeetingRegistrationDetail(models.Model):      # name?
    meeting_registration = ForeignKey(MeetingRegistration)
    reg_type = ForeignKey(RegistrationTypeName)
    ticket_type = ForeignKey(RegistrationTicketTypeName, blank=True, null=True)    # there are older registrations where this is empty

And use a custom object manager like this:

class MeetingRegistrationManager(models.Manager):
    def onsite(self):
        return self.get_queryset().filter(meetingregistrationdetail__reg_type__slug='onsite')
    def remote(self):
	# don't count people that also have a onsite registration
        return self.get_queryset().filter(meetingregistrationdetail__reg_type__slug='remote').exclude(meetingregistrationdetail__reg_type__slug='onsite')

And queries like this:
MeetingRegistration.objects.onsite().filter(meeting__number=121)

We could combine reg_type and ticket_type, but the most frequent queries are those for specific reg_type “onsite” or “remote” so it seems reasonable to keep them separate. Are there better names for new classes, fields above?

Considerations:
Should we keep “checkedin” and “affialiation” on the MeetingRegistration record? Checkedin would only be set from a reg_type=onsite record. Hackathon or remote registrations would not affect this field. In theory you could have a different affiliation value for each registration, maybe prefer value from onsite over remote, and regular onsite|remote over hackathon, and populated over empty?

For simplicity is it acceptable to clear and recreate the related objects on update?

The above implementation is most flexible. Allowing the capture of all registration types currently represented, and new ones with the addition of new name types. Would a simpler approach of using boolean fields has_onsite, has_remote (has_hackathon_onsite, has_hackathon_remote) be sufficient. Note, you couldn’t represent two onsite one-day passes in that implementation.

Thoughts?

rjsparks:
>Should we keep “checkedin” and “affialiation” on the MeetingRegistration record?

We need checkedin _somewhere_. I think it's fine to be on the MeetingRegistration record - we can make it clear in the model that it means "picked up a badge onsite at the meeting".

I think we should keep "affiliation", but need to read the code to see if the datatracker is using it. Keeping it on MeetingRegistration makes sense.

jennifer-richards:
> Update: The new proposal is to return to having a single MeetingRegistration record per person per meeting. The various registrations a single person can have for one meeting will be moved to a related object like this:
> 

I like the concept. A couple key things to watch are, 1) that this is simplifying our lives, not just mooshing complexity/bugs around; and 2) that we're not replicating parts of the registration app that shouldn't leak into the datatracker.


> class MeetingRegistrationDetail(models.Model): # name? meeting_registration = ForeignKey(MeetingRegistration) reg_type = ForeignKey(RegistrationTypeName) ticket_type = ForeignKey(RegistrationTicketTypeName, blank=True, null=True) # there are older registrations where this is empty
> 

I guess `RegistrationTicketTypeName` includes things like "hackathon", "week", "day"?

Name-wise, I like the ticket metaphor and would be inclined to call it a `MeetingRegistrationTicket` and change `RegistrationTypeName` to `AttendanceTypeName` or something (plus a matching change to `reg_type`). 

It might be too early for this level of minutiae, but I'd encourage avoiding `blank=True, null=True` on the `ticket_type` field. Instead, use an `unknown` or some other type name that's marked as `used=False` and is not an option for new work. This will be closer to a guarantee we don't start creating incomplete records in the future.

> And use a custom object manager like this:
> 
> class MeetingRegistrationManager(models.Manager): def onsite(self): return self.get_queryset().filter(meetingregistrationdetail__reg_type__slug='onsite') def remote(self): # don't count people that also have a onsite registration return self.get_queryset().filter(meetingregistrationdetail__reg_type__slug='remote').exclude(meetingregistrationdetail__reg_type__slug='onsite')
> 
> And queries like this: MeetingRegistration.objects.onsite().filter(meeting__number=121)
> 

Checking an assumption, do we definitely need to support both onsite and remote registrations at the same meeting? I'd say we do either if we know we need it and/or if the reg system supports it, just so the datatracker side is expressive enough to capture everything. If not, life would be easier if the `reg_type` could move onto the `MeetingRegistration` model.

> We could combine reg_type and ticket_type, but the most frequent queries are those for specific reg_type “onsite” or “remote” so it seems reasonable to keep them separate. Are there better names for new classes, fields above?
> 

As noted above, I think you're describing an attendance type and it should be labeled as such instead of the more vague "registration type".  The business logic about what combinations make sense belongs on the reg side and datatracker should be able to represent whatever reg says and not second guess its records.

> Considerations: Should we keep “checkedin” and “affialiation” on the MeetingRegistration record? Checkedin would only be set from a reg_type=onsite record. Hackathon or remote registrations would not affect this field. In theory you could have a different affiliation value for each registration, maybe prefer value from onsite over remote, and regular onsite|remote over hackathon, and populated over empty?
> 

I don't know for sure about affiliation, but my instinct says it is a `MeetingRegistration` property that applies to the person / meeting combination. Our records afaik only show one affiliation per person at each meeting. I can imagine why we might want separate affiliations, but I wouldn't add it here unless it's something supported by the reg side that we need to allow for.

I think checkedin should be tied to the "ticket" that was checked in, not to the whole `MeetingRegistration`. For day passes, the datatracker needs to be able to tell that a day pass is active and to which day it's being applied.

> For simplicity is it acceptable to clear and recreate the related objects on update?
> 

This depends on what the datatracker is doing with the data. It'd be _nice_ if they didn't move around, but might not be a hard requirement. I think we'll want to be able to refer at least to MeetingRegistrations so those should be stable. For things below that, we mostly need to be clear about which things we're allowed to point at with FKs and be careful about deletion policies. We don't want to get into a situation where the datatracker prevents reg from accurately recording the true registration data.

> The above implementation is most flexible. Allowing the capture of all registration types currently represented, and new ones with the addition of new name types. Would a simpler approach of using boolean fields has_onsite, has_remote (has_hackathon_onsite, has_hackathon_remote) be sufficient. Note, you couldn’t represent two onsite one-day passes in that implementation.
> 

I prefer the better-normalized approach with the additional model to a raft of boolean flags. Not being able to represent multiple one-day passes makes it a non-starter.

jennifer-richards:
Robert posted his reply while I was writing my tome. His comment (and an out-of-band discussion) convinced me that the suggestion I have about putting "checkedin" on the "ticket" instances is really a different concept.

rpcross:
This PR is replaced by https://github.com/ietf-tools/datatracker/pull/8408

jennifer-richards:
Closing as it's been superseded

PR #7729, opened 2024-07-21
fix: Send create user email for password resets where we have an email and person, but no user.

### LINKED ISSUE #6458: Create Users on password reset requests
We have people who for whatever reason had a Person record with no User record created decades ago, usually because they were a co-author on a draft. Now when they want to interact with the datatracker (using the same email that was associated with that Person decades ago), the password reset link says it sent something, but it didn't send anything because there was no User.

Instead, create a User in this situation. Tie in the logic that looks for possible duplicate persons and send the warning message to the secretariat (that code already exists). 

The security posture of the result should be the same as a password reset for a Person that already has a User - proof of control of the email address.

(If that turns out to be problematic, change this issue to tell the person that they need to contact the secretariat to establish a new password instead).

ThisIsMissEm:
I think this fixes: https://github.com/ietf-tools/datatracker/issues/6458

I've tested manually by deleting my own `auth_user` database entry (and associated entries in `person_historicalperson` and `person_historicalemail`) to simulate a "missing user but known email and person" condition.

I'm not sure how the codebase is tested, since I'm not usually a python developer, however `ietf/manage.py test --settings=settings_test` seems to run fine. Though I do think it'd be good to have some sort of test coverage for exactly this case, since it seems like it's a pretty common issue for new editors/authors to encounter when collaborating with an existing IETF member on an I-D.

Sure, we could use a different email template to be a bit more specific about onboarding, but the overall goal I think here is to get the person setup with an account, and the create account flow does exactly that.

Whilst working on this, I did notice that from within the docker container setup, `Site.objects.get_current().domain` gives `datatracker.ietf.org` instead of `localhost:8000` like I'd have expected, so in the logs the links appeared as:

```
https://datatracker.ietf.org/accounts/reset/confirm/eyJ1c2VybmFtZSI6ImVtZWxpYUBicmFuZGVkY29kZS5jb20i....YCR3KmkLDZw2ArA_lPg/
```

Instead of:

```
http://localhost:8000/accounts/reset/confirm/eyJ1c2VybmFtZSI6ImVtZWxpYUBicmFuZGVkY29kZS5jb20i....YCR3KmkLDZw2ArA_lPg/
```

I also happen to have [Mailpit](https://mailpit.axllent.org/) running in my development environment, on port 2025, however the emails were just logged to the console; I assume this is just a standard development thing in python, but it could be neat to say "actually, yes, send all emails to this test email server"

### DIFF
--- a/ietf/ietfauth/views.py
+++ b/ietf/ietfauth/views.py
@@ -491,8 +491,11 @@ def password_reset(request):
             if not user:
                 # try to find user ID from the email address
                 email = Email.objects.filter(address=submitted_username).first()
-                if email and email.person and email.person.user:
-                    user = email.person.user
+                if email and email.person:
+                    if email.person.user:
+                        user = email.person.user
+                    else: 
+                        send_account_creation_email(request, email.email_address())
 
             if user and user.person.email_set.filter(active=True).exists():
                 data = {
--- a/ietf/ietfauth/views.py
+++ b/ietf/ietfauth/views.py
@@ -495,8 +495,15 @@ def password_reset(request):
                     if email.person.user:
                         user = email.person.user
                     else: 
-                        send_account_creation_email(request, email.email_address())
-
+                        # Create a User record with this (conditioned by way of Email) username
+                        # Don't bother setting the name or email fields on User - rely on the
+                        # Person pointer.
+                        user = User.objects.create(
+                            username=email.address.lower(), 
+                            is_active=True,
+                        )
+                        email.person.user = user
+                        email.person.save()
             if user and user.person.email_set.filter(active=True).exists():
                 data = {
                     'username': user.username,
--- a/ietf/ietfauth/tests.py
+++ b/ietf/ietfauth/tests.py
@@ -542,8 +542,8 @@ class IetfAuthTests(TestCase):
         self.assertEqual(len(outbox), 1)
         lastReceivedEmail = outbox[-1]
         self.assertIn(email.address, lastReceivedEmail.get('To'))
-        self.assertTrue(lastReceivedEmail.get('Subject').startswith('Confirm registration at'))
-        self.assertContains(r, 'We have sent you an email with instructions', status_code=200)
+        self.assertTrue(lastReceivedEmail.get('Subject').startswith("Confirm password reset"))
+        self.assertContains(r, "Your password reset request has been successfully received", status_code=200)
 
     def test_review_overview(self):
         review_req = ReviewRequestFactory()
--- a/ietf/ietfauth/tests.py
+++ b/ietf/ietfauth/tests.py
@@ -527,6 +527,24 @@ class IetfAuthTests(TestCase):
         self.assertIn(secondary_address, to)
         self.assertNotIn(inactive_secondary_address, to)
 
+    def test_reset_password_without_user(self):
+        """Reset password using email address for person without a user account"""
+        url = urlreverse('ietf.ietfauth.views.password_reset')
+        email = EmailFactory()
+        person = email.person
+        # Remove the user object from the person to get a Email/Person without User:
+        person.user = None
+        person.save()
+        # Remove the remaining User record, since reset_password looks for that by username:
+        User.objects.filter(username__iexact=email.address).delete()
+        empty_outbox()
+        r = self.client.post(url, { 'username': email.address })
+        self.assertEqual(len(outbox), 1)
+        lastReceivedEmail = outbox[-1]
+        self.assertIn(email.address, lastReceivedEmail.get('To'))
+        self.assertTrue(lastReceivedEmail.get('Subject').startswith('Confirm registration at'))
+        self.assertContains(r, 'We have sent you an email with instructions', status_code=200)
+
     def test_review_overview(self):
         review_req = ReviewRequestFactory()
         assignment = ReviewAssignmentFactory(review_request=review_req,reviewer=EmailFactory(person__user__username='reviewer'))


ThisIsMissEm:
I just noticed the various tests from https://github.com/ietf-tools/datatracker/blob/main/ietf/ietfauth/tests.py#L397 (which I somehow missed before — I'm definitely not an experienced python developer)

I would've expected at least `test_reset_password_address_handling` or `test_reset_password_without_username` to fail, given this change, but they didn't, and I'm not sure why, unless `PersonFactory()` automatically creates a user record? But I don't think that's what those are doing?

rjsparks:
Thanks for the PR!

Lots of questions above - we'll answer as best we can as the meeting allows.

re: Mailpit (or any other thing) - when in development mode (what you'd be in while running `runserver`, mail goes to `localhost` port 2025. You can change that in settings_local.py by changing:
```
>>> settings.EMAIL_HOST
'localhost'
>>> settings.EMAIL_PORT
2025
```
Note that it's localhost from _inside_ the container's perspective - you can map ports out to your host machine to get to services running there.

If you're using the `docker/run` command, it starts a debuging python smtp server that just prints whatever it gets onto stdout. You can modify that in docker-init.sh (which is created from https://github.com/ietf-tools/datatracker/blob/aa36f481e13c99bc0865282a9e890e6b049cdb98/docker/scripts/app-init.sh#L104

PersonFactory _does_ indeed create a User. The new tests can say `PersonFactory(user=None)`. See https://github.com/ietf-tools/datatracker/blob/aa36f481e13c99bc0865282a9e890e6b049cdb98/ietf/person/factories.py#L70

Again - busy week with meeting - we'll look into the proposed solution and suggest good ways to test it soon.

ThisIsMissEm:
@rjsparks actually, I just tried `PersonFactory(user=None)` and PersonFactory isn't happy at all because it assumes the existence of a user, so you get several stacktraces like:

```
  File "/workspace/ietf/ietfauth/tests.py", line 534, in test_reset_password_without_user
    person = PersonFactory(user=None)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/base.py", line 40, in __call__
    return cls.create(**kwargs)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/base.py", line 528, in create
    return cls._generate(enums.CREATE_STRATEGY, kwargs)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/django.py", line 121, in _generate
    return super()._generate(strategy, params)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/base.py", line 465, in _generate
    return step.build()
  File "/home/dev/.local/lib/python3.9/site-packages/factory/builder.py", line 270, in build
    step.resolve(pre)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/builder.py", line 211, in resolve
    self.attributes[field_name] = getattr(self.stub, field_name)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/builder.py", line 356, in __getattr__
    value = value.evaluate_pre(
  File "/home/dev/.local/lib/python3.9/site-packages/factory/declarations.py", line 67, in evaluate_pre
    return self.evaluate(instance, step, context)
  File "/home/dev/.local/lib/python3.9/site-packages/factory/declarations.py", line 119, in evaluate
    return self.function(instance)
  File "/workspace/ietf/person/factories.py", line 71, in <lambda>
    name = factory.LazyAttribute(lambda p: normalize_name('%s %s'%(p.user.first_name, p.user.last_name)))
AttributeError: 'NoneType' object has no attribute 'first_name'
```

rjsparks:
ok - then after you have something like
`p = PersonFactory...`, do a
```
p.user = None
p.save()
```
Other things may complain

ThisIsMissEm:
Okay, finally managed to get the tests fully working; I had to use EmailFactory instead of PersonFactory, but that still creates a Person so, I don't know why PersonFactory errored out continously.

ietf/ietfauth/views.py
```diff
@@ -491,8 +491,11 @@ def password_reset(request):
             if not user:
                 # try to find user ID from the email address
                 email = Email.objects.filter(address=submitted_username).first()
-                if email and email.person and email.person.user:
-                    user = email.person.user
+                if email and email.person:
+                    if email.person.user:
+                        user = email.person.user
+                    else: 
+                        send_account_creation_email(request, email.email_address())
```
ThisIsMissEm, line 498: I've manually asserted that if I remove these lines, then the test case fails too.

rjsparks:
Unfortunately, I think we're going to need a different email round-trip sequence than reusing `send_account_creation_email` or we're going to need to carefully replumb that message and the handlers for the links it contains.

I walked through the rest of the process manually, and the result was very confusing (even if it mostly ended up with the right result).

The mail that gets emitted looks like this:
```
MIME-Version: 1.0
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: 7bit
From: IETF Secretariat <ietf-secretariat-reply@ietf.org>
To: <foo@bar.baz>
Subject: Confirm registration at datatracker.ietf.org
X-Test-IDTracker: yes
X-IETF-IDTracker: 12.19.0-dev
Auto-Submitted: auto-generated
Precedence: bulk
Message-ID: <172168088514.42273.346891258396035602@a675050b3ff6>
Date: Mon, 22 Jul 2024 13:41:25 -0700
X-Peer: ::1


Hello,

We have received an account creation request for
foo@bar.baz at datatracker.ietf.org.

If you do not already have datatracker account, please go to the following
link and follow the instructions there in order to set a new password for the
foo@bar.baz account:

   https://datatracker.ietf.org/accounts/create/confirm/bunch-of-token-hash-bits-here/

This link will expire in 3 days.

If you already have a datatracker account, please go to the following link
instead, and add the new email foo@bar.baz to your
existing account:

   https://datatracker.ietf.org/accounts/profile/


Best regards,

	The datatracker login manager service
	(for the IETF Secretariat)


```

Which is disconnected from reality in this case - the "account" exists as far as the end user is concerned, so following the account exists instructions results in the wrong thing happening. Following the "if you don't have an account" path almost results in the right thing, but the user gets a form where they have to provide their name and ascii-name and a password. The provided name and ascii-name don't get used afaict.

Also, we're got this `send_account_creation_exists_email` path that can get used, that is asking people to do something that won't work, even with this fix. That's going to need to be reconciled while straightening this out.

@ThisIsMissEm - please feel free to continue to push if you have the energy, but if not, we can pick this up and finish it in a bit. We'll definitely use what you've provided so far.

ThisIsMissEm:
@rjsparks yeah, I totally agree that it's almost right but not quite — though, me being a new contributor + not really a python/django developer, I didn't think it'd be wise for me to setup a completely new flow, due to the security implications of that, hence reusing the existing flow to try to do a minimal "fix this so new comers who are on I-D's as co-authors don't get stuck" sense.

rjsparks:
The push above takes the solution in a slightly different direction, but I think it's the better path.

ietf/ietfauth/views.py
```diff
@@ -491,9 +491,19 @@ def password_reset(request):
...
+                if email and email.person:
+                    if email.person.user:
+                        user = email.person.user
+                    else: 
+                        # Create a User record with this (conditioned by way of Email) username
+                        # Don't bother setting the name or email fields on User - rely on the
+                        # Person pointer.
+                        user = User.objects.create(
+                            username=email.address.lower(), 
+                            is_active=True,
```
ThisIsMissEm, line 503: Doesn't this lack the verification of the email though?

jennifer-richards: To my read it's ok. It does create a `User` without verification, but it's a `User` that should already have existed and login won't be possible until they use the reset email which proves they received it.

(I haven't reviewed fully yet)

rjsparks: That's correct  - the user's password will initially be something unusable. Only a successful response to the password reset email (which serves as proof of control of the email) will make the user object usable.

jennifer-richards (changes requested):
As much a question as a change request - see inline.

Though I think we should remove the comment about "The form validation checks that a matching User exists" because that part of the comment is outright false after the last round of refactoring. (The form no longer does anything beyond email syntax validation.)

ietf/ietfauth/views.py
```diff
@@ -491,9 +491,19 @@ def password_reset(request):
             if not user:
                 # try to find user ID from the email address
                 email = Email.objects.filter(address=submitted_username).first()
-                if email and email.person and email.person.user:
-                    user = email.person.user
-
+                if email and email.person:
```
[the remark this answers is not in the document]
rjsparks: Nice rock you looked under  🫤 

The "there are no useable addresses" case _does_ exist, and we should handle it differently. I'm on the fence about whether it should be in a separate PR though. I'm leaning towards accepting this one and merging it (it makes the "no useable address" case no worse, and handling that situation with a different issue/PR. Would that work for you?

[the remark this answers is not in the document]
jennifer-richards: Yeah, that works.

jennifer-richards (approved):

PR #7754, opened 2024-07-26
feat: track docs that fail pdfize task

jennifer-richards:
Haven't thought hard about return values, this is mostly to get feedback on the approach


rjsparks:
General approach looks good. I'd like to talk about when/how we tune the "that took too long" decision.

rjsparks:
When should we pursue this further?

jennifer-richards:
> When should we pursue this further?

"soon" but I think it's out of the critical path since we've limited the endpoint to logged-in users

Michealinc (commented):
Explict

Michealinc (commented):
Explict

PR #7760, opened 2024-07-29
fix: Document type badges

holloway:
See #7475


ietf/doc/templatetags/document_type_badge.py
```diff
@@ -0,0 +1,23 @@
...
+
+@register.simple_tag
+def document_type_badge(doc, snapshot):
+    if doc.type == "rfc":
+        return render_to_string(
+            "doc/badge/doc-badge-rfc.html",
+            {"doc": doc, "snapshot": snapshot},
+        )
+
+    return render_to_string(
```
rjsparks, line 20: Python nit: this really is an else.

ietf/doc/templatetags/document_type_badge.py
```diff
@@ -0,0 +1,23 @@
...
+        "doc/badge/doc-badge-other.html",
```
rjsparks, line 21: this should be called doc-batch-draft - it's only badging Internet-drafts.

It might be worth using this kind of shape:

```
if doc.type == "rfc":
   ...
elif doc.type == "draft":
  ...
else:
   return no badge, raise an error in dev/test, and log a warning in production.
```

ietf/templates/doc/badge/doc-badge-rfc.html
```diff
@@ -0,0 +1,13 @@
...
+{% load static %}
+{% load ietf_filters %}
+{% load person_filters %}
+{% origin %}
+
+<span class="text-success" style="border: dashed 3px green">RFC
+    {% if not document_html %}
+    - {{ doc.std_level }}
+    {% else %}
+        <span class="badge rounded-pill badge-{% if not snapshot %}{{ doc|std_level_to_label_format }}{% else %}draft{% endif %}">{{ doc.std_level }}</span>
```
rjsparks, line 11: Seems wrong to fall back to a class of "badge-draft". Probably needs a new class to fall back to.

ietf/templates/doc/badge/doc-badge-other.html
```diff
@@ -0,0 +1,16 @@
+{% load origin %}
+{% load static %}
+{% load ietf_filters %}
+{% load person_filters %}
+{% origin %}
+{# Non-RFC #}
+
+{% if doc.became_rfc %} 
+    <div{% if document_html %} class="alert alert-warning small"{% endif %} style="border: dashed 3px green">This is an older version of an Internet-Draft that was ultimately published as <a href="{% if document_html %}{% url 'ietf.doc.views_doc.document_html' name=doc.became_rfc.name %}{% else %}{% url 'ietf.doc.views_doc.document_main' name=doc.became_rfc.name %}{% endif %}">{{doc.became_rfc.name|prettystdname}}</a>.</div>
```
rjsparks, line 9: why introduce new styling inline? Can this be moved to the css?

ietf/templates/doc/document_info.html
```diff
@@ -59,16 +54,6 @@
                         {% if submission %}({{ submission|safe }}){% endif %}
```
rjsparks, line 54: We should look to make sure, but off the cuff, I don't think submission will be (or at least should be) non None if draft.type == "rfc"

rjsparks (changes requested):

ietf/templates/doc/document_info.html
holloway: Sorry I don't know how to interpret the code changes for this. Are you suggesting that the code should look like,
```
{% if doc.type_id == "rfc" %}({{ submission|safe }}){% else %}something else?{% endif %}
```

ietf/templates/doc/badge/doc-badge-rfc.html
holloway: Have chosen 'badge-generic' as the fallback. Naming improvements welcome :)

ietf/templates/doc/document_info.html
rjsparks: well, it would have been the other way around - the submission type really applies to documents of type `draft` not of type `rfc`. I read through the code and you can leave this line as is.

That said, what the code is doing is questionable (this PR shouldn't change it, but I'll capture things here to share knowledge): document_main has a conditional branch for documents of type `rfc` that renders document_rfc.html with `submission` in its context. It computes the value for `submission` here: https://github.com/ietf-tools/datatracker/blob/ff8898186ba06072d0e226a6c88bb6e371cc3dee/ietf/doc/views_doc.py#L316-L329 using the RFCs group instead of the draft the RFC came from. This value is rendered next to the draft name if the document came from a draft. For example, for RFC8989 it's the word "individual" at:
image

Notice that for an April 1 RFC (which does not come from a draft), the `submission` value is not rendered at all. (See RF9564).

What should really happen is that submission should be calculated from the draft the rfc came from. It is likely at the moment that the rfc _always_ inherits the information used to calculate `submission` from the draft with no change - but someday that may not be true.

rjsparks:
@holloway : You've addressed my comments so far. What remains before this PR moves out draft status?

rjsparks (dismissed):

rjsparks:
@holloway (or anyone else puzzled by the test failures earlier): at [1c3fd00](https://github.com/ietf-tools/datatracker/pull/7760/commits/1c3fd009a705bfbaeba1cbf3b67add9aa51e0c9e) - the issue that caused test fails reporting an invalid document type of "Draft" was the compare and render using type vs type_id:

```
>>> from ietf.doc.models import Document
>>> d = Document.objects.filter(type_id="draft").last()
>>> type(d.type)
<class 'ietf.name.models.DocTypeName'>
>>> d.type=="draft"
False
>>> str(d.type)
'Draft'
```

One bit of django queryset syntactic "assistance" to be aware of. The following are equivalent, because the ORM does helpful work - assume `dt = d.type` from above (that is, dt is `<DocTypeName: Draft>`)

```
Document.objects.filter(type_id="draft")
Document.objects.filter(type="draft")
Document.objects.filter(type=dt)
```
All of which turn into this SQL
```
>>> str(Document.objects.filter(type=dt).query)
'SELECT "doc_document"."id", "doc_document"."time", "doc_document"."type_id", "doc_document"."title", 
"doc_document"."stream_id", "doc_document"."group_id", "doc_document"."abstract", "doc_document"."rev", 
"doc_document"."pages", "doc_document"."words", "doc_document"."intended_std_level_id", 
"doc_document"."std_level_id", "doc_document"."ad_id", "doc_document"."shepherd_id", "doc_document"."expires", 
"doc_document"."notify", "doc_document"."external_url", "doc_document"."uploaded_filename", "doc_document"."note", 
"doc_document"."internal_comments", "doc_document"."rfc_number", "doc_document"."name" FROM "doc_document" 
WHERE "doc_document"."type_id" = draft'
```

ietf/templates/doc/badge/doc-badge-other.html
rjsparks: I think we still have inline styling in this branch and that it's being applied in cases that aren't intended - right now the tests are failing complaining about the inline styling, and because instead of this on main:

image

we're seeing this on the PR branch:

image

This is an active draft, so why the dashed lines, and what happened to the `(stir WG)` link?

rjsparks: That `(stir WG)` link is what's passed in as `submission`

ietf/doc/templatetags/document_type_badge.py
```diff
@@ -0,0 +1,32 @@
...
+from warnings import warn
+
+# from ietf.group.models import Group
+# from ietf.name.models import GroupTypeName
+
+register = template.Library()
+
+
+@register.simple_tag
+def document_type_badge(doc, snapshot):
```
rjsparks, line 15: `submission` from the context is being lost here I think

ietf/templates/doc/document_html.html
```diff
@@ -185,13 +187,7 @@
...
                                 {% endif %}
                                 <br>
-                                <span class="badge rounded-pill badge-{% if not snapshot %}{{ doc|std_level_to_label_format }}{% else %}draft{% endif %}">
-                                    {% if not snapshot %}
-                                        {{ doc.std_level }}
-                                    {% else %}
-                                        Internet-Draft
-                                    {% endif %}
-                                </span>
+                                {% document_type_badge doc snapshot%}
```
rjsparks, line 190: this also needs the context?

rjsparks (changes requested):

rjsparks (dismissed):

rjsparks:
Asking for a second review given the twisty passages

jennifer-richards (changes requested):
Just a couple minor things

ietf/doc/templatetags/document_type_badge.py
```diff
@@ -0,0 +1,33 @@
...
+        return render_to_string(
+            "doc/badge/doc-badge-draft.html",
+            context,
+        )
+    else:
+        error_message = f"Unsupported document type {doc.type_id}."
+        if settings.SERVER_MODE != 'production':
+            raise ValueError(error_message)
+        else:
+            warn(error_message)
```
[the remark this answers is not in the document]
jennifer-richards: Well, I was imagining keeping the fallback to "" when the assertion doesn't interrupt, which would do the same thing as here. The assertion rules are slightly different than a test for production mode, which might or might not be preferable, and it avoids adding more places we look at `SERVER_MODE`

ietf/doc/templatetags/document_type_badge.py
```diff
@@ -0,0 +1,33 @@
+# Copyright The IETF Trust 2015-2020, All Rights Reserved
+from django import template
+from django.conf import settings
+from django.template.loader import render_to_string
+
+from warnings import warn
+
+# from ietf.group.models import Group
```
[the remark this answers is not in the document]
holloway: Done

ietf/doc/templatetags/document_type_badge.py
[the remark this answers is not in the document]
holloway: Like this [8e3253f](https://github.com/ietf-tools/datatracker/pull/7760/commits/8e3253f57206b86b35f06aa8d8332593e4a981b6)?

Or is the idea that lines 29-32 have an `assertion`... i'm not sure what we're asserting is that point in the code.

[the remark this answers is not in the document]
jennifer-richards: That's fine. The assertion method would be to add `log.assertion("doc.type_id in ['rfc', 'draft']")` just before line 17, then reduce the `else` to only returning "".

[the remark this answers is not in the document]
rjsparks: (log.assertion asserts in dev mode, but only logs in production mode.)

jennifer-richards (approved):

PR #7765, opened 2024-07-29
feat: recognize HTTPS via proxy

jennifer-richards:
This will cause Django to treat HTTPS connections to CloudFlare as secure. This will cause it to generate `https://...` URLs when generating absolute URLs despite seeing non-TLS proxied requests.

Before deploying this, we should confirm that CloudFlare is behaving as required in the warning at https://docs.djangoproject.com/en/dev/ref/settings/#secure-proxy-ssl-header. We also need to be sure we never deploy this setting without a proxy that handles the `X-Forwarded-Proto` header as required.


rjsparks (approved):

jennifer-richards:
For the record: we've successfully confirmed Cloudflare's `X-Forwarded-Proto` handling, and it does indeed send it set to `https` for an incoming `https` request.

PR #7766, opened 2024-07-30
fix: Sort RFCs by date

### LINKED ISSUE #7740: Sorting RFC's by date on Datatracker profile pages is incorrect
### Describe the issue

Current behaviour: When visiting a user profile page, if you click to sort RFC's by date, it sorts it based on the Month Year string. This results in entries being sorted by Month and then year. eg: April 2024 can appear before June 2000.

Expected behaviour: Sort by year and then by month, thus sorting in chronological order.

Changing the displayed dates to a YYYY-MM-DD format for example may be a viable fix.

### Code of Conduct

holloway:
Fixes #7740

Allows sorting by `data-sort-number` attribute on the cell, rather than by the content of the cell which might be formatted.

### DIFF
--- a/ietf/static/js/list.js
+++ b/ietf/static/js/list.js
@@ -5,8 +5,10 @@ import {
 function text_sort(a, b, options) {
 
     function prep(e, options) {
-        return $($.parseHTML(e.values()[options.valueName]))
-            .text()
+        const el = $($.parseHTML(e.values()[options.valueName]));
+        const cell_el = e.elm.querySelector(`.${options.valueName}`)
+        const sort_by_number = cell_el?.getAttribute('data-sort-number')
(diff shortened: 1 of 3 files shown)


NGPixel (dismissed):

ietf/static/js/list.js
```diff
@@ -5,8 +5,10 @@ import {
 function text_sort(a, b, options) {
 
     function prep(e, options) {
-        return $($.parseHTML(e.values()[options.valueName]))
-            .text()
+        const el = $($.parseHTML(e.values()[options.valueName]));
```
jennifer-richards, line 8: `el` gato 🐱

jennifer-richards (dismissed):

jennifer-richards (approved):

PR #7772, opened 2024-07-31
feat: Agenda URL #hash scrolls to 'now' or specific day

### LINKED ISSUE #7245: Button to go straight to “now”
### Description

I probably pulled this page up on mobile dozens and dozens of times during 119. And then proceeded to scroll down to the helpfully highlighted section. Would it be possible to add a prominently placed button at the top of the mobile page to go to “now”, preferably even something that I can link to with a link on the phones Home Screen? If this doesn’t make sense or is unclear, please ask :-)

### Code of Conduct

holloway:
Fixes #7245

Allows scrolling to either `#now` or `#agenda-day-YYYY-MM-DD` on page load.

Adds a developer-only utility to modify server agenda dates to be the same as local browser time, so that the 'Now' button is available.


client/agenda/AgendaQuickAccess.vue
```diff
@@ -206,8 +206,6 @@ function scrollToDay (dayId, ev) {
 }
 
 function scrollToNow (ev) {
-  ev.preventDefault()
```
holloway, line 209: Removed so that when clicking 'Now' (`<a class="nav-link" href="#now"><i class="bi bi-arrow-right-short d-none d-xxl-inline me-2"></i><span>Now</span></a>` ... Screenshot from 2024-08-01 10-52-50 )

...it will update the URL with `#now` so that future reloads will have a hashstate that expresses 'Now'

client/agenda/store.js
```diff
@@ -154,6 +154,29 @@ export const useAgendaStore = defineStore('agenda', {
           this.timezone = agendaData.meeting.timezone
         }
 
+        // The following is for developers to modify `agendaData` dates to match
+        // your current system time so that features depending on an overlap of times
+        // such as the 'Now' button and red divider are available.
+        // This should be commented out when not in 
+        // (function(){
```
holloway, line 161: This is the developer-only utility to modify server agenda dates to be the same as local browser time, so that the 'Now' button is available.

I could make it a utility function and add tests etc, or make it active with a specific url param or something like that, but I'm not sure how far to go with it. What approach does this codebase usually take?

rjsparks: There's not a pattern to follow.

rjsparks: You _could_ follow the python settings.SERVER_MODE, or have a separate front-end server mode setting but I don't the the LOE required is the right investment right now. The danger of having a commented out block like this is that it will rot without CI noticing.

rjsparks:
This is looking good on desktop, but on screens the size of an iphone 11, its not acting as intended. With the dev code uncommented, looking in this view, the now link doesn't change what's in the url bar, and pasting http://localhost:8000/meeting/agenda/#now into a blank tab doesn't cause the viewport to move to the red line.

holloway:
@rjsparks the functionality should work now. I still need to fix the tests.

holloway:
@rjsparks fixed. ready for review

client/agenda/store.js
NGPixel: You can already modify the current time from the agenda settings panel for debugging:

image

NGPixel (changes requested):

client/agenda/store.js
holloway: This block has been removed in favour of the debugging panel

rjsparks (dismissed):

NGPixel (approved):

PR #7787, opened 2024-08-05
feat: Unify slide upload and proposal

### LINKED ISSUE #7178: Chair should be able to propose slides
### Description

Chairs may want to recuse themselves on a document that they are co-author on.
Being able to just post slides on my documents makes me feel like an elephant stepping on my co-chairs' feet.
So I should be able to go through the same slide submission that every other author has to go through, if I chose so.

### Code of Conduct

### LINKED ISSUE #7705: Try to combine the /slides and /propose-slides endpoints in meeting material management.
Right now, it is confusing what link a chair should send to their presenters when encouraging them to upload slides. The chair sees /slides, but the presenters have to use /propose-slides.

It might be better to use a single endpoint (/slides) and make the endpoint behave differently (being very transparent about what's happening) given the role of the person using it.

RjS

jimfenton:
This removes the propose_session_slides page, URL, and view function and merges them into upload_session_slides. It adds a checkbox, visible to chairs, allowing them to propose, rather than automatically approve, slides they are submitting.

Fixes #7705
Fixes #7178 

The tests still have separate test_upload_slides and test_propose_session_slides functions. Some of this may be redundant, but I thought it better to do possibly overtest. Perhaps the tests should be combined as well.


ietf/meeting/tests_views.py
```diff
@@ -6835,6 +6854,8 @@ def test_submit_and_approve_multiple_versions(self, mock_slides_manager_cls):
         self.assertIn('third version', contents)
 
 
+
+
```
rjsparks, line 6858:
```suggestion
```

rjsparks (dismissed):

ietf/meeting/tests_views.py
jimfenton: Not sure where those blank lines came from!

jennifer-richards (approved):

Issue #7287, opened 2024-04-03
User may be sent to incorrect doc after editing ballot position when using multiple tabs/windows


jennifer-richards:
### Describe the issue

If a user is reviewing / editing ballots for two or more documents in different browser tabs or windows they may be taken to the wrong document after saving a ballot position.

To reproduce:
1. Log in to a dev instance/sandbox as an AD and navigate to `/iesg/discusses/`
2. Duplicate the browser tab. Call the original tab A and the new tab B.
3. In tab A, click on the first document's name (leftmost column) then select the "IESG evaluation record" tab
4. Switch to tab B. click on the second document's name then select the "IESG evaluation record" tab
5. Switch back to tab A and click the "Edit position" button near the top. Click "Save"
6. Notice that you're now at the "IESG evaluation record" page for the document in tab B, not the one that was in tab A

The problem is that the `views_doc.document_ballot()` view puts its URL in the session as `ballot_edit_return_point` and the `views_ballot.edit_position()` view uses this value (when available) to set the return URL. When multiple tabs / windows are in use, these can get out of sync.

Need to straighten this out. Keep in mind that this mechanism is used to allow the ballot edit view to return to different places depending on how it was reached. It is accessed at least from the `document_ballot()` view and the IESG agenda's view. Searching for places that `ballot_edit_return_point` is set in the session should identify the places it might need to return to.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

jennifer-richards:
As a workaround: If you reload the page in the browser immediately before clicking the "Edit position" button each time then this bug should not occur. I.e., reload before clicking the button in step 5 of the instructions above.

Luis900927:
views_doc.document_ballot()Ariel rojas

PR #7788, opened 2024-08-05
fix: Ballot return to url via url params rather than session

holloway:
See #7287

This migrates the `ballot_edit_return_point` from session to a url param, with some additional safety checks to protect against phishing attacks.

Tasks
* [x] verify all incoming routes
* ~~[x] Playwright tests~~


ietf/doc/views_ballot.py
```diff
@@ -185,9 +187,11 @@ def edit_position(request, name, ballot_id):
...
     balloter = login = request.user.person
 
-    if 'ballot_edit_return_point' in request.session:
-        return_to_url = request.session['ballot_edit_return_point']
-    else:
+    return_to_url = request.GET.get("ballot_edit_return_point")
+
+    if return_to_url is None:
+        print("*** DEFAULT return_to_url *********** ")
+        raise ValueError("ballot_edit_return_point is required param")
```
holloway, line 194: This is a 'fail fast' approach to find incoming links that don't have the param. It will be removed later in dev.

holloway: Removed

ietf/doc/views_ballot.py
```diff
@@ -337,11 +344,24 @@ def send_ballot_comment(request, name, ballot_id):
...
+    return_to_url = request.GET.get('ballot_edit_return_point')
 
+    if return_to_url is not None:
+        # we need to ensure return_to_url isn't used for phishing attacks.
+        # return_to_url is used in HttpResponseRedirect() which could redirect to Datatracker or offsite.
+        # Eg http://datatracker.ietf.org/?ballot_edit_return_point=https://example.com/phishing-attack
+        # offsite links could be phishing attempts so let's reject them all, and require valid Datatracker
+        # routes
+        try:
+            urlresolve(return_to_url)
```
jennifer-richards, line 356: I wonder if we should instead (or maybe in addition) check that the URL is not absolute? I think that'd prevent the phishing issue.

I wonder if it'd be worth putting an even tighter check here and only redirect to the view that we want. E.g., someone could make a nuisance link that'd cause logout (at least, until we disable log out by GET). I worry that there might be more harmful games possible with this.

jennifer-richards (changes requested):
A trivial style nit, plus a worry about the impact of moving the redirect into the URL. You've guarded against phishing, but I wonder if we need a more aggressive guard there. One thought: we only have a few places to return to; would it be tractable to put parameter less general than the URL in the query parameter so that it can't be exploited?

Also, I think my inline comment was wrong and you are putting an absolute URL in the parameter - if that could be made into a relative one, that'd be somewhat safer.

Finally, I think the playwright tests are failing. Is that something to fix in the code or is that github struggling?

ietf/doc/views_ballot.py
holloway: > check that the URL is not absolute?

yeah I thought about checking whether it was a relative vs absolute URL through checking if it started with a `/` but then there's that protocol-relative URL syntax of `//example.com/phish` ([now considered an antipattern](https://stackoverflow.com/questions/28446314/why-use-protocol-relative-urls-at-all) so I'd have to use a real URL parser to be safe, and then if I were checking for valid routes by string that's a subset of relative URLs so it didn't feel necessary.

> only redirect to the view that we want

there are currently 5 views that assign `request.session['ballot_edit_return_point']` so I'm guessing there are 5 ways into the ballot editing and 5 ways we should redirect out. We could have an allow list of those 5 route patterns.

holloway: ^^ working on this now

holloway: Done. Thoughts?

ietf/doc/test_return_to_url.py
```diff
@@ -0,0 +1,37 @@
+# Copyright The IETF Trust 2013-2024, All Rights Reserved
+# -*- coding: utf-8 -*-
+
+import debug                            # pyflakes:ignore
+
+from ietf.utils.test_utils import TestCase
+from .return_to_url import parse_ballot_edit_return_point
+
+
+class ReturnToUrlTests(TestCase):
```
jennifer-richards, line 10: Now assuming `return_to_url.py` is empty, make this a class in `tests_ballot.py`?

ietf/doc/views_ballot.py
```diff
@@ -12,11 +12,11 @@
 from django.shortcuts import render, get_object_or_404, redirect
 from django.template.defaultfilters import striptags
 from django.template.loader import render_to_string
-from django.urls import reverse as urlreverse, resolve as urlresolve, Resolver404
+from django.urls import reverse as urlreverse
 from django.views.decorators.csrf import csrf_exempt
 from django.utils.html import escape
 from urllib.parse import urlencode as urllib_urlencode
-
+from .return_to_url import parse_ballot_edit_return_point
```
jennifer-richards, line 19: Minor, but organizationally we generally split imports into "basic packages", "django stuff", and "datatracker stuff" (i.e., `ietf.*`), roughly alphabetically in each section. This should then move down into the `ietf.*` section.

We're also slowly adopting the relative import syntax, so great to use it. If it's in a module that's not already using it, I've usually just followed whatever style is used there, but we should probably convert existing imports from `ietf.thisapp.whatever` to `.whatever` when doing so to keep it consistent. 

(If you accept my reorg suggestions then that'll be moot here, though)

jennifer-richards (changes requested):
I like the approach. I think it is likely to be useful outside this spot so I have some slight reorg suggestions inline.

jennifer-richards (changes requested):
I probably should have caught this on the last round, but the ValueErrors in the validate method will bubble up and, because they aren't handled, cause 500 errors. They should be caught in the views and lead to a 4xx response instead (probably a 400 unless there's a more specific appropriate one).

I'd be inclined to keep the response message simple and just say that the url is not allowed, rather than throwing url path names at the user. If you want to preserve the details you could log them, but I don't know if that's needed.

jennifer-richards (approved):

PR #7813, opened 2024-08-09
feat: total ids, pre-pubreq counts and pages left to ballot on on the AD dashboard

rjsparks:


evyncke:
Suggest to rename "pre pubreq" into something else because it looks like "pub req" (both are 2-works column title).

Is it "WG adopted document" ? or "WG I-D passed the WGLC" ? It is a little unclear.

rjsparks:
It's documents that have wg or ad state that indicate they are a draft that's on its way to stream processing. The details can be seen here (scroll the window) - look in particular at L496:
https://github.com/ietf-tools/datatracker/blob/c4b12d6cb13e76ac098e2018c7f3c162a25cca6f/ietf/doc/views_search.py#L488-L509

See also the discussion at https://github.com/ietf-tools/datatracker/issues/4577

I'm happy to take a suggestion for a different name, but this one is the one that was asked for.

jennifer-richards (approved):
Approving. We ought to move the inline javascript out of the template, but I'm not going to insist we do it here and now.

PR #7832, opened 2024-08-16
fix: remove unused notes field from SchedSessTimeAssignment model

rjsparks:
It's unclear what this column was ever intended to contain. All current objects have either None or "" for values. I've scrubbed the python for any use and haven't found any. Another careful pass, looking particularly for use in the js driving the agenda editing pages, would be prudent.


jennifer-richards (approved):
Looks fine to me - I don't see any use of the field in js, vue, or html files.

PR #7845, opened 2024-08-22
fix: Install weasyprint as a xml2rfc extra

kesara:
Since xml2rfc v3.23.0, WeasyPrint version is pinned.

This will also enforce the `pydyf` version[^1].

[^1]: https://github.com/ietf-tools/xml2rfc/blob/main/setup.cfg#L59-L60

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -71,5 +71,5 @@ tqdm>=4.64.0
 Unidecode>=1.3.4
 urllib3>=2
 weasyprint>=59
-xml2rfc>=3.12.4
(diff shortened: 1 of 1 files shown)


rjsparks (approved):

rjsparks:
I thought about dropping the explicit declaration of weasyprint from datatracker's requirements, relying on inheriting it from xml2rfc (to avoid some future update of the versions we require, perhaps coming in by automation, from breaking against the pin in xml2rfc), but making it obvious we have the dependency (for more than xml2rfc) is worth that maintenance burden.

PR #7873, opened 2024-08-29
feat: Use meetecho-player.ietf.org for session recording

kesara:


rjsparks (approved):
I still think we need to make the datatracker aware of whether a recording exists before we provide a link to it. Created #7888

PR #7877, opened 2024-08-30
feat: API changes needed for "real" draft imports

jennifer-richards:
This is the datatracker side of https://github.com/ietf-tools/rpc/pull/178


ietf/api/views_rpc.py
```diff
@@ -83,16 +97,17 @@ def rpc_draft(request, doc_id):
...
                 {
                     "id": p.pk,
                     "plain_name": p.person.plain_name(),
                 }
                 for p in d.documentauthor_set.all()
             ],
-        }
-    )
-
+        "shepherd": d.shepherd.formatted_ascii_email() if d.shepherd else "",
```
rjsparks, line 108: See my comment on the companion RPC PR - I'd rather not leak this information into the RPC app unless we really need to. The field will go through a big model change soon.

rjsparks (approved):
Approved with one reservation ☝️

ietf/api/views_rpc.py
ajeanmahoney: The Document Shepherd email address is included in the CC of the AUTH48 messages

rjsparks: We could fetch that directly from the datatracker when generating the message.

jennifer-richards: Leaving the shepherd email address in the API so it can be shown, but no longer storing it on the RPC side at all.

PR #7200, opened 2024-03-16
fix: Missing button text for PostScript

evyncke:
This fixes #6900 by avoiding any Postscript file in the list of built URL for a document,

PR #7889, opened 2024-09-03
fix: Missing button text for PostScript in RFC

### LINKED ISSUE #7879: Postscript document button is broken
### Describe the issue

image

ex) https://datatracker.ietf.org/doc/rfc2616/

### Code of Conduct

sh-cho:
Skip Postscript file in View.

In #7200, PS file is ignored but only in draft. So I appled also in rfc.

I guess removing 'ps' from `settings.RFC_FILE_TYPES`, `settings.IDSUBMIT_FILE_TYPES` should work, but don't know the side effect.. since I'm pretty new here 😅

Resolves #7879

### DIFF
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -1046,6 +1046,8 @@ def build_file_urls(doc: Union[Document, DocHistory]):
 
         file_urls = []
         for t in found_types:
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
lgtm

PR #7918, opened 2024-09-10
test: Use timezone aware datetime

kesara:
This might prevent errors like https://github.com/ietf-tools/datatracker/actions/runs/10786037653/job/29912248540#step:6:1902

### DIFF
--- a/ietf/meeting/tests_models.py
+++ b/ietf/meeting/tests_models.py
@@ -158,7 +158,7 @@ class SessionTests(TestCase):
 
     def test_session_recording_url(self):
         group_acronym = "foobar"
-        meeting_date = datetime.date.today()
+        meeting_date = datetime_today()
         meeting_number = 123
 
         # IETF meeting
--- a/ietf/meeting/tests_models.py
+++ b/ietf/meeting/tests_models.py
@@ -158,7 +158,7 @@ class SessionTests(TestCase):
 
     def test_session_recording_url(self):
         group_acronym = "foobar"
-        meeting_date = datetime_today()
+        meeting_date = date_today()
         meeting_number = 123
 
         # IETF meeting


ietf/meeting/tests_models.py
```diff
@@ -158,7 +158,7 @@ def test_alpha_str(self):
 
     def test_session_recording_url(self):
         group_acronym = "foobar"
-        meeting_date = datetime.date.today()
+        meeting_date = datetime_today()
```
jennifer-richards, line 161: This is a `datetime` - you probably want `date_today()`

jennifer-richards (changes requested):
Modulo date vs datetime, this should help.

There'll still be a very brief window around midnight UTC when this could fail. I wonder if we should write a test helper named something like `date_or_wait(dt: TimeDelta)` that will sleep if it's called within `dt` of a date change, then return the next day. I think this could protect us from random failures.

ietf/meeting/tests_models.py
kesara: Changed to `date_today()`.

kesara:
> Modulo date vs datetime, this should help.
> 
> There'll still be a very brief window around midnight UTC when this could fail. I wonder if we should write a test helper named something like `date_or_wait(dt: TimeDelta)` that will sleep if it's called within `dt` of a date change, then return the next day. I think this could protect us from random failures.

We can use something like [Freezgun](https://github.com/spulec/freezegun) to set a specific date too. The test I introduce follows the patterns of the existing test suite. The test for date is trivial & and go without it?

jennifer-richards:
> We can use something like [Freezgun](https://github.com/spulec/freezegun) to set a specific date too. The test I introduce follows the patterns of the existing test suite. The test for date is trivial & and go without it?

Seems worth investigating, but what you've got here is fine.

jennifer-richards (approved):

PR #7932, opened 2024-09-12
fix: rectify mixed types in gathering mailtrigger recipients

rjsparks:


jennifer-richards (approved):

rjsparks:
Merging this even though unrelated tests still fail as they already fail on main.

PR #7935, opened 2024-09-13
fix: start to reconcile internal inconsistencies wrt multiple from values

rjsparks:
Changes to parseaddr between 3.9.19 and 3.9.20 uncovered errors that have existed in our mail sending code since it was first written, and highlighted that the ietf.utils.mail module is inconsistent in its design for messages with multiple From header values.

This is a stopgap PR to preserve behavior to the extent possible as we upgrade python. More work will be needed to address the inconsistent ideas around multiple from header values.

We have not sent a message from the datatracker with multiple from header values in nearly 15 years, so for the moment, we're warning (in a way that intentionally raises test failures), but otherwise keeping the existing behavior if new attempts to use multiple values occur.

The older test that appeared to be testing multiple values was only passing as a coincidence of bugs in the code. It has been restricted to a single value until the module can be reworked.

### DIFF
--- a/ietf/utils/mail.py
+++ b/ietf/utils/mail.py
@@ -92,7 +92,17 @@ def send_smtp(msg, bcc=None):
     '''
     mark = time.time()
     add_headers(msg)
-    (fname, frm) = parseaddr(msg.get('From'))
+    # N.B. We have a disconnect with most of this code assuming a From header value will only
+    # have one address.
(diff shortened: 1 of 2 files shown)


jennifer-richards (approved):
lgtm. Could add a test that the assertions fire as expected, but I'm not convinced it's worth the trouble.

Issue #7455, opened 2024-05-22
Canceled interim meeting shows wrong date on group meetings page


cindymorgan:
### Describe the issue

The IAB had a meeting scheduled for 2024-05-22 that was canceled on 2024-05-21. On the IAB's meeting page in the Datatracker, it shows the canceled meeting's date as 2024-05-21 instead of 2024-05-22.

Also, would it be possible to have the "Cancelled" notification for a meeting also show up on the cancelled meeting's agenda page (e.g. https://datatracker.ietf.org/doc/agenda-interim-2024-iab-17-iab-01/)?

Thanks,
Cindy

<img width="1040" alt="Screen Shot 2024-05-22 at 12 30 22 PM" src="https://github.com/ietf-tools/datatracker/assets/22178783/be346fb8-d901-4ce3-be1c-53fda99eb590">

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
Still an issue:
![image](https://github.com/user-attachments/assets/2b7c10a6-b6cb-43ae-af3c-ae3dd799451d)

PR #7959, opened 2024-09-20
feat: warn about materials for cancelled sessions

rjsparks:
addresses the remainder of #7455

### DIFF
--- a/ietf/doc/views_doc.py
+++ b/ietf/doc/views_doc.py
@@ -872,9 +872,10 @@ def document_main(request, name, rev=None, document_html=False):
             )
 
         session_statusid = None
-        if doc.session_set.count() == 1:
-            if doc.session_set.get().schedulingevent_set.exists():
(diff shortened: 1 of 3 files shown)


ietf/doc/views_doc.py
```diff
@@ -870,6 +870,12 @@ def document_main(request, name, rev=None, document_html=False):
                     and doc.group.features.has_nonsession_materials
                     and doc.type_id in doc.group.features.material_types
             )
+
+        session_statusid = None
+        if doc.session_set.count() == 1:
+            if doc.session_set.get().schedulingevent_set.exists():
```
jennifer-richards, line 876: Based on the test failure, this needs to handle the case that `doc` is a `DocHistory`

jennifer-richards (changes requested):
See inline - for once, the test failure was real and not just factory-related noise

jennifer-richards (approved):

Issue #5560, opened 2023-05-01
On draft-expiration, documents with IESG state shouldn't change IESG state.


rjsparks:
Before we added an explicit ID-Exists draft-iesg state, a decision was made (with IESG input) to have the expiration code move the iesg-state for an expired document to the Dead IESG state. Note that this could only happen if the draft was in an IESG state that did not prevent expiration.

Now that we have the ID-Exists state, is it a better target? Which will cause the IESG the least surprise (if a document moves out of "AD is watching" for example)?

rjsparks:
@evyncke @wkumari

wkumari:
On Tue, May 02, 2023 at 5:01 AM, Eric Vyncke ***@***.***>
wrote:

> FYI, I personally use "I-D exists" state for WG documents that I want to
> easily see in "my documents" page (usually for WG/document specific
> reasons). I.e., there are no semantic associated for me to this state.
>
> I was unaware that when a document action holder is the IESG in some
> specific states (?) and when the I-D expires then it is moved to a "dead"
> state. I do not remember seeing it ;)
>
> All in all, I prefer to keep the current process, i.e., move it to a "IESG
> dead" state upon expiration.
>

Yes, I think I agree with this — it also helps keep it clearer where the
document is in the process…

W

rjsparks:
@6051 argues otherwise.

I think we have a real problem here to work out, and suggest that we want an entirely different mechansim for letting an AD put a draft that's not yet in IESG processing show up on their dashboard rather than giving it an IESG state. The very existance of an "AD is watching" state was (as my memory serves) entirely a hack to enable the original IESG tracker (pre-datatracker days) to hold information about a document that hadn't been pubreqed, and it got carried over and went essentially unused for awhile, then became used exactly and only to make a document show on the dashboard.

I'm leaning towards making it so that an AD can just mark any draft as watching, and it gets _no_ IESG state - all it does is make it show up on the dashboard, with an explanation of _why_ it shows up on the dashboard. Watching would automatically be turned off for a set of reasons (draft enters real IESG processing being one of them, expiry would not be - an AD would remove long expired drafts from their watch list when they were convinced they no longer needed watching).

@wkumari: With that in mind, Please say more about "keep it clearer where the document is in the process". How, in the existing system, would automatically returning a document to "idexists" make things less clear than an automatic return to "dead"? In the new system, if a watched draft never got IESG state until it was pubreqed, how would that make anything less clear?

[...]

PR #7960, opened 2024-09-20
feat: remove "AD is watching" state

### LINKED ISSUE #6316: Remove the "AD is Watching" draft-iesg state.
### Describe the issue

Putting a doc into "AD is watching" state makes the AD the action holder. That should IMO not happen, because watching does not imply any action being needed by the AD.

Example: https://datatracker.ietf.org/doc/draft-halpern-gendispatch-antitrust/

### Code of Conduct

### LINKED ISSUE #7583: Remove “AD is watching” state
### Description

“AD is watching” has outlived its usefulness, which Robert knows the history of better than I do. It only serves to confuse new ADs now until they learn to ignore it. Let’s get rid of it. 

### Code of Conduct

jennifer-richards:
Migrates any docs in the `watching` state to `idexists`, updating their histories and adding a comment to their event log. Removes all references to the `watching` state from the code. Removes the `create_in_state` choice field from the `edit_info()` view used to begin the IESG process and instead uses the `pub-req` state exclusively.

Fixes #6316
Fixes #7583

### DIFF
--- a/ietf/doc/migrations/0024_remove_ad_is_watching_states.py
+++ b/ietf/doc/migrations/0024_remove_ad_is_watching_states.py
@@ -69,9 +69,7 @@ def forward(apps, schema_editor):
         DocHistoryAuthor=apps.get_model("doc", "DocHistoryAuthor"),
     )
 
-    draft_state_type = StateType.objects.get(slug="draft")
     draft_iesg_state_type = StateType.objects.get(slug="draft-iesg")
-    dead_state = State.objects.get(type=draft_iesg_state_type, slug="dead")
     idexists_state = State.objects.get(type=draft_iesg_state_type, slug="idexists")
     watching_state = State.objects.get(type=draft_iesg_state_type, slug="watching")
     system_person = Person.objects.get(name="(System)")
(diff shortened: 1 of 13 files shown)


jennifer-richards:
Created as draft because there are still a bunch of references to the `"watching"` state in code and tests that need to be dealt with.

jennifer-richards:
In deciding what to do with docs in the "AD is watching" state (of which there were 60 as of a ~week-old dev DB snapshot), I'm running into #5560 - my first pass at the migration put expired drafts into `dead`, but it is not clear whether that's the right thing to do. From the discussion there, it seems that perhaps `idexists` would be better. Relevant, the first few `watching` drafts I looked at were similar to https://datatracker.ietf.org/doc/draft-murata-kohn-lilley-xml/history/, where the history incorrectly indicates they're in the `dead` state.

rjsparks:
yes, many transitions to dead should now be transitions to idexist. I can't think of any place we should have code choosing the dead state - an AD should have to choose that explicitly.

The code that does is leftover from before we made an idexists state for the draft-iesg state machine

ietf/doc/migrations/0024_remove_ad_is_watching_states.py
```diff
@@ -0,0 +1,121 @@
...
+    for doc in Document.objects.filter(states=watching_state):
+        assert doc.type_id == "draft"
+        doc.states.remove(watching_state)
+        doc.states.add(idexists_state)
+        e = StateDocEvent.objects.create(
+            type="changed_state",
+            by=system_person,
+            doc=doc,
+            rev=doc.rev,
+            desc=f"{draft_iesg_state_type.label} changed to <b>{idexists_state.name}</b> from {watching_state.name}",
```
rjsparks, line 87: cringing at the html tags, but keeping the set of things to mutate later regular was a good call

ietf/templates/doc/draft/change_state.html
```diff
@@ -12,9 +12,8 @@ <h1>
     </h1>
     {% if state.slug == "dead" %}
         <p class="alert alert-warning my-3">
-           This document is in IESG state "Dead". It is unusual to change
-           this to anything other than "AD is watching", and this should
-           never be used as a replacement for Begin IESG Processing.
+            This document is in IESG state "Dead". It is unusual to change this to
```
rjsparks, line 15: Add back the "never be used as a replacement for Begin IESG Processing". (New IESG members assume that moving things into pubreq will have all the side-effects that the form/view for Begin IESG Processing applies, but that's not possible.

We could also consider not allowing the document to make any other transition in this particular case.

rjsparks (changes requested):
Once guidance to new ads change requested.

rjsparks (dismissed):

rjsparks (approved):

PR #7725, opened 2024-07-21
Add copyrights

richsalz:
This is dependant on #7717.  It adds a script to add copyrights, and includes a commit to manually add them to some particularly troublesome files.

Think of this as WIP, because this PR does not include the changes for the rest of the files that don't have a copyright. You can run the ietf/add-copyright script by hand and review the changes.

Issue #7690, opened 2024-07-15
Auto-fix copyright years


richsalz:
### Description

I wrote a tool for OpenSSL to update copyright comments automatically, https://github.com/openssl/tools/blob/master/release-tools/do-copyright-year

We could add it to our release process.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
We could make it part of PR validation and (maybe?) provide a button to update changed files.

PR #7983, opened 2024-09-27
feat: support for building rfc authors

rjsparks:


richsalz (commented):
Should we look at the copyright stuff #7690 and #7725 ?

rjsparks:
We'll get there, but not yet.

PR #7995, opened 2024-10-01
fix: fix wg anchor tags (case-sensitive)

### LINKED ISSUE #7994: Links to Areas on https://datatracker.ietf.org/area/ do not work as expected
### Describe the issue

It looks like there is a case mismatch in the anchor tags.

For example:

<a href="/wg/#wit">(WIT)</a>

does not lead to the correct spot on https://datatracker.ietf.org/wg/ but:

<a href="/wg/#WIT">(WIT)</a>

would.

### Code of Conduct

microamp:
fixes #7994 

In https://github.com/ietf-tools/datatracker/blob/main/ietf/templates/group/active_wgs.html, it uses `upper`,
```jinja
    {% for area in areas %}
        <h2 class="mt-5" id="{{ area.acronym|upper }}">{{ area.name }} ({{ area.acronym|upper }})</h2>
```

so it's now doing the same on the area's side.

### DIFF
--- a/ietf/group/tests_info.py
+++ b/ietf/group/tests_info.py
@@ -88,11 +88,11 @@ class GroupPagesTests(TestCase):
             self.assertEqual(r.status_code, 200)
             self.assertContains(r, g.acronym)
             if t == "area":
-                self.assertContains(
-                    r,
-                    f'<a href="/wg/#{g.acronym.upper()}">({g.acronym.upper()})</a>',
-                    html=True,
-                )
+                q = PyQuery(r.content)
+                wg_url = urlreverse("ietf.group.views.active_groups", kwargs=dict(group_type="wg"))
+                href = f"{wg_url}#{g.acronym.upper()}"
+                self.assertEqual(q(f"h2#id-{g.acronym} a").attr("href"), href)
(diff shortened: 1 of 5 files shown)


microamp:
The IESG agenda page also has this issue. It's been addressed here, https://github.com/ietf-tools/datatracker/pull/7995/commits/4b2e45ab296d66a50d8c2c44cb21522aa6952c73.

jennifer-richards (dismissed):
lgtm

Only comment is that there's a slight tension in hard-coding the `/wg/` URL in the tests vs using `urlreverse()` to find it. I think it's probably ok either way because we're not likely to move the URL and it's a quick / obvious fix if we do.

rjsparks (changes requested):
Please address Jennifer's concern - use urlreverse instead of assuming /wg/).
I'd also suggest using jQuery to look for the element in the right place in the document rather than using assertContains - that may let you ignore the path part altogether and let you just test that the terminal part is capitialized.

jennifer-richards:
> jQuery

pyquery?

rjsparks:
yes, I meant pyquery

rjsparks (approved):

jennifer-richards (approved):

PR #8004, opened 2024-10-04
fix: 404 instead of 300 for ambiguous email_or_person

### LINKED ISSUE #8003: Unhandled exception / server error in lookup_community_list() when raising a MultiplePersonError when Person does not have a user
### Describe the issue

If more than one `Person` matches the `email_or_name` argument to `lookup_community_list()` and one or more of those `Person`s has no `user`, an unexpected exception occurs. This usually results in a 500 (server error) response.

The method is trying to populate a `MultiplePersonError` with a list of usernames that can be used to put together a list of disambiguated options for a 300 status code response. That's a helpful spirit, but given that we can't necessarily support that feature in general (and because it's unclear that browsers will actually do helpful things with the 300), I think it'd be better just to return a 404. If someone has trouble accessing their community list, they'll ask for help.

### Code of Conduct

jennifer-richards:
This changes both the CommunityList family of views and the Person photo view to return a 404 instead of a 300 if the `email_or_name` in the URL is ambiguous. In both cases, the old approach was attempting to be helpful but assumed that every `Person` matched had a `User` whose username could be used as a unique identifier. That is not true, so _something_ needs to be done.

While the 300 response with a list of usernames (or email addresses) is arguably helpful and friendly, I don't think a 404 is wrong - we're just declaring that we don't have a CommunityList or Photo that can be found at that URL. In either context, someone who is bothered by the missing resource is likely to complain about it and we can address the underlying issue.

Fixes #8003


rjsparks:
The line exposing username was incorrect anyhow.
I suspect we'll end up needing to revisit this, likely with person search.

rjsparks (approved):

PR #8007, opened 2024-10-04
fix: turn off automatic escaping in session request templates

### LINKED ISSUE #7993: HTML vs UTF-8 in messages from scheduling tool
### Describe the issue

```
Mon Sep 30 00:57:30 2024 Windl, Ulrich <u.windl@ukr.de> 
From: 	"Windl, Ulrich" <u.windl@ukr.de>
To: 	"postmaster@ietf.org" <postmaster@ietf.org>
Date: 	Mon, 30 Sep 2024 07:57:18 +0000
Subject: 	FW: [EXT] [Ntp] ntp - New Meeting Session Request for IETF 121

Hi Postmaster,

the message below using HTML entities (&#x27;) in the name used "Content-Type: text/plain; charset="utf-8"" in the header, so the HTML entity is displayed literally. I guess it's not intended that way.

Kind regards,
Ulrich


-----Original Message-----
From: IETF Meeting Session Request Tool <session-request@ietf.org>
Sent: Wednesday, September 18, 2024 9:01 PM
To: session-request@ietf.org
Cc: ek.ietf@gmail.com; kodonog@pobox.com; ntp-chairs@ietf.org; ntp@ietf.org
Subject: [EXT] [Ntp] ntp - New Meeting Session Request for IETF 121
Importance: High



A new meeting session request has just been submitted by Karen
O&#x27;Donoghue, a Chair of the NTP Working Group.


---------------------------------------------------------
Working Group Name: Network Time Protocols
Area Name: Internet Area
Session Requester: Karen O&#x27;Donoghue


Number of Sessions: 1
Length of Session(s): 1.5 Hours
Number of Attendees: 30
Conflicts to Avoid:
Chair conflict: jose




Participants who must be present:

Resources Requested:

Special Requests:
Earlier in the week would be better if possible. There is a plan to have a Hackathon project, and if we can get an meeting session earlier in the week, we will likely have more in person attendees.

```

### Code of Conduct

microamp:
fixes #7993

### DIFF
--- a/ietf/secr/sreq/tests.py
+++ b/ietf/secr/sreq/tests.py
@@ -15,7 +15,7 @@ from ietf.meeting.factories import MeetingFactory, SessionFactory
 from ietf.name.models import ConstraintName, TimerangeName
 from ietf.person.models import Person
 from ietf.secr.sreq.forms import SessionForm
-from ietf.utils.mail import outbox, empty_outbox, get_payload_text
+from ietf.utils.mail import outbox, empty_outbox, get_payload_text, send_mail
 from ietf.utils.timezone import date_today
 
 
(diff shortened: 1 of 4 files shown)


jennifer-richards (approved):
I'm a little concerned because this ends up outputting the user-provided `person.name` in an unescaped context. Since this is destined for plaintext email I think it's ok, but we should be careful.

rjsparks:
If recipients honor the text/plain mime-type these are sent with, I don't think we have any concern (and we do use autoescape=off for most other templates that generate text for email.

But I think we should, separately, turn up the level of sanitization of user input for names - they can be blobs, but the blobs don't need to contain html, for example. I've been toying with comparing the various names stored in the datatracker (and names we derive from them with `bleach.clean(name,[],strip=True)` hand haven't found a case where we wouldn't want to have rejected the name if those came up different.

rjsparks (approved):

PR #8040, opened 2024-10-10
fix: versioned href for slides URL

jennifer-richards:
Sends a versioned URL constructed by `DocumentInfo.get_href()` instead of `get_versionless_href()`. This should avoid cache-related issues when Meetecho updates their slides. It also fixes a bug (which, afaik, has not been encountered) where the latest slides would always be shown, even if a specific version was selected.

The comment about why `get_versionless_href()` was used appears to be stale - I think it was fixed in #7172, which was committed before any of this ever hit the main branch to begin with. I checked all the uses and do not see any situations where it a slides document (or related `SessionPresentation`) is changed/saved after a `SlidesManager()` method is called.

### DIFF
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -1840,7 +1840,7 @@ def agenda_extract_slide(item):
         "id": item.id,
         "title": item.title,
         "rev": item.rev,
-        "url": item.get_versionless_href(),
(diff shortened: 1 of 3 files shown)


rjsparks:
cc: @alexamirante (for awareness)

rjsparks (approved):

PR #8050, opened 2024-10-14
chore: Bypass Firefox for Playwright test due to flakiness

holloway:
These tests don't test anything Firefox-specific (it's conventional JavaScript mostly dependant on a 3rd-party library) so we've decided to bypass these tests in Firefox and just rely on Chrome.

### DIFF
--- a/playwright/tests/status/status.spec.js
+++ b/playwright/tests/status/status.spec.js
@@ -20,6 +20,10 @@ test.describe('site status', () => {
     by: 'Exile is a cool Amiga game'
   }
 
+  test.beforeEach(({ browserName }) => {
+    test.skip(browserName === 'firefox', 'bypassing flaky tests on Firefox')
+  })
+
   test('Renders server status as Notification', async ({ page }) => {
     await page.route('/status/latest.json', route => {
       route.fulfill({


NGPixel (approved):

holloway:
fyi @rjsparks

jennifer-richards (commented):
Is there a way to simply remove Firefox from the list being tested, or is skipping it the only way to do that?

I think part of our motivation for testing against firefox was to ensure we're not using corners that are browser-dependent. Not for this PR, but is there some alternative way to test that?

rjsparks:
I think we had also hoped that testing against Firefox and Chrome would also give us an early warning if either browser changed behavior in a way it shouldn't that we would need to adjust for (this has happened to us at several meetings).

I agree that right now the firefox playwright tests are not helping us because of the random inexplicable failures. I'm ok with muting that noise for the moment, but I think we should put pressure on making sure we're not surprised by activity at firefox.

NGPixel:
One reason to keep the Firefox tests is to make sure we don't start using a CSS or JS feature that's not yet supported in Firefox. Firefox tends to be behind times when it comes to implementing new standards.

PR #8053, opened 2024-10-16
feat: django-rest-framework

jennifer-richards:
Add django-rest-framework and friends, then reimplement the existing API. Generates nearly the same API schema with a few notable exceptions:
* Several URLs have moved around a bit (e.g., things that were under `/doc` are under `/draft` or `/rfc`). This is transparent when using the API generators. We should aim to end up with something logical, but it's easy to adjust.
* Endpoints that mapped `list[<id>]` to `dict(<id>, <some complex structure>)` now just return `list[<some complex structure>]` because that's easier to express. 
* Error return values are now included via drf-standardized-errors. (Currently only 400/404 to keep noise down, but we can add other types if they're interesting)

Re: the dict mapping, since refactoring those endpoints I believe I have a better idea how to wrangle DRF so I believe we could revert that API change.

The API schema can be generated by running `ietf/manage.py spectacular --file rpcapi-generated.yaml`. I've temporarily committed a copy of this for comparison (though diffs are pretty noisy)


corp-jones-damarcus (commented):
jennifer-richards:rpc-api-drf

corp-jones-damarcus (commented):
jennifer-richards:rpc-api-drf

corp-jones-damarcus (commented):
jennifer-richards:rpc-api-drf and damarcus Jones

Michealinc (commented):
Explict

Michealinc (commented):
Explict

PR #8062, opened 2024-10-21
feat: explicit names for meetecho recordings

### LINKED ISSUE #3454: Build API to capture URLs to Meetecho hosted video recordings for a Session
`type_defect`   |    by rjsparks@nostrum.com

___

Meetecho is bringing back the hosted video recordings that have a syncronized view of the entire session experience. We should add a link to them from the agenda - we can either derive the link to the Meetecho resource by parsing a token out of the YouTube URL that Meetecho currently posts or add an API to let Meetecho tell the datatracker about the URL explicitly. If we add an API, it would be easier for Meetecho to tell the datatracker when the recording stops being available.

___
_Issue migrated from trac:3454 at 2022-03-04 09:18:48 +0000_

rjsparks:
fixes #3454

### DIFF
--- a/ietf/api/tests.py
+++ b/ietf/api/tests.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2015-2020, All Rights Reserved
+# Copyright The IETF Trust 2015-2024, All Rights Reserved
 # -*- coding: utf-8 -*-
 import base64
 import datetime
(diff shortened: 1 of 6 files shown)


ietf/api/tests.py
```diff
@@ -222,6 +222,70 @@ def test_api_set_session_video_url(self):
...
+
+        badrole = RoleFactory(group__type_id="ietf", name_id="ad")
+        badapikey = PersonalApiKey.objects.create(endpoint=url, person=badrole.person)
+        badrole.person.user.last_login = timezone.now()
+        badrole.person.user.save()
+        r = self.client.post(url, {"apikey": badapikey.hash()})
+        self.assertContains(r, "Restricted to role: Recording Manager", status_code=403)
+
+        r = self.client.post(url, {"apikey": apikey.hash()})
+        self.assertContains(r, "Too long since last regular login", status_code=400)
```
jennifer-richards, line 246: If we don't already, we ought to cover features like this in a unit test of the `require_api_key()` decorator and trust that. (Not actually asking for a change here unless you're enthusiastic about it, just pointing it out.)

ietf/api/urls.py
```diff
@@ -39,6 +39,8 @@
     url(r'^iesg/position', views_ballot.api_set_position),
     # Let Meetecho set session video URLs
     url(r'^meeting/session/video/url$', meeting_views.api_set_session_video_url),
+    # Let Meetecho tell us the name of its recordings
+    url(r'meeting/session/recordingname$', meeting_views.api_set_meetecho_recording_name),
```
jennifer-richards, line 43: Need a `^` at the start.

Maybe make it `^meeting/session/recording-name` to better match `recording_name` in the model (and kebab-case that we've used in other URLs, like `doc/draft-aliases/` at line 27)

ietf/meeting/models.py
```diff
@@ -1332,17 +1333,23 @@ def onsite_tool_url(self):
...
 
     def session_recording_url(self):
         url_formatter = getattr(settings, "MEETECHO_SESSION_RECORDING_URL", "")
         url = None
-        if url_formatter and self.video_stream_url:
-            url = url_formatter.format(session_label=self._session_recording_url_label())
+        name = self.meetecho_recording_name
+        if name is None or name.strip() == "":
+            name = self._session_recording_url_label()
+        if url_formatter.strip()!="" and name is not None:
```
jennifer-richards, line 1351: style nit: spacing around `!=`

jennifer-richards (changes requested):
Couple small nits.

I wonder if it'd make sense to simply make this the `.../recording` URL endpoint instead of specializing it. The structure of the payload would lend itself to other metadata relatively transparently. OTOH maybe it's better to plan on retiring this endpoint when that happens rather than mutating its behavior.

rjsparks:
we may need to use (build on top of) the recording url endpoint for non-meetecho generated youtube videos (e.g. those for MOQ right now).

ietf/api/tests.py
rjsparks: I felt the lack-of-reuse pain, but didn't pursue it as I think we might want to rework _all_ of these to use DRF going forward maybe?

ietf/api/urls.py
rjsparks: will get the ^.
I can kebob case it, but I don't think we have consistency there.

jennifer-richards (approved):

PR #8084, opened 2024-10-25
feat: Session adding notes and recordings

### LINKED ISSUE #8033: Interface to add youtube links to meetings
### Description

Our interim meetings do not use Meetecho.

I manually file a ticket each week with support@ietf.org to upload the recording to YouTube using the IETF account. It would be nice to add a link to the Meeting's datatracker page, but there is no button there to do that.

If you're concerned about inappropriate videos, restricting it to videos on the IETF's account would be a reasonable feature.

### Code of Conduct

holloway:
fixes #8033

**Notes and recordings**
Screenshot from 2024-10-25 21-43-23

**Adding notes and recordings**
Screenshot from 2024-10-25 21-43-41

### DIFF
--- a/ietf/meeting/urls.py
+++ b/ietf/meeting/urls.py
@@ -16,6 +16,7 @@ class AgendaRedirectView(RedirectView):
 safe_for_all_meeting_types = [
     url(r'^session/(?P<acronym>[-a-z0-9]+)/?$',  views.session_details),
     url(r'^session/(?P<session_id>\d+)/drafts$',  views.add_session_drafts),
+    url(r'^session/(?P<session_id>\d+)/notes_and_recordings$', views.add_session_notes_and_recordings),
     url(r'^session/(?P<session_id>\d+)/attendance$', views.session_attendance),
     url(r'^session/(?P<session_id>\d+)/bluesheets$', views.upload_session_bluesheets),
     url(r'^session/(?P<session_id>\d+)/minutes$', views.upload_session_minutes),
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -20,7 +20,7 @@ from calendar import timegm
 from collections import OrderedDict, Counter, deque, defaultdict, namedtuple
 from functools import partialmethod
 import jsonschema
-from urllib.parse import parse_qs, unquote, urlencode, urlsplit, urlunsplit
+from urllib.parse import parse_qs, unquote, urlencode, urlsplit, urlunsplit, urlparse
 from tempfile import mkstemp
 from wsgiref.handlers import format_date_time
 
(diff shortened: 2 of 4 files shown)


ietf/meeting/urls.py
```diff
@@ -16,6 +16,7 @@ def get_redirect_url(self, *args, **kwargs):
 safe_for_all_meeting_types = [
     url(r'^session/(?P<acronym>[-a-z0-9]+)/?$',  views.session_details),
     url(r'^session/(?P<session_id>\d+)/drafts$',  views.add_session_drafts),
+    url(r'^session/(?P<session_id>\d+)/notes_and_recordings$', views.add_session_notes_and_recordings),
```
rjsparks, line 19: lets make this endpoint just "recordings"

ietf/meeting/views.py
```diff
@@ -2538,6 +2538,55 @@ def add_session_drafts(request, session_id, num):
                     'form': form,
                   })
 
+class SessionNotesAndRecordingsForm(forms.Form):
```
rjsparks, line 2541: Again, just recordings

ietf/meeting/views.py
```diff
@@ -2538,6 +2538,55 @@ def add_session_drafts(request, session_id, num):
...
+
+    session_number = None
+    sessions = get_sessions(session.meeting.number,session.group.acronym)
+    if len(sessions) > 1:
+       session_number = 1 + sessions.index(session)
+
+    if request.method == 'POST':
+        form = SessionNotesAndRecordingsForm(request.POST,already_linked=already_linked)
+        if form.is_valid():
+            title = form.cleaned_data['title']
```
rjsparks, line 2577: Currently, for the meetecho uploads, the title is set by automata to look something like, e.g.,`Video recording for tools on 2024-10-15 at 18:00:00`.  We should probably suggest a title of that form rather than starting with an empty control. Giving chairs the ability to pick a title here will be a new thing, but I think it's ok.

ietf/meeting/views.py
```diff
@@ -2538,6 +2538,55 @@ def add_session_drafts(request, session_id, num):
...
+        super(self.__class__, self).__init__(*args, **kwargs)
+
+    def clean(self):
+        url = self.cleaned_data['url']
+        parsed_url = urlparse(url)
+        if parsed_url.hostname not in ['www.youtube.com', 'youtube.com', 'youtu.be', 'm.youtube.com', 'youtube-nocookie.com', 'www.youtube-nocookie.com']:
+            raise forms.ValidationError("Must be a YouTube URL")
+        problems = set(url).intersection(set(self.already_linked)) 
+        if problems:
+           raise forms.ValidationError("Already linked: %s" % ', '.join([d.name for d in problems]))
```
rjsparks, line 2556: This has a different semantic than the meetecho upload API, which can _update_ a link in case there was an error. We should think through what a chair would do if they entered the wrong link and wanted to fix it.

rjsparks (changes requested):
This is really good, and close, but it is introducing some new behavior with edges we should talk about more.

If it turns out that any changes based on that discussion aren't quick to resolve, lets consider splitting showing the existing recordings on the materials page into a separate PR as it would have immediate utility even before adding the ability to upload more recordings.

And this is all about recordings - please scrub mentions of notes away in the new code (existing code should be linking to notes in that section - we'll address that it isn't separately).

jennifer-richards (commented):
I have a couple technical comments/suggestions inline.

The handling of `already_linked` feels a little strange to me. It makes it a little tricky to describe the new form class. It's not really logically a `SessionRecordingForm` because it doesn't actually know anything about the `Session` class, but it's not quite a more general `RecordingForm` because it's expecting this extra piece of data that is pulled out of the Session.

I guess I'd slightly prefer that either the form take a `session` parameter and then be responsible for knowing how to pull the `already_linked` data out of the session, or not take an extra parameter and do the `already_linked` checking in the view.

All that said, I'm not sure that a change is necessary, just something to consider.

ietf/templates/meeting/add_session_notes_and_recordings.html
```diff
@@ -0,0 +1,70 @@
...
+        </small>
+    </h1>
+    {% comment %} TODO: put the session name here or calculate the number at the meeting {% endcomment %}
+    {% if session.is_material_submission_cutoff %}
+        <div class="alert alert-warning my-3">
+            The deadline for submission corrections has passed. This may affect published proceedings.
+        </div>
+    {% endif %}
+    <div class="alert alert-info my-3">
+        This form will link additional Notes and Recordings to this session with a revision of "Current at time of presentation". For more fine grained control of versions, or to remove Notes and Recordings from a session, adjust the sessions associated with an Internet-Draft from the Internet-Draft's main page.
```
rjsparks, line 23: Sorry - missed commenting on this earlier: it's copied from the workflow that lets chairs associate drafts with sessions, and the lifecycle of such things is not the same. With drafts, chairs sometimes want to associate an older version with a session. Here, older recordings aren't really something to point to. It's a good indication that recording as a type of document may have been a mis-model. But in any case, please just remove this paragraph.

holloway:
> lets consider splitting showing the existing recordings on the materials page into a separate PR as it would have immediate utility even before adding the ability to upload more recordings.

https://github.com/ietf-tools/datatracker/pull/8102

holloway:
WIP for adding session recordings https://github.com/holloway/datatracker/tree/add-recordings

holloway:
See #8218

PR #8109, opened 2024-10-29
feat: limit line width for PlaintextDraft

jennifer-richards:
Long lines cause massive slowdowns parsing text drafts with the `PlaintextDraft` class - up to 1000x compared to shorter lines. This limits the width that is examined when parsing a draft to avoid this problem.

Removes an unused method that happens to use `PlaintextDraft` as well.


jennifer-richards:
Cutting off the lines impacted the `PlaintextDraft` parsing results for many older drafts, so abandoning this for now.

PR #8126, opened 2024-10-30
feat: api to gather draft authors

rjsparks:


jennifer-richards (approved):
Looks good. Does raise issues with confusion about revision.

rjsparks:
re: revision : It's intentional to get the DocumentAuthors linked to the current Document object. If we find some future use of this where we care about history, we can refactor.

PR #8133, opened 2024-11-03
feat: add wiki button to agenda list for hackathon sessions

NGPixel:
- Adds a wiki button (blue book icon) in lieu of the notepad button for hackathon sessions (notepad being irrelevant and never used for them AFAIK). See below for example.
- Replaces the special arrow symbol on the "Switch to text-only version" button while the agenda is loading with a standard `&raquo;` symbol. Should fix the arrow sometimes changing to a weird icon because of incorrect font / charset / font not loaded.

image


jennifer-richards (dismissed):

rjsparks:
I'm not sure on this one.
I think we should split the arrow font out into its own PR.
I'd like to talk about generalizing this to other sessions that have different things than the default notes location.

jennifer-richards (approved):

PR #8135, opened 2024-11-03
feat: add group leadership list

rpcross:


ietf/group/views.py
```diff
@@ -437,6 +438,39 @@ def prepare_group_documents(request, group, clist):
...
     return docs, meta, docs_related, meta_related
 
+def get_leadership(group_type):
+    people = Person.objects.filter(
+        role__name__slug='chair',
+        role__group__type=group_type,
+        role__group__state__slug__in=('active', 'bof', 'proposed')).distinct()
+    leaders = []
+    for person in people:
+        parts = person.name_parts()
```
jennifer-richards, line 448: Is this needed vs something like `person.plain_name()`? It'd be nice not to add to the variety of ways we handle names in various places. I guess this is to produce a "last name-first" list so maybe there's not much choice if that's needed for the Secretariat's purposes.

ietf/templates/group/group_leadership.html
```diff
@@ -0,0 +1,36 @@
+{% extends "base.html" %}
+{# Copyright The IETF Trust 2015, All Rights Reserved #}
```
jennifer-richards, line 2: Copyright 2024

ietf/templates/group/group_leadership.html
```diff
@@ -0,0 +1,36 @@
...
+            {% for leader in leaders %}
+                <tr>
+                    <td>{{ leader.name }}</td>
+                    <td>{{ leader.groups }}</td>
+                </tr>
+            {% endfor %}
+        </tbody>
+    </table>
+{% endblock %}
+{% block js %}
```
jennifer-richards, line 35: Can you omit the empty block?

jennifer-richards (changes requested):
The tests should probably check that `rg` groups are handled correctly. As it stands, I think `rg` groups could be ignored and the tests would still pass. (`RoleFactory(group__type_id="rg", name_id="chair")` might be a useful tool for that.

It'd also be nice to test that active/bof/proposed groups are included but, e.g., conclude groups are ignored. Here, `RoleFactory(..., group__state_id="conclude", ...)` might help.

FYI, the existing group you're seeing ("Sops") is set up in `ietf/utils/test_data.py`.

Suggest running the Black styler over the new code. (Or at least using double-quotes instead of single)

ietf/group/views.py
rpcross: Yes. The list is sorted by last name to match the way the meeting badges are sorted.

ietf/templates/group/group_leadership.html
rpcross: Done

ietf/templates/group/group_leadership.html
rpcross: Done

ietf/group/views.py
```diff
@@ -437,6 +438,48 @@ def prepare_group_documents(request, group, clist):
 
     return docs, meta, docs_related, meta_related
 
+
+def get_leadership(group_type):
+    people = Person.objects.filter(
+        role__name__slug="chair",
```
rjsparks, line 444: Not asking for a change now, but do think about the future when we might want to add IAB programs to these pages - we'll need to look for other role names at that point.

jennifer-richards (approved):
Thanks for the changes!

PR #8169, opened 2024-11-06
feat: Add idnits3 for post submission nits check button

### LINKED ISSUE #8107: Update nits check URL
### Description

The drafts view has a link for checking nits, which currently calls the author tools by using the `idnits` v2 endpoint and supplying the `.txt` draft as the source.

We should consider updating this to use the `idnits` v3 endpoint, providing `.xml` draft as the source. 

However, for drafts that lack an XML source, we may need to continue using the v2 endpoint temporarily, as `idnits` v3 isn’t ready to check `.txt` drafts yet.

### Code of Conduct

kesara:
This change adds a new Nits 3 button in addition to the existing Nits button.

Screenshot 2024-11-19 at 13 56 27

Fixes #8107

### DIFF
--- a/ietf/doc/tests.py
+++ b/ietf/doc/tests.py
@@ -2997,7 +2997,7 @@ class RawIdTests(TestCase):
         self.should_404(dict(name=draft.name, rev='00',ext='html'))
 
     @override_settings(IDNITS3_BASE_URL="https://example.org/idnits3")
-    def test_nits_url(self):
+    def test_nits3_url(self):
         draft = WgDraftFactory(name="draft-ietf-moon-test",rev="01", create_revisions=range(0,2))
 
         dir = settings.INTERNET_ALL_DRAFTS_ARCHIVE_DIR
@@ -3006,7 +3006,7 @@ class RawIdTests(TestCase):
         self.assertContains(r, settings.IDNITS3_BASE_URL)
 
     @override_settings(IDNITS_BASE_URL="https://example.org/nits")
-    def test_nits_url_txt_only(self):
+    def test_nits_url(self):
         draft = WgDraftFactory(name="draft-ietf-moon-test",rev="01", create_revisions=range(0,2))
 
         dir = settings.INTERNET_ALL_DRAFTS_ARCHIVE_DIR
--- a/ietf/doc/tests.py
+++ b/ietf/doc/tests.py
@@ -2996,6 +2996,24 @@ class RawIdTests(TestCase):
         self.should_succeed(dict(name=draft.name, rev='00',ext='txt'))
         self.should_404(dict(name=draft.name, rev='00',ext='html'))
 
+    @override_settings(IDNITS3_BASE_URL="https://example.org/idnits3")
+    def test_nits_url(self):
+        draft = WgDraftFactory(name="draft-ietf-moon-test",rev="01", create_revisions=range(0,2))
+
+        dir = settings.INTERNET_ALL_DRAFTS_ARCHIVE_DIR
+        (Path(dir) / f"{draft.name}-01.xml").touch()
+        r = self.client.get(urlreverse("ietf.doc.views_doc.document_main", kwargs=dict(name=draft.name)))
+        self.assertContains(r, settings.IDNITS3_BASE_URL)
+
+    @override_settings(IDNITS_BASE_URL="https://example.org/nits")
+    def test_nits_url_txt_only(self):
+        draft = WgDraftFactory(name="draft-ietf-moon-test",rev="01", create_revisions=range(0,2))
+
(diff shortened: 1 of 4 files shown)


ietf/doc/tests.py
```diff
@@ -2996,6 +2996,24 @@ def test_raw_id(self):
         self.should_succeed(dict(name=draft.name, rev='00',ext='txt'))
         self.should_404(dict(name=draft.name, rev='00',ext='html'))
 
+    @override_settings(IDNITS3_BASE_URL="https://example.org/idnits3")
+    def test_nits_url(self):
+        draft = WgDraftFactory(name="draft-ietf-moon-test",rev="01", create_revisions=range(0,2))
```
jennifer-richards, line 3001: A comment as to why you're setting up this particular arrangement of revisions would be helpful

jennifer-richards (approved):

jennifer-richards (changes requested):
Oops, meant to request changes for the comment (code itself looks fine)

rjsparks:
If someone really wanted the v2 nits, can they get back to them quickly?

rjsparks:
Nudge:

kesara:
> If someone really wanted the v2 nits, can they get back to them quickly?

No there are no quick links for that. Author tools just shows the output from idnits.

rjsparks:
so maybe instead of replacing the idnits2 button, we provide both idnits2 and idnits3 buttons until idnits3 becomes the thing in the submit path?

cabo:
Where can I find idnits3?
https://github.com/ietf-tools/idnits does not appear to be in a useful state.

kesara:
> Where can I find idnits3? https://github.com/ietf-tools/idnits does not appear to be in a useful state.

@cabo, You can install with `npm` [^1] or use [idnits3 on Auhtor Tools](https://author-tools.ietf.org/idnits3).

[^1]: https://github.com/ietf-tools/idnits?tab=readme-ov-file#installation

kesara:
> so maybe instead of replacing the idnits2 button, we provide both idnits2 and idnits3 buttons until idnits3 becomes the thing in the submit path?

One of the motivations of this change is to stop unnecessary idnits2 checks because this makes a call to datatracker which seems slow and could affect datatracker performance.

cabo:
> @cabo, You can install with `npm` [1](#user-content-fn-1-0f99bd6715f56c1bbcc1a7817eaea9a9) or use [idnits3 on Auhtor Tools](https://author-tools.ietf.org/idnits3).

OK, I have that, so "3.0.0-alpha.36" appears to be the most recent revision.
That is not very useful.

https://author-tools.ietf.org/api/idnits3?url=https://www.ietf.org/archive/id/draft-ietf-cbor-cddl-more-control-07.xml
https://mailarchive.ietf.org/arch/msg/last-call/IM2SIo26RzBq0tO0lG5JETyGGNo

kesara:
> > @cabo, You can install with `npm` [1](#user-content-fn-1-0f99bd6715f56c1bbcc1a7817eaea9a9) or use [idnits3 on Auhtor Tools](https://author-tools.ietf.org/idnits3).
> 
> OK, I have that, so "3.0.0-alpha.36" appears to be the most recent revision. That is not very useful.
> 
> https://author-tools.ietf.org/api/idnits3?url=https://www.ietf.org/archive/id/draft-ietf-cbor-cddl-more-control-07.xml https://mailarchive.ietf.org/arch/msg/last-call/IM2SIo26RzBq0tO0lG5JETyGGNo

@cabo, idnits v3 is still in alpha stage of development. But you can report any issues to idnits repository[^1] so we can correct them early.

[^1]: https://github.com/ietf-tools/idnits/issues/new

rjsparks:
We need to signal that this is a new, not yet complete, and under development tool. Maybe we can use a tooltip, but there are accessibility concerns with that. Maybe we should put it on its own row with explanatory text? Other ideas?

rjsparks:
Maybe we can land this one during the 122 codesprint?

kesara:
@rjsparks How about 
Screenshot 2025-03-15 at 11 31 20

Michealinc (commented):
Explict

evyncke:
Humm indeed #9167 is similar (I will then close it if I have the right to do so).

Else, any reason why the PR is not merged ? (unclear from the above comments)

kesara:
@rjsparks, With #9175 DT already displays id-nits3 for the IESG on the doc page.
Is there any value in working on this?

rjsparks:
The original proposal of this PR was to make the link available to everybody - we're probably there, but should label the idnits3 button beta until it is out of beta.

kesara:
Closing this as solution for #8107 is removing role checks in #9175.

Issue #7167, opened 2024-03-11
UTF8 only in Authors list


cindymorgan:
### Describe the issue

Subject: 	UTF8 only in Authors list
From: 	"Michael Richardson" <mcr@sandelman.ca>
To: 	support@ietf.org
Date: 	Sun, 10 Mar 2024 13:58:18 -0400

In the announcement for:

> Internet-Draft draft-ietf-lamps-rfc5019bis-05.txt is now available. It is a

it says:

Authors: 伊藤 忠彦
Clint Wilson
Corey Bonnell
Sean Turner

and of course, I'm too ignorant to read the Kanji.
Looking at the datatracker copies, I see it rendered as:

伊藤 忠彦 (T. Ito)

which I'm much better able to understand.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

richsalz:
@Spectre17, are you going to submit that `dublin-1` branch, or should I pull out the diff and make a single PR fro it?

rjsparks:
In the comments at https://github.com/ietf-tools/datatracker/pull/8178, that first attempt ran into issues and this really looks like it should be done with a bigger effort around how we handle names througout the datatracker. But if you think you see an easy way past the concerns Jennifer was raising, please poke at it.

richsalz:
After chatting at the codesprint, I am going to check with some UTF8 experts to see if checking for any multibyte codepoints is enough.

cabo:
> After chatting at the codesprint, I am going to check with some UTF8 experts to see if checking for any multibyte codepoints is enough.

I don't know what "checking for any multibyte codepoints" means but the point of (incorrectly named) asciiXXX is to give alt. text in Roman (Latin) characters when the actual XXX isn't Latin.  The assumption being that all people using the site can read Latin text but not necessarily other writing systems.
"multibyte codepoints" is not related to this requirement in any way.

richsalz:
Look at the commit https://github.com/Spectre17/datatracker/commit/d54560483daf25c2d5d0d90f4667761436fbabf5 which shows it's using the author full name and maybe also the ascii-full name.

That PR was closed.  I believe that if will convert the XML name to array of bytes and if the UTF8 bytes have any value > 0x80, then append the ascii-full-name.

PR #8178, opened 2024-11-07
feat: add support for display of differing fullname and asciiFullname

Spectre17:
This is used, for example, when the fullname uses extended UTF-8 characters and it's helpful to also include the asciiFullname for those who cannot read the UTF-8 name. The form is:

fullname (asciiFullname)

This came up in issue #7167, which noted that the announcement email generated when a new Internet-Draft is uploaded, it only showed the fullname despite the draft itself containing both name forms.

Change the render_author_name test set for the revised name format.

### DIFF
--- a/ietf/utils/xmldraft.py
+++ b/ietf/utils/xmldraft.py
@@ -193,6 +193,9 @@ class XMLDraft(Draft):
         # Use fullname attribute, if present
         fullname = author_elt.attrib.get("fullname", "").strip()
         if fullname:
+            asciifullname = author_elt.attrib.get("asciiFullname", "").strip()
+            if asciifullname and asciifullname != fullname:
+                fullname = fullname + ' (' + asciifullname + ')'
             return fullname
         surname = author_elt.attrib.get("surname", "").strip()
         initials = author_elt.attrib.get("initials", "").strip()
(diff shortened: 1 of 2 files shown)


ietf/utils/xmldraft.py
```diff
@@ -193,6 +193,9 @@ def render_author_name(author_elt):
         # Use fullname attribute, if present
         fullname = author_elt.attrib.get("fullname", "").strip()
         if fullname:
+            asciifullname = author_elt.attrib.get("asciiFullname", "").strip()
+            if asciifullname and asciifullname != fullname:
```
jennifer-richards, line 197: I'm not sure that literal equality is the right test. In xml2rfc, a [method is used to check whether a name is in Latin script](https://github.com/ietf-tools/xml2rfc/blob/bf52f861886a19aacc95938d5ba8ea5b62646785/xml2rfc/uniscripts/__init__.py#L28-L66) and the ascii name is appended if not. That might better reflect when the asciiName is needed.

jennifer-richards (commented):
Not approving or requesting changes, wondering whether we need to handle the fullname vs asciiFullname check with more subtlety.

ietf/utils/xmldraft.py
jennifer-richards: We should probably look at a bunch of drafts and see how often we're going to add extra text to a name already legible to someone who can read the rest of the draft.

rjsparks:
I'm closing this as we've learned through discussion that we need a larger project around how we use names throughout our applications.

PR #8205, opened 2024-11-12
fix: clarify "Replace" in I-D submission form

### LINKED ISSUE #8059: Submission: What does "replace" mean?
### Describe the issue

A not-so-frequent submitter just submitted a -07 for a draft.

He was confused by the field

*Any Internet-Drafts that this document replaces (approval required for replacing an Internet-Draft you are not the author of)*

Of course, his new draft replaces -06, but there was no way to enter that.

The legend needs to explain that this is about replacing I-D *series*, not the single I-Ds that make up a series (which are automatically "replaced" by the next one in the series).

### Code of Conduct

ghwood:
Update help text In the "Replacement Information" section of the "Status" tab to fix #8059


ietf/submit/forms.py
```diff
@@ -502,7 +502,7 @@ def clean_name(self):
         return name
 
 class ReplacesForm(forms.Form):
-    replaces = SearchableDocumentsField(required=False, help_text="Any Internet-Drafts that this document replaces (approval required for replacing an Internet-Draft you are not the author of)")
+    replaces = SearchableDocumentsField(required=False, help_text="An I-D submission will be automatically attached to the previous version of the I-D with the same name, or assigned "-00" version if no previous version exists. If this submission is intended to replace an I-D with a different name (e.g. when an individual I-D is adopted as a Working Group document), please enter that name here. (Approval required to replace an I-D for which you are not an author.)")
```
jennifer-richards, line 505: There's a syntax error here because of the `"-00"` inside the quoted string. You can change the outer quotes to `'` to fix this.

jennifer-richards (changes requested):
See inline

rjsparks:
Thanks for pushing on this. Some thoughts below:

> An I-D submission will be automatically attached to the previous version of the I-D with the same name, or assigned "-00" version if no previous version exists. If this submission is intended to replace an I-D with a different name (e.g. when an individual I-D is adopted as a Working Group document), please enter that name here. (Approval required to replace an I-D for which you are not an author.)

The first sentence doesn't work for me - it sees to imply "-00" is a thing that happens at this point, but -00 was chosen by the submitter long before the submitter gets to this form.

I also don't think "automatically attached to a previous" is what happens, or is the right model of what's happening for other people to build. Instead, we need to reinforce that there are versions of an Internet-Draft, and that an Internet-Draft might get replaced with some other Internet-Draft(s).

Maybe imperative statements rather than explanation would work better? My alternate suggestion:

-----

Do not enter anything here if you are only submitting the next version of your Internet-Draft. Only enter items here if this submission is intended to replace an I-D with a different name. A typical use of this field is to note what individual I-Ds are replaced by a new -00 Working group I-D. Note that additional approval will be required to replace an I-D for which you are not an author.

rjsparks:
mph - just spotted that you're merging back into your fork's main. That will be painful for you in the long run. I'll followup with you on it offline.

For anyone else that may stumble across this - it's better to leave main in your fork alone - sync it with ietf-tools/main periodically. Build your PRs in another branch.

PR #8206, opened 2024-11-13
feat: POST for document search requests

jennifer-richards:
This refactors the `ietf.doc.search()` view and some adjacent views to require a POST, including a CSRF key, before executing a document search. Search parameters are expected in the request body instead of the URL query string.

If a GET request with a query string is received the search is not executed. The user is sent back to the search page with the form pre-filled using the values from their request and an error box at the top of the page indicates that the GET interface is deprecated. (Maybe there's a better term - it's actually been retired - so suggestions on phrasing are welcome.)

I've refactored the cache key calculation and moved it into the `SearchForm` since it depends on details internal to the form. This helps with a circular import problem.

The `SearchForm` and results cache is reused in the stats module. I've so far refactored this to use the updated form but not to switch it over to POST. I haven't seen anything actually hit that endpoint in over 30 days (though I haven't exhaustively searched logs), so maybe it's better to drop it. That'd allow removal of an additional helper function that really doesn't need to exist.

A few things left before taking this out of draft:
- [x] check / update tests
- [x] rephrase the `messages.error()` notice to reflect that the GET interface is retired, not just deprecated
- [x] confirm that the expensive forms of `/doc/<some-value>` are all covered by this patch
    * The `search_for_name` view is probably in need of a refactor, but it's really a separate task so I'm not going to change it here.
    * When no simple match is found, a redirect to the GET-based search is returned. I think that's ok - we want to wean people away from using `/doc/whatever` as an alias for searching. 

It'd also be nice to update the `search_form.html` template to display form validation errors. I'm on the fence as to whether that's a separate feature or belongs here. Thinking it is probably separate and that the better fix should also handle opening the "Additional search criteria" accordion when non-default options are selected.


ietf/doc/views_stats.py
```diff
@@ -124,14 +124,11 @@ def chart_newrevisiondocevent(request):
 
 #@cache_page(60*15)
 def chart_data_newrevisiondocevent(request):
```
rjsparks, line 126: Pretty sure this doesn't work as intended, and it's not being used, so lets remove it.

rjsparks: removal done in #8210

rjsparks (dismissed):

jennifer-richards:
Marking as draft until #8210 is merged. I'll pull that into this branch and do cleanup here.

rjsparks (approved):

PR #8218, opened 2024-11-17
feat: Add session recordings

holloway:
Replaces #8084

* Allows a secretariat to add and delete session recordings (youtube urls)

Screenshot from 2024-11-18 11-11-17

Screenshot from 2024-11-18 11-11-37


ietf/meeting/views.py
```diff
@@ -2568,6 +2569,66 @@ def add_session_drafts(request, session_id, num):
...
+
+    session_number = None
+    sessions = get_sessions(session.meeting.number,session.group.acronym)
+    if len(sessions) > 1:
+       session_number = 1 + sessions.index(session)
+
+    if request.method == 'POST':
+        delete = request.POST['delete']
+        if delete:
+            delete_recording(pk=delete, session=session)
```
holloway, line 2603: >  We should think through what a chair would do if they entered the wrong link and wanted to fix it. @rjsparks 

I've added a delete button. Is that ok?


Screenshot from 2024-11-18 10-56-26

holloway: Regarding the 'Revision' column is that valid for recordings? How would someone update the revision of a recording anyway?

ietf/meeting/views.py
```diff
@@ -2568,6 +2569,66 @@ def add_session_drafts(request, session_id, num):
...
+        else:
+            form = SessionNotesAndRecordingsForm(request.POST,already_linked=already_linked)
+            if form.is_valid():
+                title = form.cleaned_data['title']
+                url = form.cleaned_data['url']
+                create_recording(session, url, title=title, user=request.user.person)
+                return redirect('ietf.meeting.views.session_details', num=session.meeting.number, acronym=session.group.acronym)
+    
+    today = datetime.datetime.now()
+    initial = {'title': f"Video recording for {session.group.acronym} on {today.strftime('%b-%d-%Y at %H:%M:%S')}"}
```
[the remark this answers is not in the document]
jennifer-richards: I don't know what the existing TZ is, but I think UTC is the right choice. I'd probably add "... UTC" to the default name, though it'd mean a break from existing filename convention.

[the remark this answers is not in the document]
rjsparks: Why today?

The recording is for a working group session and the times should be the times for that session, not the time of the upload of the recording.

We should consider capturing the moment of "upload"  (really, setting a URL) as a docevent, but I don't think we're currently doing that when meetecho reports uploads to youtube, so maybe this is a separate issue for improving both.

But yes, we should present in UTC by default - some views offer browser or meeting local time as well - this upload doesn't need to though, I think.

ietf/meeting/views.py
rjsparks: It's valid as long as we're modeling recordings as Document objects.

To date, we have not revised a recording document. Doesn't mean we couldn't (with the current model anyhow). An -01 for youtube would just point to a different URL the same as an -01 for a draft points to a different filename.

ietf/meeting/views.py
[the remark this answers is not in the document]
holloway: Done, it now looks like this

Screenshot from 2024-11-19 09-51-01

ietf/meeting/views.py
holloway: Thoughts on the delete button? Should it have a confirmation modal?

ietf/meeting/views.py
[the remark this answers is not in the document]
holloway: Based on the assumption that `session.meeting.date` is already UTC. Is that correct?

ietf/meeting/views.py
holloway: I'm going to assume we need a confirmation modal.

ietf/meeting/views.py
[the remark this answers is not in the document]
jennifer-richards: Sadly, I don't think so, it'll be the date in the meeting timezone.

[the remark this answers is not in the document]
jennifer-richards: `Meeting.start_datetime()` will give you a timezone-aware midnight in the meeting timezone. If you have a `Session` (which I think you do), then the way to get the start time of that session is:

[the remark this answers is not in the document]
jennifer-richards: Also, this needs a guard against `official_timeslotassignment()` not existing unless it's guaranteed to exist. It should exist for anything that is on an agenda.

ietf/meeting/views.py
```diff
@@ -2568,6 +2569,56 @@ def add_session_drafts(request, session_id, num):
...
+    if not session.can_manage_materials(request.user):
+        raise Http404
+    if session.is_material_submission_cutoff() and not has_role(request.user, "Secretariat"):
+        raise Http404
+
+    session_number = None
+    sessions = get_sessions(session.meeting.number,session.group.acronym)
+    initial = {
+        'title': f"Video recording for {session.group.acronym} on {session.meeting.date.strftime('%b-%d-%Y at %H:%M:%S')}"
+    }
```
jennifer-richards, line 2596: To be more concrete about my comments elsewhere, I think something like this is what is needed. The `assertion()` call is not bulletproof, there'll still be a 500 if this comes up.
```suggestion
    timeslot = session.official_timeslotassignment()
    assertion("timeslot is not None")
    initial = {
        'title': f"Video recording for {session.group.acronym} on {timeslot.utc_start_time().strftime('%b-%d-%Y at %H:%M:%S')}"
    }
```

ietf/meeting/views.py
rjsparks: I have minor heartburn that we're allowing a Document object to be deleted. It violates an assumption about what Document objects are _for_, but I suppose we have that problem with these recording documents-that-are-just-external-links already and some future project should migrate them to some other model instead.
 
I'll request one change and we can move forward with this.

ietf/templates/meeting/add_session_recordings.html
```diff
@@ -0,0 +1,113 @@
...
+                <tbody>
+                    <tr>
+                        <td colspan="3" class="text-center font-italic">(none)</td>
+                    </tr>
+                </tbody>
+            {% endif %}
+        </table>
+    </form>
+    <dialog id="delete_confirm_dialog">
+        <p>Really delete <a href="#" id="delete_confirm_link"></a>?</p>
```
rjsparks, line 59: Could we say `Really delete the link to` instead? We don't want to leave anyone with the impression that this will remove the recording at youtube.

rjsparks (changes requested):

ietf/templates/meeting/add_session_recordings.html
holloway: Ok, it looks like this now

Screenshot from 2024-11-20 07-33-26

rjsparks (approved):

rjsparks (changes requested):
I think this will work now (and it seems fine with manual testing). We need CI tests though.

rjsparks (changes requested):

ietf/meeting/utils.py
```diff
@@ -870,6 +870,14 @@ def create_recording(session, url, title=None, user=None):
 
     return doc
 
+def delete_recording(session, pk):
+    '''
+    Delete a session recording
+    '''
+    document = Document.objects.filter(pk=pk, group=session.group).first()
+    if document:
+        document.delete()
```
rjsparks, line 879: This is what caught my eye. This function would be better accepting a Document object rather than a pk, and perhaps should live in the Session model so that the connection to Session is naturally enforced.

ietf/meeting/views.py
jennifer-richards: Noticed I botched the `timeslotassignment.timeslot` step - glad you caught it!

jennifer-richards:
This is a suggestion following up @rjsparks' comments inline regarding deletion of the `Document` and the interface for doing so. The inline conversation was long enough I think this might be easier to follow.

I think the "Delete the link to..." action should delete the `SessionPresentation` that links to the `Document` rather than the `Document` itself. This avoids the worry about changing the historical policy about `Document` lifetimes. It also lets the `delete_recording()` action restrict itself to the `Session` rather than using `session.group` as a proxy and trusting the discretion/awareness of materials managers.

When deleting the `SessionPresentation`, the `Document` should have its "recording" state set to "deleted" rather than actually being deleted from the database.

As far as pk vs instances, I don't think the modeling will let us use `ModelForm` in the way @rjsparks was suggesting. If it can, that'd be better, but `SessionPresentation` is really just an annotated through model between `Document` and `Session`. If this was planned to be a traditional HTML forms interface, the right approach would be a `FormSet` that could do that magic, but as an endpoint for a JS front end I think we have to roll our own.

rjsparks:
This particular view isn't using a js frontend.

jennifer-richards:
> This particular view isn't using a js frontend.

Ah, I guess I'd mixed it up with work in other PRs. In that case we might want to use a formset

rjsparks (approved):

PR #8219, opened 2024-11-18
feat: Subscriptions

holloway:
* Removing old profile emails handling and replacing it with a link to...
* ...a new subscriptions UI in Vue (on a new route)

replaces #8074


holloway:
fyi @jennifer-richards it's pretty much ready to use a subscriptions API now

jennifer-richards:
> fyi @jennifer-richards it's pretty much ready to use a subscriptions API now

Great - working on that and hope to have something usable soon

holloway:
@rjsparks @jennifer-richards  I just wanted to check whether I've changed the email UX in the way you'd expect. The existing profile's page emails table was replaced with a link to manage subscriptions. Thoughts?

**before**
Screenshot from 2024-11-20 15-29-19

**after**
Screenshot from 2024-11-20 15-27-29

jennifer-richards:
Hm. I'd assumed there'd be a distinction between addresses linked to the person and subscriptions using those addresses. The addresses are used in other ways, and include addresses that may be entirely disabled. Some of the "other ways" are subscriptions of a sort if you squint, but not to mailing lists. (Thinking of things like alerts, draft update notifications, etc)

Not to say it can't work the way you've suggested, just that I'd need to see more about what's behind the Manage Subscriptions button and how that'd accommodate the other address features.

holloway:
That scope of removal was just a guess, and I have no particular opinion about the scope, so I'm happy to change it to whatever is decided.

It sounds like they still need to be able to add emails on this route. Do they also still need to change primary and active email?

I would be interested in seeing/developing a wireframe UX mockup before more development if that's possible? Any level of quality accepted.

rjsparks:
I think a realtime conversation in addition to sketching might make sense.
One bit of confusion that might be present - the addresses used for Roles are not subscriptions.
We have two concepts on the existing form and we're adding a third eventually:
1. The email addresses known for this Person and their usability state
2. The email addresses to use when something in the datatracker is going on that involves one of a Person's Roles (this is not just sending mail, but is also about what address to present in the UI when speaking about that Person in that Role).
3. (yet to be added) The set of mailman3 subscriptions managed for this Person

PR #8233, opened 2024-11-18
feat: django-rest-framework + Person/Email API

jennifer-richards:
This adds django-rest-framework to the datatracker and implements the first few API endpoints using it. These allow a user to retrieve their set of email addresses, update them, and request new email addresses.

I've put the datatracker API under `/api/core/` anticipating that we'll have other endpoints. A schema is served at `/api/schema/`

This does not yet enforce a single "primary" address. That should be enforced by the UI initially. I'll add an enhancement that prevents multiple addresses from being assigned as primary.

Permissions are done via session cookies. For the PATCH/POST/PUT API calls the CSRF token must be sent both in the `X-CSRFToken` header and in the `csrftoken` cookie.

Note that some existing code was refactored to share code with the API.

More work will be needed on this, but it's definitely time to get more eyes on it.


ietf/ietfauth/validators.py
```diff
@@ -0,0 +1,34 @@
...
+    if "anonymous" in name_without_spaces.lower():
+        raise forms.ValidationError("Please pick another name - this name is reserved.")
+
+
+def is_allowed_address(value):
+    """Validate that an address complies with datatracker requirements"""
+    for pat in settings.EXCLUDED_PERSONAL_EMAIL_REGEX_PATTERNS:
+        if re.search(pat, value):
+            raise ValidationError(
+                "This email address is not valid in a datatracker account"
```
rjsparks, line 33: I realize this is old, but is it conflating User.username concerns with Email.address concerns? If so, we should clarify.

jennifer-richards: I think this is doing things like preventing `some-wg-list@ietf.org` from being signed up as an email. I don't think it's for the username concerns.

rjsparks (approved):

jennifer-richards:
The new API needs tests and some work on view naming / namespacing for URL reversing, but I'm going to leave those as separate work so we can move forward on the feat branch.

PR #8238, opened 2024-11-19
fix: show meetecho recordings for groups with more than one session

### LINKED ISSUE #8236: Proceedings only show one session recording link for groups that had multiple sessions at 121

rjsparks:
Fixes #8236


jennifer-richards (approved):
lgtm. The only fault I see is the very opaque way that `_format_materials()` filters out non-Truthy materials, but that's for another day...

PR #8244, opened 2024-11-20
fix: clarify replace wording

ghwood:


rjsparks:
Thanks for pushing on this. Some thoughts below:

> An I-D submission will be automatically attached to the previous version of the I-D with the same name, or assigned "-00" version if no previous version exists. If this submission is intended to replace an I-D with a different name (e.g. when an individual I-D is adopted as a Working Group document), please enter that name here. (Approval required to replace an I-D for which you are not an author.)

The first sentence doesn't work for me - it sees to imply "-00" is a thing that happens at this point, but -00 was chosen by the submitter long before the submitter gets to this form.

I also don't think "automatically attached to a previous" is what happens, or is the right model of what's happening for other people to build. Instead, we need to reinforce that there are versions of an Internet-Draft, and that an Internet-Draft might get replaced with some other Internet-Draft(s).

Maybe imperative statements rather than explanation would work better? My alternate suggestion:

-----

Do not enter anything here if you are only submitting the next version of your Internet-Draft. Only enter items here if this submission is intended to replace an I-D with a different name. A typical use of this field is to note what individual I-Ds are replaced by a new -00 Working group I-D. Note that additional approval will be required to replace an I-D for which you are not an author.

rjsparks (approved):

PR #8292, opened 2024-12-02
fix: test file existence using metadata

rjsparks:
Without something like this, each view of a draft's main page reads the entire file twice.

### DIFF
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -546,8 +546,12 @@ class DocumentInfo(models.Model):
         path = Path(self._text_path())
         if not path.exists():
             return None
-        with path.open('rb') as file:
-            raw = file.read(size)
+        try:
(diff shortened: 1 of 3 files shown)


jennifer-richards (changes requested):
This is really begging to be refactored the rest of the way using `pathlib.Path`. I won't insist, though.

However, `text_exists()` is not used - I assume you have a use case in mind? Seems like the call should come in with the function.

rjsparks:
More significant refactoring coming in with the blob work, but in the meantime:...

jennifer-richards (dismissed):
The change from catching IOError to testing path.exists() introduces a (tiny) window for a race. It may also mean additional exotic read failures turn into server errors. I think that is ok, but calling it out in case you think it might come up when the server is operating normally.

jennifer-richards (approved):

PR #8300, opened 2024-12-05
fix: disable raw/include in RST

jennifer-richards:


kesara (approved):
LGTM

Another recommendation in [Deploying Docutils Securely](https://docutils.sourceforge.io/docs/howto/security.html), is to lower the `line_length_limit` [^1]. Default value is `10 000`. Not sure that can cause any issues.

[^1]: https://docutils.sourceforge.io/docs/howto/security.html#cpu-and-memory-utilization

rjsparks (approved):

jennifer-richards:
> Another recommendation in [Deploying Docutils Securely](https://docutils.sourceforge.io/docs/howto/security.html), is to lower the `line_length_limit` [1](#user-content-fn-1-97a43afdd1a6bff6281d610f248614f8). Default value is `10 000`. Not sure that can cause any issues.

I don't have any idea what a more reasonable limit would be. Given the access controls guarding places we accept user input to the RST processor and the hopefully limited lifespan of supporting RST, I think we can leave it as it is.

Issue #3712, opened 2022-03-21
Remove the /stats module


rjsparks:
The /stats module as it currently exists is an attractive nuisance. It was intended to move the nice things that Jari has been maintaining into the datatracker, but the implementation didn't really achieve that.

As it is, what's presented ranges from incomplete to actually wrong.

We're going to remove it and schedule a new effort to build better statistics.

rjsparks:
There are a few things, like the MeetingRegistration model, that should have been in other apps to begin with that will have to be moved.

rjsparks:
Tero mentioned at <https://mailarchive.ietf.org/arch/msg/tools-discuss/0Tw_KT4pYdUxoIf36p8yzn_bvEM> that some parts of the existing implementation may remain useful, and we should review whether/what to move into the relevant apps while waiting for re-implementation.

larseggert:
Are you still interested in a PR that rips out the document stats?

rjsparks:
Yes, and the meeting stats. It would be best if the same PR moved the review-team stats into the review application.

jennifer-richards:
The `stats` app still exists, but has been mostly removed. Closing this. The work to replace it is tracked as #8511

PR #8307, opened 2024-12-05
fix: remove unreliable statistics

rjsparks:
Partially addresses #3712


ietf/settings.py
```diff
@@ -809,8 +809,8 @@ def skip_unreadable_post(record):
 SESSION_REQUEST_FROM_EMAIL = 'IETF Meeting Session Request Tool <session-request@ietf.org>' 
 
 SECRETARIAT_SUPPORT_EMAIL = "support@ietf.org"
-SECRETARIAT_ACTION_EMAIL = "ietf-action@ietf.org"
-SECRETARIAT_INFO_EMAIL = "ietf-info@ietf.org"
+SECRETARIAT_ACTION_EMAIL = SECRETARIAT_SUPPORT_EMAIL
```
jennifer-richards, line 812: What were the replaced addresses doing? Old aliases in the mail system?

jennifer-richards (approved):
One question inline, but that's just curiosity.

ietf/settings.py
rjsparks: the secretariat used to use these aliases to drop tickets into separate queues for different types of requests. Now, everything is intended to go support@ instead, with queuing happening later, so that the community doesn't have to figure out which of many addresses to use. We have code that could be refactored to simplify having this many settings, but that should be a separate PR.

rjsparks: also - the other aliases just forward to support@ currently

Issue #8212, opened 2024-11-15
Unable to edit IESG state for an expired draft


cindymorgan:
### Describe the issue

draft-ietf-roll-nsa-extension is expired, but it doesn't show up on the list of expired drafts for ROLL. As is pretty much always the case when a doc doesn't show up on a WG's page, the document expired at some point after an AD had returned it to the WG and put it into the IESG state of AD is Watching, so after expiring it went into the IESG state of Expired (IESG: Dead).

For a document that wasn't expired, I would fix that by changing the IESG state back to I-D Exists. However, when I try to do that for this document, I get a 404 error when clicking on the button to edit the IESG state (https://datatracker.ietf.org/doc/draft-ietf-roll-nsa-extension/edit/state/).

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
The issue is with some outdated logic at https://github.com/ietf-tools/datatracker/blob/c89646ef8d00d23cd9914f457ee5114900307a3e/ietf/doc/views_draft.py#L98-L99

cindymorgan:
Another doc in a similar state: draft-ietf-idr-wide-bgp-communities (they have been asked to submit a new version to un-expire this in order to clean this up.)

PR #8321, opened 2024-12-10
fix: keep draft-iesg state on expiration. Update action holders.

### LINKED ISSUE #5560: On draft-expiration, documents with IESG state shouldn't change IESG state.

rjsparks:
Fixes #5560 and #8212


rjsparks:
Keeping at draft as #8212 fix isn't complete and to investigate if we can remediate docs recently moved to dead just because of expiry.

ietf/doc/tests_draft.py
```diff
@@ -843,6 +846,69 @@ def test_clean_up_draft_files(self):
...
+                d.docevent_set.count(), dead_from_expires_event_count[d] + 2
+            )
+            self.assertIn(
+                "due only to document expiry", d.latest_event(type="added_comment").desc
+            )
+            self.assertEqual(
+                d.latest_event(StateDocEvent).desc,
+                "IESG state changed to <b>I-D Exists</b> from Dead",
+            )
+        self.assertEqual(len(outbox), 0)
```
jennifer-richards, line 909: If this is meant to test that community notifications aren't sent, additional work will be needed because the signal handler doesn't actually send messages when running tests. (I'm not that's what you're checking)

ietf/doc/expire.py
```diff
@@ -238,3 +227,42 @@ def move_file_to(subdir):
         except Document.DoesNotExist:
             # All uses of this past 2014 seem related to major system failures.
             move_file_to("unknown_ids")
+
+
+def repair_dead_on_expire():
+    by = Person.objects.get(pk=1)  # a.k.a. "(System)"
```
jennifer-richards, line 233: We canonically use `Person.objects.get(user__username="(System)")`. Better to stick to that?

jennifer-richards (changes requested):
One minor change suggested (getting the system Person by name instead of pk, though if you want to start a change in our practice that's fine).

More significantly, it looks to me like you might be trying to check that notification emails aren't being sent in a way that I don't think tests that. This might be my misunderstanding. If you are trying to check that, I think you need to test that the `skip_community_list_notification` flag is set on the event you save. We already have a test that proves the flag works as intended. However, getting the event instance might be a bit involved.

ietf/doc/expire.py
rjsparks: looks like it's mostly `(name="(System)")` rather than `user__username`?

jennifer-richards: Er, yes, `name`. Sorry, I don't know why I thought I saw that.

ietf/doc/tests_draft.py
```diff
@@ -886,7 +886,11 @@ def test_repair_dead_on_expire(self):
             dead_from_expires.append(d)
             dead_from_expires_event_count[d] = d.docevent_set.count()
 
-        empty_outbox()
+        notified_during_factory_work = mock_notify.call_count
+        for call_args in mock_notify.call_args_list:
+            e = call_args.args[0]
```
jennifer-richards, line 891: Wow! I had missed that `call_args` now has `args` and `kwargs` properties! That is so much better than `call_args[0][0]`.

jennifer-richards (approved):
Looks good.

An alternative to counting events and manipulating indexes would be to call `mock_notify.reset_mock()` before calling your method under test. That'll empty `call_args` and reset `called` to False. Purely optional though.

PR #8323, opened 2024-12-10
feat: Increase indent of figures

martinthomson:
This carries some risk, but on balance, I think that it is a good change.

We see a lot of documents that include sourcecode or ASCII art that appears at the same indent level as the rest of the text.  This makes it hard to visually distinguish figures.

This increases the indent for these, where possible.  It retains the indentation flexbox hack that ensures that wider figures reduce the margin.

The risk is that this will double the indent where people have manually added one.  The text formatter of xml2rfc doesn't add an indent, so this is also an inconsistency in appearance.  My hope is that xml2rfc will be updated to match, with people learning not to add their own indentation to documents.


Michealinc (commented):
Explict

jennifer-richards (approved):
I think the risk you describe is just that the art is doubly indented as a block and not that it's left illegible, is that right? Assuming so, sounds fine to me.

(It did not really take a year of staring at the one character changed here to come up with that, I just expected someone with a stronger opinion about the risk would get back to it before it fell off my radar!)

martinthomson:
Yeah, I think that the ideal state is to have this consistent across formats.  That means a corresponding change in xml2rfc.  Otherwise, this just looks inconsistent.  It should all be readable still.  The existing logic in xml2rfc for wider artwork will remove extra indentation if necessary.  (And the flexbox haxx here will do the same.)

jennifer-richards:
From tools team discussions I think we're going to wait for clarity on changing xml2rfc to match this before merging. This will remain idle but not forgotten in the meantime.

PR #8338, opened 2024-12-14
fix: homepage logo width

### LINKED ISSUE #8308: Mobile view improvements
### Describe the issue

IETF logo doesn't seem to be resizing on the homepage.

The screenshot is from Safari on iPhone 13 mini.
Same results on Firefox (on macOS) with responsive design mode.

Image

### Code of Conduct

NGPixel:
Fixes #8308

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -236,8 +236,11 @@ th,
 
 // Helper to constrain the size of the main logo
 .ietflogo {
-    width: 75%;
+    width: 100%;
     max-width: 300px;
+}
+.ietflogo > img {
+    width: 100%;
 }
 
 // Make revision numbers pagination items fixed-width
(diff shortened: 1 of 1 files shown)


ietf/static/css/ietf.scss
```diff
@@ -236,9 +236,12 @@ th,
 
 // Helper to constrain the size of the main logo
 .ietflogo {
-    width: 75%;
```
rjsparks, line 239: @larseggert Where did this .75 originally come from?

larseggert: No idea. But that used to look correct with the old CSS. 🤷

ietf/static/css/ietf.scss
```diff
@@ -236,9 +236,12 @@ th,
...
+    width: 100%;
     max-width: 300px;
```
rjsparks, line 240: would it make sense to set a min-width (of, perhaps, `100px`) here as well?

limck5856: Ok

rjsparks (approved):

PR #8347, opened 2024-12-18
feat: copy the 1wg files to so... many... places...

rjsparks:


jennifer-richards (approved):
If this were likely to be long-lived, I'd prefer refactoring `CHARTER_COPY_PATH` as a list. However, as I don't think that's true and we want to encourage adding more dests, this is fine.

rjsparks:
>and we want to encourage adding more dests
🕶️ 

>If ... refactoring ... list ...

That's the refactor I started down that led to me asking about typechecking iterables, but we would have to gut the current implemenation of the utility that creates temporary directories at settings names to make that work.

I'm still contemplating creating a new file/class to handle external persistence and isolate all the concerns there, at which point the structure of the relevant settings could be entirely reconsidered.

In any case, when we stop having directories, the tmp directory solution in the utility will be obviated.

jennifer-richards:
> >and we want to encourage adding more dests
> 
> 
> 
> 🕶️ 
> 

Oops, I left out the "don't"

PR #8348, opened 2024-12-18
chore: remove single-use repair_dead_on_expire task

rjsparks:


jennifer-richards (changes requested):
Should also remove the `repair_dead_on_expire()` function from `expire.py` I think?

jennifer-richards (dismissed):

jennifer-richards (approved):

PR #8401, opened 2025-01-07
fix: keep the ftp dir in sync when the datatracker creates artifacts

rjsparks:


ietf/submit/checkers.py
```diff
@@ -280,6 +281,15 @@ def check_file_txt(self, path):
 
             dest = os.path.join(settings.SUBMIT_YANG_DRAFT_MODEL_DIR, model)
             shutil.move(path, dest)
+            ftp_dest = Path(settings.FTP_DIR) / "yang" / "draftmod" / model
```
jennifer-richards, line 284: We write models when we do the checking? That seems odd. Shouldn't that happen when the draft is posted? (Not advocating a change as part of this PR, just wondering)

jennifer-richards (approved):
Looks like it does the advertised things to me.

One question about whether the underlying behavior in the yang checker is good practice, but that's more of an aside.

PR #8404, opened 2025-01-08
feat: monitoring API

jennifer-richards:
Adds a set of endpoints at `/api/core/monitoring/<flavor>/` to allow various types of monitoring to be served.

Considered using a viewset but since this is really not a resource-oriented endpoint, it was cleaner to implement it as a standalone `APIView`. I've put it in the same URL prefix-space that the (commented-out) `core_Router` uses because it's part of the same API. This means collisions between the router and the explicit route are possible, but I don't think this will be a big issue.

Draft PR for a monitoring API using django-rest-framework. This is as much for discussion as actual merging.

### DIFF
--- a/ietf/api/serializers.py
+++ b/ietf/api/serializers.py
@@ -1,0 +1,10 @@
+# Copyright The IETF Trust 2025, All Rights Reserved
+"""Serializers for django-rest-framework"""
+from rest_framework import serializers
+
+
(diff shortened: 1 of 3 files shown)


jennifer-richards:
OBE - going in a different direction because the monitoring interface is not a good match for a general purpose API approach

PR #8405, opened 2025-01-09
feat: additional filesystem monitoring

rjsparks:


ietf/api/urls.py
```diff
@@ -82,6 +82,8 @@
     url(r'^version/?$', api_views.version),
     # Application authentication API key
     url(r'^appauth/(?P<app>authortools|bibxml)$', api_views.app_auth),
+    # NFS metrics endpoint
+    url(r'^nfs_metrics/?$', api_views.nfs_metrics),
```
jennifer-richards, line 86: bikeshed/overcomplication: maybe `metrics/nfs` to leave a path to other metrics?

dev/deploy-to-container/settings_local.py
```diff
@@ -64,6 +64,7 @@
 BIBXML_BASE_PATH = '/assets/ietfdata/derived/bibxml'
 IDSUBMIT_REPOSITORY_PATH = INTERNET_DRAFT_PATH
 FTP_DIR = '/assets/ftp'
+TMP_DIR = '/assets/tmp'
```
jennifer-richards, line 67: Sort of a bike shed, but this is really a special purpose path. Calling it `TMP_DIR` might tempt someone to use it for random temp files when they'd be better off using a path not served over nfs. Maybe `NFS_TEST_DIR` (or `NFS_TMP_DIR` to at least emphasize its nature)

jennifer-richards (changes requested):
I think the `TMP_DIR` name should probably be refined - the other suggestion is less important

jennifer-richards (approved):

PR #8408, opened 2025-01-10
feat: new meeting registration implementation

rpcross:
New implementation of meeting registrations. Moved from stats.MeetingRegistration to meeting.Registration and meeting.RegistrationTickets. Also new API and management command, migrate_registrations, to copy MeetingRegistration records to Registration.

This is just the first step in a series. The management command can be called from cron regularly to populate the new table, Registration, from the old. Future commits will migrate all the existing code that uses MeetingRegistration to using Registration. Once that is complete the Registration system can be changed to use the new API, stop running the migration, then remove the old table and API.

This PR replaces https://github.com/ietf-tools/datatracker/pull/7724


ietf/meeting/utils.py
```diff
@@ -997,3 +998,91 @@ def participants_for_meeting(meeting):
...
+                    'person': preferred_reg.person,
+                    'attended': preferred_reg.attended,
+                    'checkedin': preferred_reg.checkedin,
+                }
+            )
+            if created:
+                for meeting_reg in meeting_regs:
+                    reg.tickets.create(
+                        attendance_type_id=meeting_reg.reg_type,
+                        ticket_type_id=meeting_reg.ticket_type,
```
jennifer-richards, line 1053: There are a number of values in `meeting_reg.reg_type` and `.ticket_type` that aren't replicated in the new `AttendanceTypeName` and `TicketTypeName` world. Do you have a plan to rationalize these?

From the staging database:
```python
>>> pp(Counter(MeetingRegistration.objects.values_list("ticket_type", flat=True)))
Counter({'': 45288,
         'week_pass': 13756,
         'full_week_pass': 3204,
         'one_day': 2523,
         'hackathon_combo': 1653,
         'student': 1359,
         'hackathon_only': 680,
         'anrw_combo': 494,
         'anrw_only': 13,
         'one_day week_pass': 2,
         'one_day full_week_pass': 1,
         'full_week_pass one_day': 1,
         'student full_week_pass': 1})
>>> pp(Counter(MeetingRegistration.objects.values_list("reg_type", flat=True)))
Counter({'': 33592,
         'onsite': 14908,
         'remote': 14835,
         'hackathon_onsite': 3068,
         'hackathon': 1397,
         'hackathon_remote': 668,
         'anrw_onsite': 507})
```

jennifer-richards (commented):
I still have a few things to think about, but wanted to get some feedback to you. I like the direction. We'll need to create a feat branch to receive the series of PRs before merging to main. Maybe `feat/registration` would be a good name?

ietf/meeting/models.py
```diff
@@ -1474,3 +1474,40 @@ class Meta:
...
+        return self.get_queryset().filter(registrationticket__attendance_type__slug='remote').exclude(registrationticket__attendance_type__slug='onsite')
+
+class Registration(models.Model):
+    """Registration attendee records from the IETF registration system"""
+    meeting = ForeignKey(Meeting)
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+    affiliation = models.CharField(blank=True, max_length=255)
+    country_code = models.CharField(max_length=2)        # ISO 3166
+    person = ForeignKey(Person, blank=True, null=True)
```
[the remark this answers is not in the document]
rpcross: fixed in commit 5c1631bdd5c08ba73023580263aa7442da9fde43

ietf/meeting/models.py
```diff
@@ -1474,3 +1474,40 @@ class Meta:
...
+    # custom manager
+    objects = RegistrationManager()
+
+    def __str__(self):
+        return "{} {}".format(self.first_name, self.last_name)
+
+class RegistrationTicket(models.Model):
+    registration = ForeignKey(Registration, related_name='tickets')
+    attendance_type = ForeignKey(AttendanceTypeName)
+    ticket_type = ForeignKey(RegistrationTicketTypeName)
```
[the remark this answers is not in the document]
rpcross: fixed in commit 5c1631bdd5c08ba73023580263aa7442da9fde43

ietf/meeting/utils.py
rpcross: I see. The differences are older outlier records that I will need to handle in the migration command, prior to the main migration.

rpcross: Addressed in migration. NOTE: the first run of migrate_registrations must be run with --initial option.
Fixed in commit 8eb4b26242cfbb28140de0de8fe6be4124a9a96d

jennifer-richards: It looks like you've cleaned up `reg_type` there, but I think there are still `ticket_type` values that will fail to match up with `attendance_type_id`. I.e., `full_week_pass`, `one_day week_pass`, `one_day full week pass`, `full_week_pass one day`, and `student full_week_pass`. Or are those handled somewhere I'm missing?

rpcross: Ah, I was accounting for that but had a mistake in my migration function. Addressed in commit [467667e](https://github.com/ietf-tools/datatracker/pull/8408/commits/467667efda92c511d0eeba5bcc34588f058ca996)

rpcross: Now the migration fails because older MeetingRegistration records don't have ticket_type, pre 116, or don't have reg_type, pre 99. Where did these records come from? Maybe scraping the registration site. Should we use defaults for the values, reg_type="onsite" (depending on scraping). But what about ticket_type? Or allow one or both of these RegistrationTicket fields to be blank?

jennifer-richards (changes requested):
Looking good. For the record, we discussed out-of-band whether the `initial=True` branch of the migration utility would be better as a Django migration or done manually, but I think it's better to keep it as it's done here.

I think the `names.json` needs an update. Would also be good to combine the migrations (i.e., remove them and regenerate)

One other comment inline.

ietf/meeting/utils.py
```diff
@@ -1109,6 +1109,17 @@ def migrate_registrations(initial=False):
     return
 
 
+def test_migrate_registrations():
```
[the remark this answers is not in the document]
rpcross: I wrote this for my own use, to confirm that migrate_registration was working as expected in testing. I just saved it here for reference, and possible use in production.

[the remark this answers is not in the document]
rjsparks: If we want to keep it, wrap it in a task that can be run. Agree with renaming it to something like "verify_registration_migrated_data_are_correct".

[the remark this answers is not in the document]
rpcross: I think I have addressed all comments above.

Michealinc (commented):
Explict

jennifer-richards (approved):

PR #8411, opened 2025-01-10
fix: correctly disconnect post_save signal

jennifer-richards:


rjsparks:
Closing as OBE - we removed the file from the project instead.

PR #8429, opened 2025-01-16
feat: write objects to blob storage

rjsparks:


ietf/doc/expire.py
```diff
@@ -225,6 +225,13 @@ def move_file_to(subdir):
             mark = Path(settings.FTP_DIR) / "internet-drafts" / basename
             if mark.exists():
                 mark.unlink()
+            if ext:
+                # Note that we're not moving these strays anywhere - the assumption
+                # is that the active-draft blobstore will not get strays.
+                # See, however, the note about "major system failures" at "unknown_ids"
+                blobname = f"{ext[1:]}/{basename}"
+                if exists_in_storage("active-draft", blobname):
+                    remove_from_storage("active-draft", blobname)
```
jennifer-richards, line 234: Just call `remove_from_storage()` and handle an exception if it complains about blobs that don't exist? The code needs to handle that possibility anyway, e.g., due to a race.

jennifer-richards (commented):
Sorry if this a comment on code not intended to be ready for review, but I happened to look... one comment inline. 

Also, for testing I wonder if we want a fake blob Storage class that does only what we need for that it's being called as intended. If we make that work with your `storage_utils`, the tests wouldn't even have to know. If the need arose, we could also test against a real blob store by adjusting the configuration.

ietf/doc/expire.py
rjsparks: The implemenation in django-storages (and in boto3) explicitly say it is not an error to remove something that doesn't exist.

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+            log(f"Failed to save {kind}:{name} - name already exists in store")
+            debug.show('f"Failed to save {kind}:{name} - name already exists in store"')
+            debug.traceback()
+            raise Exception("Not ignoring overwrite attempts while testing")
+        else:
+            try:
+                new_name = self.save(name, file)
+                now = timezone.now()
+                existing_record = StoredObject.objects.filter(store=kind, name=name)
+                if existing_record.exists():
```
jennifer-richards, line 44: Should use `update_or_create()`

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+                        created=now,
+                        modified=now,
+                        doc_name=doc_name,
+                        doc_rev=doc_rev,
+                    )
+                if new_name != name:
+                    complaint = f"Error encountered saving '{name}' - results stored in '{new_name}' instead."
+                    log(complaint)
+                    debug.show("complaint")
+                    # Note that we are otherwise ignoring this condition - it should become an error later.
```
jennifer-richards, line 67: Might as well raise an exception now since we have the log-and-swallow in place

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+                log(f"Failed to save {kind}:{name}", e)
+                raise e
+                debug.show("type(e)")
+                debug.show("e")
+                debug.traceback()
+            finally:
+                del self.in_flight_custom_metadata[name]
+        return None
+
+    def exists_in_storage(self, kind: str, name: str) -> bool:
```
jennifer-richards, line 78: again, why not override?

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+        try:
+            # open is realized with a HEAD
+            # See https://github.com/jschneier/django-storages/blob/b79ea310201e7afd659fe47e2882fe59aae5b517/storages/backends/s3.py#L528
+            with self.open(name):
+                return True
+        except FileNotFoundError:
+            return False
+
+    def remove_from_storage(
```
jennifer-richards, line 87: ... and again ... override the remove method?

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+        now = timezone.now()
+        try:
+            with self.open(name):
+                pass
+            self.delete(name)
+            # debug.show('f"deleted {name} from {kind} storage"')
+        except FileNotFoundError:
+            if warn_if_missing:
+                complaint = (
+                    f"WARNING: Asked to delete non-existant {name} from {kind} storage"
```
jennifer-richards, line 99: spelling: exist**e**nt

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+        else:
+            # Note that existing_record is a queryset that will have one matching object
+            existing_record.update(deleted=now)
+
+    def _get_write_parameters(self, name, content=None):
+        # debug.show('f"getting write parameters for {name}"')
+        params = super()._get_write_parameters(name, content)
+        if "Metadata" not in params:
+            params["Metadata"] = {}
+        if not is_seekable(content):
```
jennifer-richards, line 117: `is_seekable` is [a strange method](https://github.com/jschneier/django-storages/blob/b79ea310201e7afd659fe47e2882fe59aae5b517/storages/utils.py#L122) - it returns `True` if the `content` does not have a `seekable()` method, so e.g. `is_seekable("strings don't seek")` is `True`. Wrapping the `content.seek(0)` in try/except would be a better guard than using the helper.

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+                log(complaint)
+                debug.show("complaint")
+        existing_record = StoredObject.objects.filter(store=kind, name=name)
+        if not existing_record.exists() and warn_if_missing:
+            complaint = f"WARNING: Asked to delete {name} from {kind} storage, but there was no matching StorageObject"
+            log(complaint)
+            debug.show("complaint")
+        else:
+            # Note that existing_record is a queryset that will have one matching object
+            existing_record.update(deleted=now)
```
jennifer-richards, line 110: This probably ought to add a `.filter(deleted__isnull=True)` to the queryset for the `update()`. That will keep `deleted` as the _earliest_ timestamp if a file is deleted more than once.

(We don't want the deleted filter for the `warn_if_missing` existence check, though)

jennifer-richards (commented):
Posting as comment to share these while I continue reviewing

ietf/doc/models.py
```diff
@@ -1538,3 +1560,31 @@ class BofreqEditorDocEvent(DocEvent):
...
+    created = models.DateTimeField(
+        null=False,
+        help_text="Instant object became known. May not be the same as the storage's created value for the instance. It will hold ctime for objects imported from older disk storage"
+    )
+    modified = models.DateTimeField(
+        null=False,
+        help_text="Last instant object was modified. May not be the same as the storage's modified value for the instance. It will hold mtime for objects imported from older disk storage unless they've actually been overwritten more recently"
+    )
+    doc_name = models.CharField(max_length=255, null=True, blank=True)
+    doc_rev = models.CharField(max_length=16, null=True, blank=True)
```
jennifer-richards, line 1581: This is not necessariiy a request for a change, but I'm having misgivings about this quasi-ForeignKey link to `Document`. We've talked about it a bit, but I wanted to write it down because it came to mind again. I think it's ok to proceed with the name/rev fields initially, but I just get a tingling that it's going to turn into another wart on the already hard-to-work-with Document class hierarchy that we'll be complaining about later.

Would this be better implemented as a many-to-many between `Document` and `StoredObject`? We landed here because `Document` gets mixed up in `DocHistory`, but there could _also_ be a many-to-many that links `DocHistory` and `StoredObject` that is populated when a `Document` becomes a `DocHistory`. The usual way we save docs should handle this for us, since it has logic to deal with many-to-many fields.

This is a little bit unpleasant because it's really a one-to-many between `Document` and `StoredObject`, and then a many-to-many between `DocHistory` and `StoredObject`. If that really bothers us, we could handle that by customizing the handling along the lines we do for `DocumentAuthor`.

Are there objections to doing this beyond the complexity involved in copying the many-to-many field around?

ietf/doc/storage_backends.py
```diff
@@ -0,0 +1,137 @@
...
+from ietf.utils.timezone import timezone
+
+
+class CustomS3Storage(S3Storage):
+
+    def __init__(self, **settings):
+        self.in_flight_custom_metadata: Dict[str, Dict[str, str]] = {}
+        return super().__init__(**settings)
+
+    def store_file(
```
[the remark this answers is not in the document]
rjsparks: Mostly to keep the structure parallel for ease of readability. But you trigger several thoughts with the question.

* I think it would be a loss to let the general datatracker code call the storage's save directly - nothing outside the utility managing the storages should know it's actually storages underneath. Specifically, I want to avoid the assumption that there's only _one_ storage or storage strategy underneath.
* We may _have_ to override save here anyhow if we want things that use the *Field implementations which _do_ expect to be bare on top of a Storage to get this metadata handling.

ietf/doc/storage_backends.py
rjsparks: again, parallel structure. Also I think we want to (for the short term) keep the parent `.exists()` bare outside this overridden class just for continued learning.

ietf/doc/storage_backends.py
rjsparks: leaving this wart for now - need to think through what it should raise when we take away log-and-swallow.

PR #8431, opened 2025-01-17
ci: dev/test blobstore image

rjsparks:


.github/workflows/build-devblobstore.yml
```diff
@@ -0,0 +1,50 @@
...
+      env:
+        DOCKER_BUILD_NO_SUMMARY: true
+      with:
+        context: .
+        file: docker/devblobstore.Dockerfile
+        platforms: linux/amd64,linux/arm64
+        push: true
+        build-args: MINIO_VERSION=${{ env.MINIO_VERSION }}
+        tags: |
+          ghcr.io/ietf-tools/datatracker-devblobstore:${{ env.MINIO_VERSION }}
```
jennifer-richards, line 49: this will tag as :latest twice when MINIO_VERSION=latest. That will probably work, but maybe want a `minio-` prefix on that one?

jennifer-richards (dismissed):

NGPixel (dismissed):

.github/workflows/build-devblobstore.yml
rjsparks: (result of out-of-band discussions): we'll leave this be unless it turns out to not work as expected.

NGPixel (approved):

PR #8443, opened 2025-01-20
chore: remove deprecated_api_upload_bluesheet method

jennifer-richards:
This option for uploading bluesheets was deprecated in mid-2023. Should confirm with @alexamirante that they're no longer using it.


alexamirante:
I confirm you can get rid of it.

rjsparks (approved):

PR #8449, opened 2025-01-21
feat: keep proceedings cache up to date via celery

jennifer-richards:
This still needs at least updates to the tests, but wanted to get it out for feedback before going further.

The idea here is to cache all the IETF meeting proceedings for a day and refresh the cache via celery. To avoid long-running celery jobs, the refresh task is designed to run hourly but only to update a couple proceedings at a time. We'll hopefully have a better method for managing the proceedings content in place before we reach more than 2-3 proceedings falling into any bin.

There were a few date parameters being supplied to the proceedings template but it looks to me like those must be relics. I traced the templates and don't see where any of the things I removed were being used indirectly. In view of that, it may be possible to adjust the boundary of the cached HTML fragment.

To avoid serving misinformation around the time when a meeting's proceedings are finalized, I've added that value to the cache key. This may result in one or a few long-running HTTP requests if we get a few people requesting the proceedings after the switch. We could consider refreshing the current proceedings immediately upon finalizing to avoid / reduce the cross section to that.


rjsparks:
This looks on target to me

rjsparks (dismissed):

rjsparks (approved):

PR #8471, opened 2025-01-27
feat: consolidate HTML sanitizing

jennifer-richards:
This drops `lxml.html.clean.Cleaner`, which was used for sanitizing uploaded HTML, instead using bleach for all sanitizing. Refactors the organization of bleach settings, moving all the cleaning into `ietf.utils.html` and leaving only linkify-related settings in `ietf.utils.text`.

This refactor drops code that added a `<meta charset="utf-8">` to the `<head>` when the latter was present in a cleaned upload. It probably also did away with some incidental cleanup that `lxml` was doing for us under some circumstances. One example I came across was that a test clean of
```html
<html><body>Text in the body element</body></html>
```
became
```html
<html><body><p>Text in the body element</p></body></html>
```
This only happened in some code paths and was neither tested nor documented, so I don't think it's a big loss. If we need that level of cleanup, we should add it as a proper feature.


jennifer-richards:
Tagging @NGPixel @holloway since we talked about this

NGPixel (dismissed):

NGPixel:
Python code looks pythony enough

rjsparks (approved):

PR #8492, opened 2025-01-31
feat: add more rfc api fields (many stubs)

jennifer-richards:
This adds the extra fields [requested here](https://github.com/ietf-tools/rfced-www/blob/main/client/utilities/rfc.mocks.ts#L107). Many of these are empty stubs, and a couple have more structure than the mock implementation. This is intended as an incremental step to allow progress on the rfced-www client side, not a finished piece of the server-side API. I'm expecting there'll be some iteration (either in this PR or in later updates) on the format / data included in this API call.

The `is_also` field (here, always an empty list) uses the historic "bad" name that represents "is-in-a-subseries-document". That will need to be implemented and given a more appropriate name.

Some other pieces can't or shouldn't be implemented yet. The "keywords" and "see also" fields don't (afaik) exist in the datatracker yet, so there's no data to include. The "formats" field could probably be implemented, but it'd currently mean doing a whole lot of looking at the file system and will be changing with the blob store work.

I'm not sure how to represent "errata" at this point, or whether it's needed at the index level.

The `draft` field will probably need some optimization - I haven't profiled, but I suspect it's generating N+1 queries to look up the draft name/title. That can easily be fixed with some prefetching, but I'm holding off until we know what it really need to include in the response.

Generally speaking, I continue to worry a bit about providing an API call that potentially includes large amounts of data about large numbers of documents. If there are fields here that could be moved to the detail views instead of the index views (i.e., are only needed for a single RFC view, not for the search results / index page), that would be a good thing to keep in mind.

### DIFF
--- a/ietf/doc/api.py
+++ b/ietf/doc/api.py
@@ -1,6 +1,6 @@
 # Copyright The IETF Trust 2024, All Rights Reserved
 """Doc API implementations"""
-from django.db.models import OuterRef, Subquery, Prefetch
+from django.db.models import OuterRef, Subquery, Prefetch, Value, JSONField
 from django.db.models.functions import TruncDate
 from django_filters import rest_framework as filters
 from rest_framework import filters as drf_filters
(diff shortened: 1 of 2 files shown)


holloway (approved):
I don't know how I missed this!

LGTM!

rjsparks (approved):

PR #8535, opened 2025-02-14
fix: remove all staging files for submission

### LINKED ISSUE #8534: Not all staging files are cleaned up when a submission is cancelled
### Describe the issue

When a submission is cancelled, it is intended that all of its artifacts will be removed. This is not done correctly now. The code in `ietf.submit.utils.remove_staging_files()` is called with the list of extensions in `submission.file_types` and only these files are removed. However, this list only includes files that the submitter uploaded - any files generated by datatracker are left over. This often includes `.txt` and I think always includes `.html`.

This will likely be OBE if #4334 is resolved in a way that directly links submission artifacts to the `Submission` object.

### Code of Conduct

jennifer-richards:
Fixes #8534

### DIFF
--- a/ietf/submit/utils.py
+++ b/ietf/submit/utils.py
@@ -671,20 +671,16 @@ def move_files_to_repository(submission):
             raise ValueError("Intended to move '%s' to '%s', but found source and destination missing.")
 
 
-def remove_staging_files(name, rev, exts=None):
-    """Remove staging files corresponding to a submission
-    
-    exts is a list of extensions to be removed. If None, defaults to settings.IDSUBMIT_FILE_TYPES.
-    """
-    if exts is None:
-        exts = [f'.{ext}' for ext in settings.IDSUBMIT_FILE_TYPES]
+def remove_staging_files(name, rev):
+    """Remove staging files corresponding to a submission"""
     basename = pathlib.Path(settings.IDSUBMIT_STAGING_PATH) / f'{name}-{rev}' 
+    exts = [f'.{ext}' for ext in settings.IDSUBMIT_FILE_TYPES]
     for ext in exts:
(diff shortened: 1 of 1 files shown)


ietf/submit/utils.py
```diff
@@ -671,20 +671,16 @@ def move_files_to_repository(submission):
...
-    """Remove staging files corresponding to a submission
-    
-    exts is a list of extensions to be removed. If None, defaults to settings.IDSUBMIT_FILE_TYPES.
-    """
-    if exts is None:
-        exts = [f'.{ext}' for ext in settings.IDSUBMIT_FILE_TYPES]
+def remove_staging_files(name, rev):
+    """Remove staging files corresponding to a submission"""
     basename = pathlib.Path(settings.IDSUBMIT_STAGING_PATH) / f'{name}-{rev}' 
+    exts = [f'.{ext}' for ext in settings.IDSUBMIT_FILE_TYPES]
```
rjsparks, line 677: This is getting lucky and will only work as long as the set of generated filetypes is a subset of the submitable file types.

rjsparks: Is there any reason not to just remove anything with the given stem?

jennifer-richards: Not certain of the history, but I think `IDSUBMIT_FILE_TYPES` is essentially "extensions that might be associated with a submission" at this point. We use it in a couple places where we move files around. (See `rename_submission_files` and `move_files_to_repository`, at least.)

I'm inclined to keep it this way and fix it all when we refactor submission artifact handling.

rjsparks: There'll be an issue to resolve when we're holding both xml2rfc and rfc2html ".html" files.

rjsparks (approved):

PR #8541, opened 2025-02-14
feat: offswitch

rjsparks:


jennifer-richards (approved):
I'm approving, but we may want to add the off switch to this:

https://github.com/ietf-tools/datatracker/blob/7ce4a3c0c181e543a1fd1fa622ddff5314d85e49/ietf/utils/storage.py#L18

That's used for meeting host logos and floorplan images - neither likely to be exercised often during the meeting and both secretariat access-only, so perhaps not an absolute must.

rjsparks:
> I'm approving, but we may want to add the off switch to this:
> 
> https://github.com/ietf-tools/datatracker/blob/7ce4a3c0c181e543a1fd1fa622ddff5314d85e49/ietf/utils/storage.py#L18

Easy enough to do

jennifer-richards (approved):

PR #8606, opened 2025-02-26
feat: run the docker container as dev

rjsparks:


jennifer-richards (dismissed):
LGTM. It increases startup time a bit with work that _could_ be shared between the app and celery containers, but that can (and should) wait.

NGPixel (approved):

PR #8621, opened 2025-03-01
fix: more submission date feedback; refactor xml2rfc log capture

jennifer-richards:
Main goal: log helpful feedback on more (hopefully now all!) paths where a `<date>` fails parsing or interpretation.

Throws in some refactoring to centralize capture of the xml2rfc log output. This is done with the `capture_xml2rfc_output` context manager. A known side effect of the refactoring in 3529f14ae548fbffc25ea565a2084d5a5980f394 is that `xml2rfc.log.write_out` and `xml2rfc.log.write_err` will be restored to their original streams after each call to `render_missing_formats()` instead of being left pointing at otherwise abandoned `StringIO` instances. I think that's a win, but wanted to call it out as a possibly non-obvious change.

### DIFF
--- a/ietf/submit/utils.py
+++ b/ietf/submit/utils.py
@@ -58,7 +58,7 @@ from ietf.utils.draft import PlaintextDraft
 from ietf.utils.mail import is_valid_email
 from ietf.utils.text import parse_unicode, normalize_text
 from ietf.utils.timezone import date_today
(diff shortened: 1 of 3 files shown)


kesara (approved):
LGTM.

rjsparks (approved):

PR #8630, opened 2025-03-03
chore: configure gunicorn secure_scheme_headers

jennifer-richards:
Instead of training Django to look at the request headers, this configures gunicorn only to respect `X-Forwarded-Proto`. It sets `wsgi.url_scheme`, which is used by Django's `request.is_secure()` to decide whether a forwarded request is actually secure.

Draft while I look back to make sure there wasn't some other corner where we needed the `SECURE_PROXY_SSL_HEADER` setting.


jennifer-richards:
As far as I can tell it'll be ok to make the change to the django config, but I don't think we should do that right before a meeting just in case. I'm going to close this and create distinct PRs instead.

PR #8634, opened 2025-03-03
chore: get django out of the forwarded header business

jennifer-richards:
This will decide whether a request is using https based on `wsgi.url_scheme`, as decided by gunicorn.

This reduces the number of places we have to touch to declare that `X-Forwarded-Proto` is authoritative vs alternative headers. It's set for us at Cloudflare and it is configured to be respected in our `gunicorn.conf.py`. 

When this goes to staging and production, a quick way to check that the configuration is correct is to retrieve `/api/openid/.well-known/openid-configuration/` and confirm that the various URLs are all `https://`. If they're `http://`, then something is misconfigured.

Creating as a draft PR until after IETF-122


Michealinc (commented):
Explict

rjsparks (approved):

Issue #8424, opened 2025-01-15
rephrase text in banner on Legacy RFCs


alicerusso:
### Description

Current:    `This RFC was published on the Legacy stream.`
Suggested:  `This RFC is labeled as "Legacy"; it was published before a formal source was recorded.`

Example: https://datatracker.ietf.org/doc/rfc872/
Scope: This banner appears on 2083 RFCs.
Rationale: Current text sounds odd because "Legacy" is the label used when there is not source information; an RFC cannot be "published on the Legacy stream".  "Legacy" is defined as 'The RFC was published before a formal source (e.g., working group or stream) existed or was recorded.'  on https://www.rfc-editor.org/source-definitions/. 

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

bkmgit:
Working on this

rjsparks:
fixed by #8670

PR #8670, opened 2025-03-15
fix: Clarify legacy description in RFC banner (#8424)

bkmgit:


bkmgit:
fixes #8670 by using the suggested text in the issue.

rjsparks:
cc: @rdanyliw

rjsparks (approved):
There's some repetition of the conditionals in the template that would be better handled as nested conditionals, but we'll improve it with a separate effort.

PR #8671, opened 2025-03-15
feat: add API for related emails

rpcross:
Add a API endpoint. Given one email address return all other related email addresses the Datatracker knows about


jennifer-richards (changes requested):
Rather than using querystring parameters to identify the email, I think it'd be better to use a URL parameter. E.g., `/api/person/email/joe@example.com/related/` instead of `/api/person/email/related?email=joe@example.com`.

We should also make sure that this can handle non-latin characters (e.g., `अजय@डाटा.भारत`, borrowing a Hindi example from https://en.wikipedia.org/wiki/International_email). This may require explicit use of `urllib.parse.quote` and `unquote` but I'm not certain. Django may do some magic there. The tests should exercise this.

(Datatracker itself may have issues with internationalized email addresses, not sure offhand, but we should make sure that this API will work with them if they exist)

Michealinc (commented):
Explict

jennifer-richards (changes requested):
Just a small change needed to cleanly reject null characters, see inline. Probably should add a test case for that.

jennifer-richards (approved):
Looks good. The tests on main will be fixed soon, so once #8804 is merged we can merge that here and let the tests run.

PR #8680, opened 2025-03-16
feat: improve apply_to_all handling in upload forms

peterthomassen:
- Relabel `apply_to_all` field because old text caused confusion ("what is being applied?")
- Set `apply_to_all=False` by default for uploading slides (as per the apparently common usecase)

### DIFF
--- a/ietf/meeting/forms.py
+++ b/ietf/meeting/forms.py
@@ -458,11 +458,10 @@ class UploadBlueSheetForm(FileUploadForm):
 class ApplyToAllFileUploadForm(FileUploadForm):
     """FileUploadField that adds an apply_to_all checkbox
 
-    Checkbox can be disabled by passing show_apply_to_all_checkbox=False to the constructor.
-    This entirely removes the field from the form.
+    Checkbox can be removed from form by passing show_apply_to_all_checkbox=False to the constructor.
     """
     # Note: subclasses must set doc_type for FileUploadForm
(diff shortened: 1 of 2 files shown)


rjsparks:
Hi Peter - thanks for this.
A question for future chair training - where did you run into the pain that led you to choose to contribute these changes? What can we do (beyond the changes you suggest here) to help others avoid that pain?

peterthomassen:
It was just that I (not a chair) was submitting slides, and asked our WG chair what that checkbox meant, because it wasn't clear to me. It was explained to me, together with the comment that they (the WG chairs) always untick that box during approval because slide decks usually go to only the session where they will be presented. So I thought it makes sense to set it to false initially.

rjsparks:
Please see https://mailarchive.ietf.org/arch/browse/wgchairs/?q=subject%3A(apply%20to%20all%20sessions)
I'll hold this PR for now, but it's likely that we'll close and not merge it, in favor of a refactor of the upload process instead that moves the checkbox to the approver rather than the submitter.

peterthomassen:
Fine with me. However,

- if I'm not mistaken the checkbox is already on the approve form (see `ApproveSlidesForm`), so the task would be merely to remove it from the upload form (happy to adjust this PR accordingly);
- the question remains open what the default value of this setting should be.

It seems to be that the default ideally should be a WG property, but my gut feeling is that that's unnecessarily complex.

Michealinc (commented):
Explict

rjsparks:
I'll shift this to draft as part of holding it aside.

PR #8681, opened 2025-03-16
fix: do not list GMT as a meeting time option (#8507)

### LINKED ISSUE #8507: Disallow "GMT" as a time zone choice
### Description

Our time zone list includes "GMT", at least for creating interim meetings (likely everywhere). While pedants might applaud the distinction, it should not be allowed. Only "UTC" should be allowed.

The list could be filtered in the form, but we should explore how we're building up the `ietf.meeting.models.timezones` in the top of `ietf/meeting/models.py`. If we can safely filter out "GMT" there, we should.

This is relevant to #5635 and perhaps #5683.

See also #8506, which adjusts our iCalendar renderer to handle "GMT", and #5393, which tracks refactoring how we generate iCalendar files.

### Code of Conduct

bkmgit:
Fixes https://github.com/ietf-tools/datatracker/issues/8507


jennifer-richards (changes requested):
This looks good, but we should have a migration to update meetings currently using `GMT` to use `UTC` instead. I think that just needs to run `Meeting.objects.filter(time_zone="GMT").update(time_zone="UTC")`.

Michealinc (commented):
Explict

rjsparks:
We'll add the migration - but it won't be in this week's release.

bkmgit:
Ok. Can do it next week, but please feel free to make the change if it needs to be sooner than that.

jennifer-richards (dismissed):
Added the migration

bkmgit:
@jennifer-richards thanks.

jennifer-richards:
@bkmgit You're welcome, and thank you for the PR!

rjsparks (approved):

PR #8685, opened 2025-03-17
fix(agenda): handle calendar view events that spread across multiple days

### LINKED ISSUE #8682: Calendar view display - local time
### Describe the issue

Hi - When selecting a set of WG sessions for Bangkok and trying to display them in calendar view AND my local time of US EDT, anything that begins before midnight (e.g. Dispatch) is invisible on the calendar, and I can't scroll down or up to see it.

### Code of Conduct

NGPixel:
Fixes #8682


rjsparks:
Would be nice if we had a set of tests that would track changes like this.

rjsparks (approved):

NGPixel:
We don't test the calendar component (other than opening the modal and validating the buttons around it) because it's an external library.

jennifer-richards (approved):

PR #8693, opened 2025-03-18
fix(agenda): Agenda 'now' fallback to next event if there is no current event

holloway:
The 'now' internal page link would fail if there was no current event. For example, clicking 'now' before  
events start for the day would fail. This PR adds a fallback where it will find the next event even if it hasn't started yet.

### DIFF
--- a/client/agenda/AgendaQuickAccess.vue
+++ b/client/agenda/AgendaQuickAccess.vue
@@ -204,10 +204,10 @@ function scrollToDay (daySlug, ev) {
 }
 
 function scrollToNow (ev) {
-  const lastEventId = agendaStore.findCurrentEventId()
+  const nowEventId = agendaStore.findNowEventId()
 
-  if (lastEventId) {
-    document.getElementById(`agenda-rowid-${lastEventId}`)?.scrollIntoView(true)
+  if (nowEventId) {
+    document.getElementById(`agenda-rowid-${nowEventId}`)?.scrollIntoView(true)
   } else {
     message.warning('There is no event happening right now.')
   }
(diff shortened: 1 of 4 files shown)


client/agenda/AgendaQuickAccess.vue
```diff
@@ -204,10 +204,10 @@ function scrollToDay (daySlug, ev) {
...
 function scrollToNow (ev) {
-  const lastEventId = agendaStore.findCurrentEventId()
+  const nowEventId = agendaStore.findNowEventId()
 
-  if (lastEventId) {
-    document.getElementById(`agenda-rowid-${lastEventId}`)?.scrollIntoView(true)
+  if (nowEventId) {
+    document.getElementById(`agenda-rowid-${nowEventId}`)?.scrollIntoView(true)
   } else {
     message.warning('There is no event happening right now.')
```
NGPixel, line 212: missing "or in the future" to be consistent with AgendaMobileBar

client/agenda/store.js
```diff
@@ -230,6 +230,38 @@ export const useAgendaStore = defineStore('agenda', {
 
       return lastEvent.id || null
     },
+    findNowEventId () {
+      const currentEventId = this.findCurrentEventId()
+      
+      if (currentEventId) return currentEventId
```
NGPixel, line 236: Add curly braces.

client/agenda/store.js
```diff
@@ -230,6 +230,38 @@ export const useAgendaStore = defineStore('agenda', {
...
+
+      const current = (this.nowDebugDiff ? DateTime.local().minus(this.nowDebugDiff) : DateTime.local()).setZone(this.timezone)
+
+      // -> Find next event after current time
+      let nextEvent = {}
+      for(const sh of this.scheduleAdjusted) {
+        if (sh.adjustedStart > current) {
+          // -> Use the first event of multiple events having identical times
+          if (nextEvent.start === sh.adjustedStart.toMillis()) {
+            continue
```
NGPixel, line 248: Use break instead? No need to keep iterating, as events are already chronological.

NGPixel (changes requested):

NGPixel (approved):

PR #8694, opened 2025-03-18
fix(agenda): strikethrough cancelled and rescheduled events

NGPixel:


rjsparks (dismissed):

rjsparks (dismissed):

client/agenda/AgendaScheduleList.vue
```diff
@@ -1087,11 +1087,13 @@ onBeforeUnmount(() => {
         .badge.is-cancelled {
           background-color: $red-500;
           text-transform: uppercase;
+          width: 100px;
```
holloway, line 1090: Minor suggestion but a `min-width` might handle large font sizes better, or using `em` units

NGPixel: I tried at 250% zoom and it scaled just fine with `width: 100px;`.

I don't think `min-width` would solve it as larger text would just keep overflowing. Do you have an example where it causes an issue?

holloway: I was more thinking about those increasing font size without zooming. 100px would increase in value when zoomed because it's a scalable CSS pixel so it typically wouldn't change layout.

As I said it's a minor suggestion. There probably isn't a issue, it's just a layout pattern that I find more flexible than fixed widths.

rjsparks (dismissed):

Michealinc (commented):
Explict

rjsparks (approved):

PR #8713, opened 2025-03-21
feat: preserve metadata when handling blobs; refactor storage_utils

jennifer-richards:


jennifer-richards:
Work to reconcile this with #8784 is in progress - merging this as it is to try to avoid future merge nightmares. Fixups may still be necessary and will be treated as future work.

PR #8738, opened 2025-03-26
feat: highlight unexpected state in AD dashboard

### LINKED ISSUE #8550: Highlight unexpected WG state for documents in Publication Requested
### Describe the issue

See https://datatracker.ietf.org/doc/draft-ietf-dnsop-rfc8624-bis/

As the responsible AD, I looked at the IESG state that was "Publication Requested" w/o looking at the WG status (my bad), which was reverted back to "Waiting for WG Chair Go-Ahead" *after* publication was requested but leaving the IESG state unchanged.

Not a big deal but it would be nice to remove the "Publication Requested" state from the IESG status.

Image

### Code of Conduct

rudimatz:
Fixes #8550

### DIFF
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -490,10 +490,7 @@ def is_unexpected_wg_state(doc):
     draft_iesg_state = doc.get_state("draft-iesg")
     draft_stream_state = doc.get_state("draft-stream-ietf")
 
-    return not (
-        (draft_iesg_state and draft_iesg_state.slug == "idexists")
-        or (draft_stream_state and draft_stream_state.slug == "sub-pub")
-    )
+    return draft_iesg_state.slug != "idexists" and draft_stream_state is not None and draft_stream_state.slug != "sub-pub"
 
 
 @register.filter
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -482,20 +482,17 @@ def state(doc, slug):
 
 
 @register.filter
-def inconsistent_state(doc):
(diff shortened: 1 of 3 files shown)


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -482,20 +482,17 @@ def state(doc, slug):
...
-    draft_state = doc.get_state("draft")
 
-    return (
-        not (draft_iesg_state and draft_iesg_state.slug == "idexists")
-        and not (draft_stream_state and draft_stream_state.slug == "sub-pub")
-        and not (draft_stream_state and draft_stream_state.slug == "dead")
-        and not (draft_state and draft_state.slug == "expired")
+    return not (
+        (draft_iesg_state and draft_iesg_state.slug == "idexists")
+        or (draft_stream_state and draft_stream_state.slug == "sub-pub")
```
rjsparks, line 495: This is still a bit of a logic puzzle for future readers. I suggest this as a clearer way to state it:
```
    return draft_iesg_state.slug != "idexists" and draft_stream_state is not None and draft_stream_state.slug != "sub-pub"
```
+/- black formatting

ietf/doc/templatetags/tests_ietf_filters.py
```diff
@@ -14,12 +14,15 @@
     ConflictReviewFactory,
     BofreqFactory,
     StatementFactory,
+    DocumentFactory,
```
rjsparks, line 17: I think you should use WgDraftFactory rather than the bare DocumentFactory - it doesn't make sense to test this against non-ietf-stream documents.

rjsparks (changes requested):
Looking good - a couple of minor changes suggested

rjsparks (approved):

PR #8761, opened 2025-03-31
feat: block iesg ballot for docs from outside streams

### LINKED ISSUE #8072: Make it harder (impossible perhaps) to issue an iesg ballot on a document not in the iesg stream
We've had conflict review documents for many years now to handle the IESG balloting on conflict reviews of documents in other streams.

However, the older ability to open a ballot on any document still exists, and is a trap for new ADs.

Change the permissions such that new ADs can't make the mistake of balloting directly on other stream's documents.

rudimatz:
Fixes https://github.com/ietf-tools/datatracker/issues/8072

### DIFF
--- a/ietf/doc/tests_ballot.py
+++ b/ietf/doc/tests_ballot.py
@@ -17,7 +17,7 @@ from django.utils import timezone
 from ietf.doc.models import (Document, State, DocEvent,
                              BallotPositionDocEvent, LastCallDocEvent, WriteupDocEvent, TelechatDocEvent)
 from ietf.doc.factories import (DocumentFactory, IndividualDraftFactory, IndividualRfcFactory, WgDraftFactory,
-                                BallotPositionDocEventFactory, BallotDocEventFactory, IRSGBallotDocEventFactory)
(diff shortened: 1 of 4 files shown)


ietf/doc/tests_ballot.py
```diff
@@ -360,7 +360,7 @@ def test_request_last_call(self):
         self.assertTrue('aread@' in outbox[-1]['Cc'])
 
     def test_edit_ballot_writeup(self):
-        draft = IndividualDraftFactory(states=[('draft','active'),('draft-iesg','iesg-eva')])
+        draft = IndividualDraftFactory(states=[('draft','active'),('draft-iesg','iesg-eva')], stream=StreamName.objects.get(slug='ietf'))
```
rjsparks, line 363: I think using `stream_id='ietf'` should work here and would avoid the `StreamName` import and db lookup.

rjsparks (changes requested):
Minor simplifying changes requested.

jennifer-richards (approved):

rjsparks (approved):

PR #8762, opened 2025-03-31
chore(deps): unpin importlib-metadata

jennifer-richards:
Both packages that depend on `importlib-metadata` now explicitly call out 8.4.0, so our pin is no longer helpful.


rjsparks (approved):

QueTee82 (commented):
C  hm kl ok d

PR #8779, opened 2025-04-03
fix: remove credentials omit from agenda fetch call

NGPixel:


jennifer-richards (commented):
Wonder if it should be `'same-origin'` to reduce risk of CSRF issues if the code gets reused?

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#including_credentials

edit: never mind, I see that's the default if I actually read my own reference 😬

jennifer-richards (approved):

rjsparks:
From discussion - instead of this PR we probably want to either change the code to make production and staging behave differently, or change the CF rules around the api endpoint at staging instead.

If we want to do that in this code, please update this PR, otherwise close it.

rjsparks:
Subsequent discussion convinced me I was incorrect above.

PR #8800, opened 2025-04-14
style: Apply Black styling to the entire Python codebase

jennifer-richards:
This appears to work for me with a couple notes:
1. If we do a squash merge, we'll need to change the commit hash in `.git-blame-ignore-revs`.
2. Requires per-user configuration by running `git config --local blame.ignoreRevsFile .git-blame-ignore-revs`. Maybe we can automate this as part of checkout/config, but I haven't found an automatic way to use this file at the git level.
3. I had to invalidate caches in my IDE to get it to register the git config change, but this is probably tool-specific

Also, very important: we'll need to modify our actions to accommodate the change from single quotes to double quotes in `ietf/__init__.py` (see the comments there)


richsalz:
Is there documentation on what the default `black` styling is? We should add a pointer in the README to it.

If we do this, we should add a GH action that checks for `black` styling and either auto-fixes or blocks the PR. And consider telling people how to run `black` in their own workspace to avoid such PR blockage.

jennifer-richards:
> At some point we should seriously think about the #7725 and #7690 PR's.

Good point - what we sort out here as far as keeping the git history/blame clean will apply to those as well.

[Black's rules are linked from here](https://github.com/psf/black?tab=readme-ov-file#the-black-code-style)

richsalz:
I think you should add a reference to the style in `README.md`

jennifer-richards:
Enough changes are going to be needed both to method and managing change attribution that I'm going to close this PR to reduce noise. Thanks for the comments, @richsalz, they have been captured.

Issue #7929, opened 2024-09-12
Move datatracker off of python 3.9


rjsparks:
Move to 3.12 if possible. 
Prepare for moving forward following  https://devguide.python.org/versions/ in mind.

jennifer-richards:
Depending on timing and where we are with Django, it might be possible to jump to 3.13. That'll require Django 5.1, though, so we shouldn't hold this up for that.

microamp:
3.13 has been released. 

https://code.djangoproject.com/ticket/34900 (a PR has been created, currently pending)

jennifer-richards:
I did a quick exploration of building a dev image against Python 3.12. The image built and seemed to resolve its dependencies without adjustments, which is encouraging. As noted below, it could not actually start up Django, though, so there may be demons waiting in the shadows.

The immediate blocker was the removal of the (long-)deprecated `smtpd` library in 3.12. ("Why did it have to be email?!" @rjsparks asked rhetorically; reports that he was in full Indiana Jones regalia are unconfirmed 🐍 ) This is only used in our dev/test mode SMTP server as far as I can tell. The fix is to [migrate from smtpd to aiosmtpd](https://aiosmtpd.aio-libs.org/en/latest/migrating.html), which is likely fairly straightforward.

PR #8805, opened 2025-04-16
refactor: smtpd->aiosmtpd

### LINKED ISSUE #8077: Refactor test_smtpserver.py using aiosmtpd
### Describe the issue

For testing / dev environments, an test SMTP server is used to handle messages the Datatracker sends. This uses the `smtpd` library which was deprecated in Python 3.6 and dropped in Python 3.12. This is blocking #7929.

General instructions for migrating from amtpd to aiosmtpd are here: https://aiosmtpd.aio-libs.org/en/latest/migrating.html

### Code of Conduct

jennifer-richards:
This is a prerequisite for moving to Python 3.12 (see https://github.com/ietf-tools/datatracker/issues/7929#issuecomment-2435388004). Fixes #8077 

The `line_length_limit` increase in commit 5bb3700 is needed because, as indicated in the comment, some of the tests are sending messages that exceed the RFC 5321 line length limit in the DATA section. From what I have found so far, the issue is that datatracker is incorrectly sending `<LF>` instead of `<CRLF>` as line endings when sending SMTP messages. It's just possible that this is a test-only bug, but it's likely affecting production. This is causing failures now because aiosmtpd is strict, both about line endings and about line lengths, whereas our previous test server was extremely lenient.

While that's embarrassing and should be investigated/fixed soon, I think we should do that separately and run with the `line_length_limit` fix for now. That only affects testing code and can only cause new failures, not hide ones we'd have caught before.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 # -*- conf-mode -*-
 setuptools>=51.1.0    # Require this first, to prevent later errors
 #
+aiosmtpd>=1.4.6
 argon2-cffi>=21.3.0    # For the Argon2 password hasher option
 beautifulsoup4>=4.11.1    # Only used in tests
 bibtexparser>=1.2.0    # Only used in tests
--- a/ietf/utils/management/tests.py
+++ b/ietf/utils/management/tests.py
@@ -12,7 +12,7 @@ from ietf.utils.management.base import EmailOnFailureCommand
 from ietf.utils.test_utils import TestCase
 
 
-@mock.patch.object(EmailOnFailureCommand, 'handle')
+@mock.patch.object(EmailOnFailureCommand, 'handle', return_value=None)
 class EmailOnFailureCommandTests(TestCase):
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 # -*- conf-mode -*-
 setuptools>=51.1.0    # Require this first, to prevent later errors
 #
+aiosmtpd>=1.4.6
```
kesara, line 4: IMO moving forward, we should try to separate test only dependencies. Not saying we should start with this PR, but better to start somewhere?

kesara: This will help us to have clear SBOM for production server.

jennifer-richards: I like the idea, though have some concerns about how we would validate changes to the production dependency list prior to deployment. 

I don't think it's practical to start in this PR because there will be quite a few tooling and deployment adjustments needed regardless of how we do it. It'll need to be its own project.

rjsparks (approved):

PR #8811, opened 2025-04-17
feat: Python 3.12

### LINKED ISSUE #7929: Move datatracker off of python 3.9

jennifer-richards:
Updates Python to 3.12. Fixes #7929.

Moves to the current version of coverage and refactors the test harness to work with this. This does not have a significant impact on speed.

Switches from using the `:latest` tag on app base to `:py312`. This will continue to work as intended after merging to main. If we want to revert to `:latest` or introduce a different tagging scheme, we can do that post-merge.

The py312 compatibility of some dependencies was ambiguous, but most of these are generically "python 3" or "python 3.7+" compatible. There's a small risk that something will break, but our tests all pass and we have not seen any signs of issues.


rjsparks (approved):

peejayquijano60-crypto (commented):
Realese

PR #8826, opened 2025-04-23
chore: install ruff formatter

jennifer-richards:
Makes the ruff formatter available


rjsparks:
See https://github.com/ietf-tools/datatracker/blob/main/docker/scripts/app-setup-python.sh#L21 and https://github.com/ietf-tools/datatracker/blob/main/.devcontainer/devcontainer.json#L44

jennifer-richards:
Yeah, probably better to do it that way. It does not have a lot of dependencies, so it's low risk for test vs prod environment issues.

PR #8828, opened 2025-04-23
feat: require strong passwords

jennifer-richards:
Adds a password validator based on [zxcvbn-python](https://github.com/dwolfhub/zxcvbn-python). Rejects passwords with strength < 3. Also rejects passwords shorter than 12 characters.

Refactors change-password and login forms to use more of Django's password validation methods.

For the login view, validation failure causes the user to be instructed to reset their password.

For the password change view, validation failure rejects the attempt to update the password. The JS is updated to do the same validation on the client side to prevent submission of the form so the user can correct mistakes before submitting and losing their form state.

image


jennifer-richards:
Removed the attempt at requiring stale accounts to reset their passwords. We don't store a password age so there's no clean mechanism for indicating that a user has gone through the password reset. We can revisit that separately.

richsalz:
We don't store a "last-login" in the user record?  Perhaps we should.

jennifer-richards:
> We don't store a "last-login" in the user record? Perhaps we should.

We do; the reverted commits used that to decide whether to let someone log in or not. The problem is that after they reset their password, they still haven't logged in, so they're again prompted to reset their password.

What's missing is a way to flag "hasn't logged in but already updated their password." It's a tractable problem, but bigger than would be reasonable to toss in to this PR.

richsalz:
That makes sense, thanks for the explanation.

rjsparks (dismissed):

rjsparks (approved):

PR #8830, opened 2025-04-23
fix: put draft name first in subject of review email

### LINKED ISSUE #5946: Subject for last-call review e-mails isn't friendly to small screens
### Describe the issue

The `Subject:` line sent for directorate reviews typically starts wtih:

`[Last-Call] Foodir last call review of draft...`

There is very little information in these characters. While that isn't an issue for someone processing them by script or viewing them on a very large screen, it makes the last-call list difficult to use on a small screen (which is increasingly prevalent). E.g., see screenshot below.

Assuming that `[Last-Call]` is going to stay, I'd suggest something like:

`[Last-Call] draft... Foodir review`

and, perhaps, even omitting `draft-ietf-` since that's always present.

image001

### Code of Conduct

mnot:
Fixes #5946.

This is the minimal fix; other optimisations might be possible, but this gets the job done for the common case -- people scanning for drafts they're interested in.

### DIFF
--- a/ietf/doc/views_review.py
+++ b/ietf/doc/views_review.py
@@ -897,7 +897,7 @@ def complete_review(request, name, assignment_id=None, acronym=None):
 
             if need_to_email_review:
                 # email the review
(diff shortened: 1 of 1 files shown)


rjsparks (approved):

rjsparks:
I suspect we'll get feedback from the review community when this is deployed, but I don't think there's a better way to _get_ that feedback.

PR #8838, opened 2025-04-24
feat: more detailed view of all docs in iesg processing

### LINKED ISSUE #7307: Unified IESG Document Dashboard
### Description

A given AD's dashboard (e.g., [Roman's](https://datatracker.ietf.org/doc/ad/roman.danyliw)) shows the documents they are processing per state under the "Documents for <insert AD name>" section.  To get a sense of what documents are being processed by the entire IESG, this enhancement would create a new dashboard notionally described as "Documents for the IESG".  

The layout out would be identical to the "Documents for <insert AD name>" by enumerating different document states and the associated documents in that state.  However, the documents listed for each state would be from the entire IESG.

As an aside, individual AD dashboards also include an AD's DISCUSS position.  An view of all of the IESG's DISCUSSes is already available at https://datatracker.ietf.org/iesg/discusses/.

### Code of Conduct

rjsparks:
fixes #7307

### DIFF
--- a/ietf/doc/tests.py
+++ b/ietf/doc/tests.py
@@ -403,6 +403,30 @@ class SearchTests(TestCase):
         self.assertContains(r, discuss_other.doc.name)
         self.assertContains(r, block_other.doc.name)
 
+    def test_docs_for_iesg(self):
+        ad1 = RoleFactory(name_id='ad',group__type_id='area',group__state_id='active').person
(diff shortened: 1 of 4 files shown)


jennifer-richards (changes requested):
You'd mentioned this is fairly slow - I think even a few minute cache on the calculations in the view would be worth adding so people don't hammer the server browsing back and forth

jennifer-richards (approved):

PR #8845, opened 2025-04-27
fix: make submission text less passive

### LINKED ISSUE #8807: https://datatracker.ietf.org/submit/ -- passive title
### Description

In menus and on the page, "Internet-Draft Submission" is very passive and non-obvious. 

I suggest "Submit an Internet-Draft" -- it's easier to find in menus.

### Code of Conduct

mnot:
Fixes #8807.

### DIFF
--- a/ietf/templates/base/menu.html
+++ b/ietf/templates/base/menu.html
@@ -95,7 +95,7 @@
     <li>
         <a class="dropdown-item {% if flavor != 'top' %}text-wrap{% endif %}"
            href="{% url "ietf.submit.views.upload_submission" %}">
-            I-D submission
+            Submit an Internet-Draft
         </a>
     </li>
     <li>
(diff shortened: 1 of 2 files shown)


rjsparks:
One minor concern - this page is not only for submitting drafts - it's for checking on the status of existing submissions, approving/canceling prior submissions and other activities.

I don't object to making the entry point a little more attractive, but there is the chance this will cause people wanting to exercise some of these other workflows friction as they are not wanting, necessarily to perform the action the new menu item says it's going to perform.

rjsparks:
btw Mark - thanks for these submissions - please continue.
It'll save me a step if you title the PR (or better, make your commits, then the PR title will happen automatically) using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

mnot:
Will do.

rjsparks (approved):

Issue #4514, opened 2022-09-27
The IAB chair and IETF chair should be able to send liaison statements with their role address in the From


rjsparks:
Right now, Mirja can't send a liaison statement from iab-chair@iab.org. The liaison tool will only let her choose ietf@kuehlewind.net. In Feb 2021, to work around this, a "IAB Chair" user was created in the datatracker and given the ability to send using the liaison tool. 

This was not the best thing to have happened, and needs to be reworked.

Instead, the liaison statement tool should recognize role holders of particular roles, specifically the IAB or IETF chair, and give them the option to send from the chair alias rather than one of their personal email addresses.

cc: @mirjak, @larseggert

larseggert:
Yes, please.

mirjak:
Also yes, please!

rjsparks:
As of https://github.com/jennifer-richards/datatracker/blob/58f8bbd9444b4efbec5e69d09dadd1921e2d03ff/ietf/liaisons/forms.py#L540-L564, the Secretariat, the IAB Chair, the IETF Chair, and the Liaison Coordinators can set the from contact to any well-formed email address. I'm going to close this issue. https://github.com/ietf-tools/liaison-tooling-2025a/issues/4 and related will make sure we preserve those permissions.

I think there's still tension, though, with what to do with this field for people with other roles (see https://github.com/ietf-tools/liaison-tooling-2025a/issues/28 in particular).

PR #8883, opened 2025-05-05
feat: adjust liaison statement to/from group choices based on role 

jennifer-richards:
This refactors some internals of the liaison statement (LS) forms to improve testability and clarity. Adjusts the options available for "To" and "From" groups based on the user's role(s). 

Note that these changes only impact the _groups_ that can be used to populate the "From > Groups" and "To > Groups" fields. The "From > From contact", "From > Response Contacts", and "To > Contacts" fields and their behavior are unchanged. Furthermore, the "From" header field on the actual email that delivers the LS is not affected by this.

In the following, "any group" means IETF, IESG, IAB, any active area, or any active WG.

Roles permitted to add new **outgoing** LS:
* Area Director
  - _From_ their area, an WG in their area, the IETF, or the IESG
  - _To_ any SDO group
* IAB Chair
  - _From_ any group **(Was previously only the IAB)**
  - _To_ any SDO group
* IAB Executive Director **(No current such role exists in the datatracker)**
  - _From_ any group **(Was previously IAB only)**
  - _To_ any SDO group
* IETF Chair
  - _From_ any group **(Was previously only the IETF)**
  - _To_ any SDO group
* Liaison Manager (of an SDO group)
  - _From_ any group
  - _To_ only their SDO group
* Secretariat
  - _From_ any group
  - _To_ any SDO group
* WG Chair
  - _From_ only their WG
  - _To_ any SDO group
* WG Secretary
  - _From_ only their WG
  - _To_ any SDO group

Roles permitted to add new **incoming** LS (these were not changed):
* Authorized Individual (of an SDO group)
  - _From_ only their SDO group
  - _To_ any group
* Liaison Manager (of an SDO group)
  - _From_ only their SDO group
  - _To_ any group
* Secretariat
  - _From_ any SDO group
  - _To_ any group

Still a WIP to confirm the changes are in the right direction and for some further code cleanup.

Related to ietf-tools/liaison-tooling-2025a#4 and #4514


mirjak:
Actually not sure we need a separate IAB Executive Director. The IAB Executive Director is a member of the Secretariat, so the role should be sufficient. This is also related to this discussion: https://github.com/ietf-tools/liaison-tooling-2025a/discussions/7

Also I think we actually want to get rid of the Authorized Individual but we need to fully double-check that we don't use it anymore but that's the outcome of https://github.com/ietf-tools/liaison-tooling-2025a/discussions/25 and https://github.com/ietf-tools/liaison-tooling-2025a/discussions/11 

I think I asked that Robert at some point and he might have looked into this already but I can find it right now: Is there a list of all currently Authorized Individuals somewhere, so we could quickly review it?

jennifer-richards:
Ok - I won't worry about the permissions for those roles then (i.e., I'll leave them as they are in this PR).

> I think I asked that Robert at some point and he might have looked into this already but I can find it right now: Is there a list of all currently Authorized Individuals somewhere, so we could quickly review it?

The easiest way to see these is through the Django admin in the sandbox. I can give you instructions for that offline. There are 137 Authorized Individuals for SDO groups.

rjsparks (dismissed):

PR #8896, opened 2025-05-13
chore(deps): bump to fixed django-celery-beat

jennifer-richards:
django-celery-beat 2.8.1 is out with the fix for the time zone issue. We could fall back to >= 2.3.0 because they've yanked the broken release, but I don't see any reason to allow the old versions.

Should be safe to install this immediately since they're all currently in UTC. We can later test that schedules in non-UTC time zones do in fact run, then convert our schedules back to America/Los_Angeles (if we want to do so).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ Django>4.2,<5
 django-admin-rangefilter>=0.13.2
 django-analytical>=3.1.0
 django-bootstrap5>=21.3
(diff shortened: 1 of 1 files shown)


rjsparks (approved):

rjsparks:
Lets take this opportunity to make the move to doing things in UTC stick.

PR #8901, opened 2025-05-14
feat: rfc authors

rjsparks:


ietf/doc/admin.py
```diff
@@ -224,3 +224,9 @@ class StoredObjectAdmin(admin.ModelAdmin):
     list_filter = ['deleted']
     search_fields = ['store', 'name', 'doc_name', 'doc_rev', 'deleted']
 admin.site.register(StoredObject, StoredObjectAdmin)
+
+class RfcAuthorAdmin(admin.ModelAdmin):
```
jennifer-richards, line 228: Should show titlepage_name on the list and definitely be able to search with it (person is nullable, so rows may have no indication of a name in some cases as it is here)

PR #8902, opened 2025-05-14
refactor: use new meeting.Registration model

rpcross:
Migrate everything that uses stats.MeetingRegistration to meeting.Registration


rjsparks (dismissed):

ietf/meeting/tests_views.py
```diff
@@ -9271,27 +9269,23 @@ def test_get_next_sequence(self):
...
+        AttendedFactory(session__meeting=m, session__type_id='plenary', person=creg.person)
         checked_in, attended = participants_for_meeting(m)
-        self.assertTrue(person_a.pk in checked_in)
-        self.assertTrue(person_b.pk not in checked_in)
-        self.assertTrue(person_c.pk in attended)
-        self.assertTrue(person_d.pk not in attended)
+        self.assertTrue(areg.person.pk in checked_in)
+        self.assertTrue(breg.person.pk not in checked_in)
+        self.assertTrue(creg.person.pk in attended)
+        self.assertTrue(dreg.person.pk not in attended)
```
jennifer-richards, line 9282: Worth also checking areg and breg with respect to `attended`?

Also, minor style, but `self.assertIn()` and `self.assertNotIn()` often give more useful error messages on a test failure

jennifer-richards (dismissed):
Looks fine to me. Curious about the `PROTECT` change - why it was added and removed. It seems potentially useful as a guard against deleting a Person who still has data tied to them.

ietf/meeting/models.py
```diff
@@ -1513,6 +1513,13 @@ class Registration(models.Model):
     def __str__(self):
         return "{} {}".format(self.first_name, self.last_name)
 
+    @property
+    def attendance_type(self):
```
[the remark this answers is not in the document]
rpcross: In practice this should not return None, but added explicit return for clarity. In commit ef7fe93120316766aa5fd6f0911fed21aa210e16

ietf/meeting/tests_views.py
rpcross: Addressed in commit ef7fe93120316766aa5fd6f0911fed21aa210e16

rpcross:
> Looks fine to me. Curious about the `PROTECT` change - why it was added and removed. It seems potentially useful as a guard against deleting a Person who still has data tied to them.

The final state is on_delete=PROTECT. I had briefly changed to SET_NULL but reverted that change.

jennifer-richards:
> The final state is on_delete=PROTECT. I had briefly changed to SET_NULL but reverted that change.

Ah, I must have confused myself by looking at individual commits.

jennifer-richards (commented):
Is there anything in the models enforcing `attendance_type` to be not-None? Perhaps it should raise an exception instead of returning None in the "no tickets" case. Unless places that use the property check for None, it's likely to cause a strange exception, so it might be better to raise one that points at "Registration has no tickets" or some other helpful message.

rpcross:
> Is there anything in the models enforcing `attendance_type` to be not-None? Perhaps it should raise an exception instead of returning None in the "no tickets" case. Unless places that use the property check for None, it's likely to cause a strange exception, so it might be better to raise one that points at "Registration has no tickets" or some other helpful message.

There is not a restriction in the models but there is in the business logic. Any new registration has a registration type (attendance type). If after a registration cancellation, there are no tickets, the Registration object is deleted. There is an edge case, you could have a Registration with only an Onsite Hackathon ticket for example and this property would return None. It's only use right now is in the proceedings view which already filters on 'onsite', 'remote' types. So it's very speculative, if a new use is added that encounters this edge should it print "None" or fail with an exception? Whichever is preferred.

jennifer-richards:
> So it's very speculative, if a new use is added that encounters this edge should it print "None" or fail with an exception? Whichever is preferred.

Sounds like `None` is fine. Thanks!

jennifer-richards (approved):

PR #8941, opened 2025-05-29
chore: Add RPC references API call to OpenAPI spec

kesara:


rpcapi.yaml
```diff
@@ -186,6 +185,26 @@ paths:
                 '404':
                     description: Not found
 
+    /doc/drafts/{docId}/references/:
+        get:
+            operationId: get_draft_references
+            summary: Get a normative references
+            description: Returns the normative for the requested ID
```
rjsparks, line 192:
```suggestion
            summary: Get normative references to Internet-Drafts
            description: Returns the id and name of each normatively referenced Internet-Draft for the given docId
```

jennifer-richards (approved):
Verified that it builds a working API client endpoint.

I think the `operationId` suggests a more general list of references than this provides. I can't come up with something I like better though. Similarly, and this should have been a comment on the original PR, but the URL  (`/doc/drafts/{docId}/references/`) feels more general.

Those can be adjusted easiiy later, though, so I'm going to merge it.

PR #8944, opened 2025-05-30
feat: liaison coordinators

rjsparks:
Addresses https://github.com/ietf-tools/liaison-tooling-2025a/issues/30


jennifer-richards (commented):
Looks ok, one question: should a Liaison Coordinator be included in the `False` set for `IncominingLiaisonStatementForm.get_post_only()`? That is currently `False` for secretariat and authorized individuals.

jennifer-richards (approved):

PR #8958, opened 2025-06-04
feat: more flexible liaison stmt "From Contact"

jennifer-richards:
Fixes https://github.com/ietf-tools/liaison-tooling-2025a/issues/4 and #4514 

Please review in particular the defaults for the from field in a8fcdc79b2bad44c2e655d2dda9c91bdd3716f48. I'm especially uncertain about the default for the Liaison Coordinator - maybe it should be the IAB instead of the chair?

FWIW, the migration takes just a few seconds, so no special care is needed. It's not reversible - we could do the reverse lookup, but as soon as someone actually uses an address like chair@ietf.org, there's no going back to a FK to Email, so I don't see a point.


jennifer-richards:
~~Converting to draft until I update the Resources to match the change~~

Ready to go

rjsparks (approved):

rjsparks:
One thought - IIRC we got to where we _had_ been because people typod a from address and got upset later that the tool didn't save them from the embarrassment of the mistake. Also while we were pointing at Email, the sending code could note that the address was no longer active and use a better one. These properties are lost, but I believe that sacrifice is what the IAB is asking for at the moment.

PR #8965, opened 2025-06-05
test: Add test for api.ietf_utils

kesara:


ietf/api/tests_ietf_utils.py
```diff
@@ -0,0 +1,46 @@
+# Copyright The IETF Trust 2025, All Rights Reserved
+
+from django.test import TestCase, RequestFactory
```
jennifer-richards: We have a custom `TestCase` class - import it as `from ietf.utils.test_utils` instead of `django.test`

ietf/api/tests_ietf_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+from ietf.api.ietf_utils import is_valid_token, requires_api_token
+
+
+class IetfUtilsTests(TestCase):
+    @override_settings(APP_API_TOKENS={"ietf.api.foobar": ["valid-token"]})
+    def test_is_valid_token(self):
+        self.assertFalse(is_valid_token("ietf.fake.endpoint", "valid-token"))
+        self.assertFalse(is_valid_token("ietf.api.foobar", "invalid-token"))
+        self.assertTrue(is_valid_token("ietf.api.foobar", "valid-token"))
```
jennifer-richards, line 21: Could you add a case that covers the case of a token like:
```python
{"ietf.api.dont_do_this": "my-token"}
```
That's a misconfiguration, as the values in the config dict are meant to be lists, but it's a common error and it'd be good to test the code that protects against it.

(The code protects it by turning a bare string into a list, but without that, any single letter in the token would be accepted for access, which would be _bad_)

ietf/api/tests_ietf_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+        result = protected_function(request)
+        self.assertEqual(result.status_code, 403)
+
+        # request with a valid token for another API endpoint
+        request = RequestFactory().get(
+            "/some/url", headers={"X_API_KEY": "another-token"}
+        )
+        result = protected_function(request)
+        self.assertEqual(result.status_code, 403)
```
jennifer-richards, line 69: Maybe test the 'no token' case?

jennifer-richards (changes requested):
Thank you for doing this! There are a couple other cases that would be good to exercise, plus an import to adjust.

ietf/api/tests_ietf_utils.py
kesara: Fixed with d89138baba5ad29a87718e66e2bf787dbe9f449a

ietf/api/tests_ietf_utils.py
kesara: Add that case with 9895c8faf20ea6c4c69783760eca82a41645ecca

ietf/api/tests_ietf_utils.py
kesara: Looks code doesn't protect against that misconfiguration: 6e5868d26a7eb52dea17051752eaef811415f82d

kesara: 6e5868d26a7eb52dea17051752eaef811415f82d and b9cbbe7f7c02320917063bca6dcc5bdb719b1254 should address this issue.

ietf/api/tests_ietf_utils.py
```diff
@@ -31,16 +44,38 @@ def protected_function(request):
...
-        # request with an valid token
+        # request with an invalid token
         request = RequestFactory().get(
             "/some/url", headers={"X_API_KEY": "invalid-token"}
         )
         result = protected_function(request)
         self.assertEqual(result.status_code, 403)
 
+        # request without a token
+        request = RequestFactory().get("/some/url", headers={"X_API_KEY": ""})
```
jennifer-richards, line 55: This tests a blank header, but I think we need the no-header-at-all case.

jennifer-richards (changes requested):
One further change request and then I'm happy :-)

ietf/api/tests_ietf_utils.py
kesara: 4e80ea1f900bd44a14493a7272449acf0e73b31a should address this.

jennifer-richards (approved):

PR #9062, opened 2025-06-26
feat: person search endpoint

jennifer-richards:
This is the datatracker side of https://github.com/ietf-tools/purple/pull/365. Adds a paginated search endpoint for Persons.

I'm conflicted on naming. The names here (`RpcPerson*`) follow the pattern we've been using with `rpc` prefixes, but have the decidedly awful result that the purple view works with something called an `RpcPerson` that is not purple's `rpc.models.RpcPerson`. Suggestions welcome.

Note that this is starting to use DRF for the datatracker's purple interface. We have so far only used that for things supporting red. The motivation here is not having to implement custom pagination. The schema changes to `rpcapi.yaml` were done by generating a schema via spectacular and manually pulling in the relevant bits. We will want to make that process better, but I think that's a separate project.


ietf/api/views_rpc.py
```diff
@@ -63,6 +66,35 @@ def rpc_persons(request):
     return JsonResponse(response)
 
 
+class RpcLimitOffsetPagination(LimitOffsetPagination):
+    default_limit = 10
+    max_limit = 100
+
+
+class RpcPersonSerializer(serializers.ModelSerializer):
```
kesara: NIT: This should be in the `api/serializers.py` or `api/serializers_rpc.py`.

rpcapi.yaml
```diff
@@ -457,6 +487,43 @@ components:
...
+                    type: string
+                    nullable: true
+                    format: uri
+                    example: https://api.example.org/accounts/?offset=200&limit=100
+                results:
+                    type: array
+                    items:
+                        $ref: '#/components/schemas/RpcPerson'
+
+        RpcPerson:
```
kesara: You can avoid defining a new schema by using `Person` model.

ietf/person/models.py
```diff
@@ -88,7 +88,7 @@ def short(self):
         else:
             prefix, first, middle, last, suffix = self.ascii_parts()
             return (first and first[0]+"." or "")+(middle or "")+" "+last+(suffix and " "+suffix or "")
-    def plain_name(self):
+    def plain_name(self) -> str:
```
kesara, line 91: 🙀

ietf/api/urls_rpc.py
```diff
@@ -16,6 +16,7 @@
     url(r"^person/persons_by_email/$", views_rpc.persons_by_email),
     url(r"^person/(?P<person_id>[0-9]+)/$", views_rpc.rpc_person),
     url(r"^persons/$", views_rpc.rpc_persons),
+    url(r"^persons/search/", views_rpc.RpcPersonSearch.as_view()),
```
kesara, line 19: General question: Isn't `Rpc` and `rpc_` are redundent because these already in `views_rpc`?

kesara (changes requested):

ietf/person/models.py
jennifer-richards: lol, drf-spectacular uses this to generate the schema... I guess I could annotate it on the serializer, but that's so crude

ietf/api/urls_rpc.py
jennifer-richards: I think so, but I didn't want to violate the convention we've been using so far in this PR. My inclination is to do a follow-up PR that refactors a bunch of this, including getting rid of the hand-crufted `rpcapi.yaml`.

These names aren't exposed through the API schema so the impacts both of this being a bad name and of changing the names in a later PR shouldn't impact the purple API client. We should be able to match the existing schema with a generated one, at least pretty closely.

jennifer-richards: (Having the prefix in the method/class names is nice if you want to do a `from views_rpc import rpc_whatever` but I don't think we're likely to do it that way.)

rpcapi.yaml
jennifer-richards: Did it this way to reduce mixing generated and hand-written schema parts, but the schemas are compatible so you're right, we should be optimistic. :-) This also avoids leaking the `RpcPerson` serializer name into the API.

ietf/api/urls_rpc.py
kesara: (not relevant to this PR), should `persons_by_email` should be `rpc_persons_by_email`?
What I struggle with is difference between rpc prefixed person and non-rpc prefixed person in `views_rpc`.

kesara (approved):

ietf/api/urls_rpc.py
jennifer-richards: I guess we weren't as consistent as I thought at a glance. On the datatracker side, there's no such thing as an RpcPerson except where the Rpc is a prefix.

Purple has an `RpcPerson` model that represents someone who has a role with the RPC - editor, manager, etc.

PR #9096, opened 2025-07-04
fix: debug drf rpcapi

jennifer-richards:
Fixes a few bugs that were discovered while reconciling the code with the rfced-xfer project.

The refactor in cd01a16 renders unnecessary the earlier change to the `DocumentInfo.title` validation, so this PR reverts that change. The code on this branch has not been deployed anywhere, so we can just remove the migration rather than reversing it in a new one.


ietf/utils/validators.py
```diff
@@ -33,9 +33,8 @@
 # Note that this is an instantiation of the regex validator, _not_ the
 # regex-string validator defined right below
 validate_no_control_chars = RegexValidator(
-    regex="^[^\x01-\x1f]*$",
-    message="Please enter a string without control characters.",
-)
+                                    regex="^[^\x00-\x1f]*$",
```
rudimatz, line 36: indentation looks off

rudimatz (approved):

ietf/utils/validators.py
jennifer-richards: Yes indeed, good catch. However, I'm inclined to leave it because this is reverting back to the strangely indented original on the main branch. When we squash-merge the feat/rpc-api-drf branch back to main, the net result will leave the lines alone so I think it's better not to link it to this PR in the history.

PR #9113, opened 2025-07-09
fix: improve pw validation without js

jennifer-richards:
Fixes issues with poor / absent feedback about rejected passwords when passwords are reset, changed, or initially created in a browser with javascript disabled.

It's tempting to go on a deeper dive refactoring the password forms, but I think this is a sensible stopping point. This will need more serious work when we move to an external auth system and/or migrate to Django 5.

### DIFF
--- a/ietf/ietfauth/forms.py
+++ b/ietf/ietfauth/forms.py
@@ -234,9 +234,18 @@ class ChangePasswordForm(forms.Form):
         conf_password = self.cleaned_data.get("new_password_confirmation")
         if new_password != conf_password:
             raise ValidationError(
-                "The password confirmation is different than the new password"
+                {
+                    # attach error to specific field
+                    "new_password_confirmation": "The password confirmation is "
+                                                 "different than the new password"
+                }
             )
-        password_validation.validate_password(conf_password, self.user)
+        try:
+            password_validation.validate_password(conf_password, self.user)
+        except ValidationError as err:
+            raise ValidationError(
+                {"new_password": err}  # attach error to specific field
+            )
 
 
 class ChangeUsernameForm(forms.Form):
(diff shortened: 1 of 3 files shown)


ietf/ietfauth/forms.py
```diff
@@ -33,20 +33,54 @@ def clean_email(self):
...
+        password = self.cleaned_data.get("password")
+        password_confirmation = self.cleaned_data.get("password_confirmation")
         if password != password_confirmation:
-            raise forms.ValidationError("The two password fields didn't match.")
-        return password_confirmation
+            raise ValidationError(
+                "The password confirmation is different than the new password" 
+            )
+        try:
+            password_validation.validate_password(password_confirmation, self.user)
```
kesara, line 80: `clean_*` methods should always returned the cleaned values even the cleaned value is unchanged.

kesara, line 80: Isn't this redudent because `clean()` method does `password_validation.validate_password()`?

jennifer-richards: Thanks. This went back and forth between being a `clean()` and a field-specific clean and that got lost in the shuffle. (Doesn't affect anything right now because nothing else looks at the confirmation field, but it should do it properly!)

jennifer-richards: There's no `clean()` method for this class, only for `ChangePasswordForm`. That's an unrelated class. I would rather use `clean()` but can't because of the multiple inheritance in `PersonPasswordForm`, which doesn't call the methods properly.

jennifer-richards: (fixed)

jennifer-richards: Refactored `ChangePasswordForm` to be a subclass of `PasswordForm`, which will hopefully make this less confusing.

rjsparks (approved):

PR #9144, opened 2025-07-15
feat: Add "IESG" group at top of Groups menu

Spectre17:
Creates an IESG group at the top of the Group dropdown menu that points at https://datatracker.ietf.org/group/iesg/about/. Maybe that value shouldn't be hardcoded into menu.html. (Holler if that's the case.) This feature is in response to issue #9018.

### DIFF
--- a/ietf/templates/base/menu.html
+++ b/ietf/templates/base/menu.html
@@ -19,7 +19,10 @@
             <span class="fw-bolder">Groups</span>
         </li>
     {% endif %}
-    {% if flavor == 'top' %}<li class="dropdown-header">By area/parent</li>{% endif %}
+    {% if flavor == 'top' %}
+        <li><a class="dropdown-item" href="https://datatracker.ietf.org/group/iesg/about/">IESG</a></li>
+        <li class="dropdown-header">By area/parent</li>
+    {% endif %}
     {% wg_menu flavor %}
     <li class="dropend">
         <a class="dropdown-item dropdown-toggle {% if flavor != 'top' %}text-wrap{% endif %}"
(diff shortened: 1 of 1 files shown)


rjsparks:
Thanks Peter - There are changes needed - comments inline.

ietf/templates/base/menu.html
```diff
@@ -19,7 +19,10 @@
             <span class="fw-bolder">Groups</span>
         </li>
     {% endif %}
-    {% if flavor == 'top' %}<li class="dropdown-header">By area/parent</li>{% endif %}
+    {% if flavor == 'top' %}
+        <li><a class="dropdown-item" href="https://datatracker.ietf.org/group/iesg/about/">IESG</a></li>
```
rjsparks:
```suggestion
        <li><a class="dropdown-item" href="{% url "ietf.group.views.group_about" acronym="iesg" %}">IESG</a></li>
```

ietf/templates/base/menu.html
```diff
@@ -19,7 +19,10 @@
             <span class="fw-bolder">Groups</span>
         </li>
     {% endif %}
-    {% if flavor == 'top' %}<li class="dropdown-header">By area/parent</li>{% endif %}
+    {% if flavor == 'top' %}
```
rjsparks: This conditional is intended to wrap decorations that only appear in the top menu but not in the left menu. A full menu entry should appear in both.

rjsparks (changes requested):

ietf/templates/base/menu.html
Spectre17: Thanks for correction. Correction committed.

ietf/templates/base/menu.html
Spectre17: A feature I didn't even know exists. Fixed it so IESG appears in both menus.

rjsparks (approved):

Issue #9147, opened 2025-07-16
In dark mode, IESG dashboard has unreadable numbers


evyncke:
### Describe the issue

When DT is in dark mode, https://datatracker.ietf.org/doc/ad/, the numbers are barely visible (dark gray on black).

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #9171, opened 2025-07-19
fix: In dark mode, IESG dashboard has unreadable numbers

evyncke:
Addressing #9147 by re-using bootstrap --bs-body-color for labels & co in highCharts.css


holloway (approved):
LGTM :+1:

PR #9173, opened 2025-07-19
feat: Append ascii name if any 8bit UTF8 chars

### LINKED ISSUE #7167: UTF8 only in Authors list

richsalz:
Inspirited by Peter Yee's earlier work.

Fixes: #7167


bkmgit:
It maybe helpful to later add the option of using [Meng Sheng Pinyin fonts](https://github.com/MaruTama/Mengshen-pinyin-font) or [Hanzi Pinyin fonts](https://github.com/parlr/hanzi-pinyin-font) as romanization to ascii can result in information loss for many tonal languages, as an example:
- "mother" (mā, 妈)
- "ant" (má, 蚂)
- "horse" (mǎ, 马)
- "scold" (mà, 骂)

There are libraries that will also do this, for example [xpinyin](https://github.com/lxneng/xpinyin). Support for other languages could be added as need arises.

richsalz:
Hiya @bkmgit , could you create a new issue with your comment? That greatly expands the scope of this rather simple approach.

richsalz:
Hit wrong GH button, re-opening

jennifer-richards (approved):
I think we'll likely want to allow some additional Latin characters (e.g., "É" and other accented characters generally recognizable to readers of US-ASCII) without adding the ascii name, but this is a step forward.

richsalz:
According to https://www.lookuptables.com/text/extended-ascii-table, it looks like all the accented characters are in the decimal range 128-154 in case we want to make an exception for them.

thanks for the review.

bkmgit:
If the language is known, [pyicu](https://gitlab.pyicu.org/main/pyicu) has options for transliteration, there is an example in the [cheatsheet](https://gist.github.com/dpk/8325992).

However, it maybe easier to do an NFC decomposition of each character and check if it contains an ascii letter, if all the NFC decompositions contain ascii characters, keep the name, otherwise use the ascii name.  This could be done using [unicodedata](https://docs.python.org/3/library/unicodedata.html).

Ideally each person would be able to update this field since the [readme of Unidecode](https://pypi.org/project/Unidecode/) indicates there will be many corner cases that will be difficult to cover with existing software.

richsalz:
After checking with some John Levin and John Klensin, the current test – see if any byte has the 0x80 bit sit – was said to be good enough.

jennifer-richards:
Thanks for the insights, Benson, I had looked briefly and more naively at the unicodedata module. I agree it will be useful. I think Rich's test as implemented will inform us a lot as to where we get bitten by pointless extra text in practice. 

(And, perfect being enemy of the good and all, merging this will fix the entirely non-Latin text cases that inspired the issue this addresses; follow-ups that deal with additional subtleties will be welcome)

Issue #6599, opened 2023-11-04
Add ability to download uploaded markdown


DavidSchinazi:
### Description

Thanks for adding the ability to upload agendas in markdown, it's amazing. However, I've noticed that when we upload markdown, I'm only able to download the rendered HTML (that's based on the markdown) but not the markdown itself. It makes collaborating between co-chairs more difficult because if one chair uploads a -00 in markdown and the co-chair wants to edit it to create a -01, they have to email the first chair to get the markdown.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

pusateri:
This used to work when I added the feature. I tested it with the following command and got markdown. Seems like now it's broken.
```
curl --header "Accept: text/markdown" https://datatracker.ietf.org/meeting/118/materials/agenda-118-dnssd-02.md
```

But it still works if you remove the .md extension. It should work with the .md extension too.
```
curl --header "Accept: text/markdown" https://datatracker.ietf.org/meeting/118/materials/agenda-118-dnssd-02
```

pusateri:
It seems now the opposite is true. Asking for .md gives markdown as expected and no suffix gives html.

richsalz:
Yes, the default is HTML. If you want the markdown, you have to do *both* the `Accept` header field, and the `.md` extension. This is problematic for browser-based clients which do not have a convenient way to set the header.  Looking to see if this is a datatracker bug or an IETF CDN issue.

richsalz:
Even in localhost/dev mode (e.g., view-source:http://localhost:8000/meeting/122/materials/agenda-122-httpapi-00.md) it outputs html, so seems like a datatracker bug.

richsalz:
(Deleted some previous, confused, comments)

All six of these commands do the right thing, giving me markdown:
```
curl -D- https://datatracker.ietf.org/meeting/122/materials/agenda-122-httpapi-00.md
curl -D- http://localhost:8000/meeting/122/materials/agenda-122-httpapi-00.md
curl -D- --header "Accept: text/markdown" https://datatracker.ietf.org/meeting/122/materials/agenda-122-httpapi-00
curl -D- --header "Accept: text/markdown" http://localhost:8000/meeting/122/materials/agenda-122-httpapi-00
[...]

PR #9180, opened 2025-07-19
feat: Enable viewing of diffs for Statements

evyncke:
Addressing #9021 by making statement-* docs 'diffable'

### DIFF
--- a/ietf/doc/views_doc.py
+++ b/ietf/doc/views_doc.py
@@ -1122,6 +1122,7 @@ def get_diff_revisions(request, name, doc):
                 "draft",
                 "charter",
                 "conflict-review",
+                "statement",
                 "status-change",
             ]
         ]
@@ -1129,7 +1130,7 @@ def get_diff_revisions(request, name, doc):
 
     if not diffable:
         return []
-
+    
     # pick up revisions from events
     diff_revisions = []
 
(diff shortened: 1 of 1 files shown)


rjsparks:
There will have to be explicit support added to author-tools before we can merge this.
@Kesara - could you look at whether it would be reasonable to handle requests like:
```
https://author-tools.ietf.org/iddiff?url1=statement-iesg-ballot-procedures-for-documents-00&url2=statement-iesg-ballot-procedures-for-documents-01&difftype=--html
```

kesara:
> There will have to be explicit support added to author-tools before we can merge this. @kesara - could you look at whether it would be reasonable to handle requests like:
> 
> ```
> https://author-tools.ietf.org/iddiff?url1=statement-iesg-ballot-procedures-for-documents-00&url2=statement-iesg-ballot-procedures-for-documents-01&difftype=--html
> ```

At the moment author tools doesn't know about statements.
How do you get a text or markdown of a statement from DT?
For example, I don't see a way to get text statement for https://datatracker.ietf.org/doc/statement-iesg-ballot-procedures-for-documents/01/

evyncke:
So, it seems that a .TXT should be generated when a new .MD statement is issued ?

OTOH, comparing the .MD would probably be "good enough"

rjsparks:
md should be compared to md when it exists.

kesara:
@rjsparks, But does DT expose md or txt files for statements?

evyncke:
Any news about DT exposing somewhere the .MD for the statements ? Else, I could try to code one.

rjsparks:
(this may be a duplicate):
There is no simple endpoint for fetching the raw markdown of a statement.
There is an ongoing conversation at #6599 on how to do this generally, and statements should follow the outcome of that.

PR #9181, opened 2025-07-19
feat: groups menu in Reka UI

holloway:
## feat

(arguably a bug fix but we'll call it a feature)

* Reimplement 'Groups' menu in [Reka UI's Navigation Menu](https://reka-ui.com/docs/components/navigation-menu) so that we can handle very tall menu items, taller than the viewport.
  * Scrapes the existing 'Groups' bootstrap menu from the DOM to build a javascript data for the items, so that we can render them in Vue.
  * Reka can't handle `position: fixed` menu items taller than the viewport by default, so this PR uses a [Vue Teleport](https://vuejs.org/guide/built-ins/teleport) to move the Reka Viewport to the `<body>` so that it can scroll (see video).
* Remove existing Groups menu

https://github.com/user-attachments/assets/cda6eaf5-5187-4857-8b9d-78cdf22f337a

### DIFF
--- a/ietf/static/js/ietf.js
+++ b/ietf/static/js/ietf.js
@@ -140,53 +140,53 @@ function ensureDropdownOnscreen(elm) {
     })
 }
 
-$(document)
-    .ready(function () {
-        // load data for the menu
-        $.ajax({
-            url: $(document.body)
-                .data("group-menu-data-url"),
-            type: "GET",
-            dataType: "json",
-            success: function (data) {
-                for (var parentId in data) {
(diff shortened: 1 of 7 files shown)


ietf/static/js/ietf.js
```diff
@@ -140,53 +140,53 @@ function ensureDropdownOnscreen(elm) {
     })
 }
 
-$(document)
```
holloway, line 143: Delete the existing groups menu JS

PR #9185, opened 2025-07-19
fix: Improvements to registration admin views

rpcross:


jennifer-richards (dismissed):

rjsparks (dismissed):

ietf/settings.py
```diff
@@ -729,6 +729,8 @@ def skip_unreadable_post(record):
     "*~",  # tilde temp-files
     "#*",  # files beginning with a hashmark
     "500.html",  # isn't loaded by regular loader, but checked by test_500_page()
+    "templates/admin/meeting/RegistrationTicket/change_list.html",
+    "templates/admin/meeting/Registration/change_list.html",
```
jennifer-richards:
```suggestion
    "ietf/templates/admin/meeting/RegistrationTicket/change_list.html",
    "ietf/templates/admin/meeting/Registration/change_list.html",
```
What's here will work, but IMO it'd be best to use `"ietf/templates/admin..."` to encourage full path specification to avoid possible ambiguity.

The reason they work is that the globbing for relative paths  is done from the right-hand side of the path. The `ietf/` prefix is not required, but makes it a bit less likely that there'd be a collision - e.g., as written these would also ignore `ietf/secr/templates/admin/meeting...`. That's not likely to happen but if someone copied the pattern in the future with a shallower path it might come up.

jennifer-richards (changes requested):
Minor tweak requested, see inline

ietf/settings.py
rpcross: Fixed in commit b9dcf39

jennifer-richards (approved):

PR #9186, opened 2025-07-19
fix: Include text to set/update photo

richsalz:
This could be the only fix that #2209 (created 2017) is going to get, but I didn't mark it as fixes it.

This PR adds "Email support@ietf.org to update your picture", whether or not there is a photo available.

### DIFF
--- a/ietf/templates/registration/edit_profile.html
+++ b/ietf/templates/registration/edit_profile.html
@@ -41,7 +41,7 @@
                     {% include "person/photo.html" with person=person %}
                     {% endif %}
                     <p>Email <a href="support@ietf.org">support@ietf.org</a>
-                    to udpate your picture.</a></div>
+                    to udpate your picture.</p></a></div>
             </div>
         <div class="row mb-3">
             <label class="col-sm-2 col-form-label fw-bold">NomCom eligible</label>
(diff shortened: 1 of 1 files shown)


rjsparks:
Should we restrict showing this to people with some Role? 
I think if this appears generally, it could put a lot of demand on the secretariat.

(I still think we should move to gravatar or the like rather than host our own pictures fwiw)

richsalz:
Do we need to discuss gravatar in tools-discuss, or could I just start on a PR to support it?

rjsparks:
we _have_ discussed it several times already. It's better done as a tools call topic

richsalz:
I know we have talked about it a lot, I was asking if we have consensus. I'll take your comment as "no" So can we put it on the agenda to get a decision?

richsalz:
I agree that showing the text if the person has a role is the right trade-off of convenience vs load-on-secretariat, so this is a fix and Gravatar is a separate future item.

ietf/templates/registration/edit_profile.html
```diff
@@ -32,14 +32,19 @@ <h1>Your account</h1>
                 <a class="btn btn-primary" href="{% url 'ietf.ietfauth.views.change_password' %}">Change password</a>
             </div>
         </div>
-        {% if person.photo %}
```
jennifer-richards, line 35: I think this needs to come back as `{% if person.photo or person.role_set.exists %}`

jennifer-richards (changes requested):

jennifer-richards (approved):

Issue #9058, opened 2025-06-25
Sort IESG and IAB Statements Pages with Active Statements at the Top


cindymorgan:
### Description

Related to https://github.com/ietf-tools/datatracker/issues/9022, re-sort the statements pages (e.g. https://datatracker.ietf.org/group/iesg/statements/) so that statements with the state "Active" are shown together at the top, and statements with a state of "Replaced" or "Historic" are shown below that under a separate header

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

evyncke:
Issue should be closed as it the PR #9198 is merged and in production

PR #9198, opened 2025-07-20
feat: Sort IESG and IAB Statements Pages with Active Statements at the Top

evyncke:
Address issue #9058 

Remaining caveat: it is sorted by statement.state but currently only "active" and "replaced" whose alphabetical order matches the requested feature. If other states are introduced (e.g., "historic"), then the order wont't be so perfect.

### DIFF
--- a/ietf/templates/group/statements.html
+++ b/ietf/templates/group/statements.html
@@ -1,51 +1,51 @@
-    {% extends "group/group_base.html" %}
-    {# Copyright The IETF Trust 2023, All Rights Reserved #}
-    {% load origin %}
-    {% load ietf_filters person_filters textfilters %}
-    {% load static %}
-    {% block pagehead %}
-        <link rel="stylesheet" href="{% static "ietf/css/list.css" %}">
-    {% endblock %}
-    {% block group_content %}
-        {% origin %}
-        <h2 class="my-3">{{group.acronym|upper}} Statements</h2>
-        {% if request.user|has_role:"Secretariat" %}
-        <div class="buttonlist">
-            <a id="start_button"
-            class="btn btn-primary"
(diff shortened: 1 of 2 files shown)


rjsparks (dismissed):

rjsparks:
It's unlikely some other status would come in alphabetically lower than active, but if it ever did, we could build a sort that deals with it.

evyncke:
This is also my thinking, let's not overdo it right now.

ietf/templates/group/statements.html
```diff
@@ -25,16 +25,25 @@ <h2 class="my-3">{{group.acronym|upper}} Statements</h2>
...
+                </th>
+            </tr>
+        </tbody>
+        <tbody>
+        {% for statement in statement_group.list %}
             <tr>
                 <td title="{{ statement.published|date:'Y-m-d H:i:s O' }}">{{ statement.published|date:"Y-m-d" }}</td>
-                <td><a href="{% url 'ietf.doc.views_doc.document_main' name=statement.name %}">{{statement.title}}</a>
-                    {% if statement.status == "replaced" %}<span class="badge rounded-pill text-bg-warning">Replaced</span>{% endif %}
+                <td><a href="{#% url 'ietf.doc.views_doc.document_main' name=statement.name %#}">{{statement.title}}</a>
```
jennifer-richards: I think you need to uncomment the url tag in the href

jennifer-richards (changes requested):
See inline

ietf/templates/group/statements.html
evyncke: Indeed, changed and pushed to the PR.

Thanks for spotting it !

rjsparks (approved):

rjsparks (approved):

jennifer-richards (approved):

PR #9207, opened 2025-07-21
fix: workaround magic false positives

rjsparks:


jennifer-richards (approved):
Probably not fair to blame OpenSUSE 15.1 in the comment any more :-)

Issue #8783, opened 2025-04-04
WG Documents: Distinguish I-Ds on WG plate from I-Ds on IESG plate


cabo:
### Description

In https://datatracker.ietf.org/wg/WGWGWG/documents/, there is a section "active I-Ds", which munches together I-Ds that are "in the WG" and those that are "in the IESG".  It would help both chairs and people looking at a WG to distinguish these by putting them into separate sections.

(It also would help to be able to sort on status in a way that is not just alphabetical, but that doesn't fully resolve to above request -- the above should always be done.)

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

evyncke:
Would a divider header of "Outside of the WG Internet-Drafts" would be OK ?

evyncke:
With #9214 isn't it time to close this issue ?

PR #9214, opened 2025-07-22
feat: Distinguish I-Ds on WG plate from I-Ds on IESG plate

evyncke:
Address #8783 

Added a new search_heading with "Outside of the WG Internet-Draft" when IESG-state != idexists

Using "outside" will make the I-Ds outside of the WG plate be listed after "Active" and "Expired" but before "RFC"


rjsparks:
We can define a sort that is not alphabetical with very little work.

I think something like "Drafts with the IESG" would be a better heading.

I don't _think_ we currently can get into a position where we have a non-`idexists` iesg state for drafts in other streams, but we _would_ have a problem with search results for individual submissions with the IESG that would today appear in related documents.

This will have enough impact that the PR should come with tests.

evyncke:
> This will have enough impact that the PR should come with tests.

I will finally have to dig in this part of the code ;-)

evyncke:
> I think something like "Drafts with the IESG" would be a better heading.

Humm the plural filter won't work then :-( moreover there should then be a RFC Editor search_heading

rjsparks:
The plural decoration is really overkill for section headers like this - a section that has only one draft but with "drafts" in the section header isn't going damage the reader's comprehension.

"With the RFC Editor" seems good too, but this looks like something that needs some design and _definitely_ some tests to make sure we're doing the right thing with drafts of all provenance.

rjsparks:
shifting to draft while aspects discussed in-person are addressed.

evyncke:
@rjsparks when you have time, you may want to review the extended test coverage.

ietf/group/views.py
```diff
@@ -432,14 +432,15 @@ def prepare_group_documents(request, group, clist):
                 d.search_heading = "Related Internet-Drafts and RFCs"
                 docs_related.append(d)
         else:
+            if d.type_id == "draft" and d.stream_id == 'ietf' and d.get_state_slug('draft-iesg') != 'idexists': # values can be: ad-eval idexists approved rfcqueue dead iesg-eva
+                d.search_heading = "Internet-Draft at the IESG"
```
jennifer-richards: This should be `"Internet-Drafts at the IESG"` to parallel the other search headings

jennifer-richards (changes requested):
Sorry to poke at the plural discussion again, but I think the heading needs to be plural to match the other headings. I pointed at one spot and the test will need a tweak to match

Otherwise lgtm

ietf/group/views.py
rjsparks: I don't think it's quite that simple - the PR currently makes all the headings singular, but looking at the result, I agree we should move back towards being able to use the pluralizer. I'll see if I can propose a different approach.

evyncke: Indeed, @rjsparks and I had a discussion in Madrid about this topic. I do not mind either way though

rjsparks: I pushed a fairly large change in, but it should be easy to follow: https://github.com/ietf-tools/datatracker/pull/9214/commits/385325aae90d8b9be29194dfa23efc5f353579ee

jennifer-richards (approved):

PR #9227, opened 2025-07-23
fix: prevent navigation on doc search header sort click

### LINKED ISSUE #9136: Sort order is not respected
### Describe the issue

Starting at https://datatracker.ietf.org

Select "Author or Affiliation Name", put "ciena" into the box, and click "Search", which gets to this page:

https://datatracker.ietf.org/doc/search?name=&sort=&rfcs=on&activedrafts=on&by=author&author=ciena

Clicking any of the column headers immediately sorts the page by that column.

Then a moment later the page reverts to the original order - at which time the "&sort=..." value in the URL changes, reflecting which header was selected.

So there are three issues:

1. The page refresh is not required.
2. The "&sort=..." in the URL doesn't work.
3. The "&sort=..." in the URL doesn't accommodate reverse sorting.

### Code of Conduct

NGPixel:
Fix #9136

### DIFF
--- a/ietf/static/js/list.js
+++ b/ietf/static/js/list.js
@@ -247,7 +247,8 @@ $(document)
 
                 $(table)
                     .find(".sort")
-                    .on("click", function () {
+                    .on("click", function (ev) {
+                        ev.preventDefault()
                         var order = $(this)
                             .hasClass("asc") ? "desc" : "asc";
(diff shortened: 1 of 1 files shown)


rjsparks (approved):

rjsparks:
cc: @kivinen

jennifer-richards (approved):

PR #9228, opened 2025-07-23
fix: better wrap of nomcom key in session

rjsparks:


richsalz (approved):
LGTM

rjsparks:
Closing this PR and making another with the same change, but without the main to release commit. (We're going to bring this in normally rather than as an emergency patch).

PR #9270, opened 2025-07-31
chore: update deps for py3.12

jennifer-richards:
Updates `requirements.txt` to bump minimum versions to be officially compatible with py3.12 or beyond. The comments on the individual commits indicate what versions were declared to be supported by the version I bumped to.

These are the packages that did not have a py3.12-compatible declaration on pypi and where I couldn't tell conclusively from a quick look at the package repo. Tests all pass despite these and I haven't seen any failures in the very light manual poking I've done so far, so I expect most are bookkeeping rather than actual compatibility problems.
* defusedxml (indicates py3.9 support)
* django-widget-tweaks (1.5.1 supports py3.12/py3.13 but is not available on pypi yet)
* github3 (4.01 supports py3.11; is stale since 2023) **candidate for dropping**
* hashids (superseded by squids, but not compatible; only used for `Person.photo_name()` method that appears to be unused) **candidate for dropping**
* html5lib (stale since 2020, but used in tests)
* jwcrypto (py3.11, only used "aspirationally")
* mypy (stuck on 1.7.1 due to django-stubs; only py3.11 but seems to work, and we'll upgrade when we move to Django 5.x)
* oic (py3.11 but seems to work, only used in tests)
* pyang (only py3.11)
* python-magic (py3.9)
* pymemcache (py3.10 per pypi, but repo comments imply it's known to work with py3.12)
* rfc2html (py3.9)
* unidecode (py3.11; repo shows that py3.12 tests are likely passing)

A few packages needed some adjustments to work with the latest versions. This included boto3 (working around the checksum issue introduced in 1.36+), pytz (migrating meeting timezone/country fields to reflect changes in the available choices), and django-vite (adjusting the settings structure)

I dropped the djlint package because it appears to be struggling to find a maintainer and because I don't think we use it often/ever any more.


rjsparks (approved):

jennifer-richards:
For the record: verified that boto3 works against R2 with the settings modified as in this PR. It seems that getting to 1.36.7 makes those settings not _required_. We're going to keep the settings changes though as a precaution.

PR #9363, opened 2025-08-15
feat: iesg dashboard of wg documents

### LINKED ISSUE #4582: IESG View of WGs
### Description

If the spirit of the AD Workload page, (https://datatracker.ietf.org/doc/ad/), I would propose a related page summarizing the disposition of the WGs.  This was inspired by Francesca and Murray's analysis of ART.

Page title: IESG View of WGs

Table 1: WG Count
* Col 1: Area Director Name
* Col 2: Count of WG for which the AD is responsible

Table 2: WG Summary
* Col 1: WG name
* Col 2: Area
* Col 3: Responsible AD
* Col 4: Count of Adopted I-Ds
* Col 5: Count of Published RFCs
* Col 6: Count of Related I-Ds
* Col 7: RFC Publication Rate (calculated by taking Published RFC / Years in existence) 
* Col 8: Count of Open Milestones
* Col 9: Count of Open Milestones in the past
* Col 10: Last F2F or interim meeting date

### Code of Conduct

rjsparks:
Fixes #4582


ietf/iesg/tests.py
```diff
@@ -182,6 +185,1578 @@ def test_ietf_activity(self):
...
+
+        self.assertEqual(
+            area_totals, {"group_count": 2, "doc_count": 2, "page_count": 13}
+        )
+
+        for d in ad_summary:
+            if d["ad"] == "Lima":
+                self.assertEqual(d["doc_group_count"], 1)
+                self.assertEqual(d["doc_count"], 1)
+                self.assertEqual(d["page_count"], 9 if d["area"] == "violet" else 4)
```
jennifer-richards, line 1097: You're not counting `dicts` in this block like you did in others - is something constraining the test to be sure that both branches of this assertion are visited?

ietf/iesg/tests.py
```diff
@@ -182,6 +185,1578 @@ def test_ietf_activity(self):
...
+                "groups_with_docs_count": 0,
+            },
+        )
+        self.assertEqual(wg_summary, [])
+
+        # Areas with WGs following the "normal" pattern
+        areas = {}
+        for area_acronym in ["red", "orange", "yellow", "green", "blue", "violet"]:
+            areas[area_acronym] = GroupFactory(type_id="area", acronym=area_acronym)
+        for ad, area, wgs in [
```
jennifer-richards, line 249: I think a comment - at least a warning - that there's method to the naming madness would be appropriate here. This should warn at least that there's significance to the first letter of the wg acronyms.

ietf/iesg/utils.py
```diff
@@ -39,30 +52,269 @@ def telechat_page_count(date=None, docs=None, ad=None):
...
+    pages_for_area = defaultdict(lambda: 0)
+    docs_for_area = defaultdict(lambda: 0)
+    groups_for_ad = defaultdict(lambda: defaultdict(set))
+    pages_for_ad = defaultdict(lambda: defaultdict(lambda: 0))
+    docs_for_ad = defaultdict(lambda: defaultdict(lambda: 0))
+    groups_for_noad = defaultdict(lambda: defaultdict(set))
+    pages_for_noad = defaultdict(lambda: defaultdict(lambda: 0))
+    docs_for_noad = defaultdict(lambda: defaultdict(lambda: 0))
+    docs_for_wg = defaultdict(lambda: 0)
+    pages_for_wg = defaultdict(lambda: 0)
```
jennifer-richards, line 128: defaultdict are a handy but can cause bugs by adding keys if later uses, e.g., have a typo. Once you've built up your data, you can set `groups_for_area.default_factory = None` (etc) and that will effectively turn it into a standard dictionary.  https://docs.python.org/3/library/collections.html#collections.defaultdict.default_factory

jennifer-richards (changes requested):
A couple minor suggestions and a question that may or may not be a request for a tweak to the test.

ietf/iesg/tests.py
rjsparks: I think the previous tests covered it and the repetition was hurting using the tests to explain what the code is supposed to do.  If we were to change this test significantly, though, one thing we could do create a utility that looked at the returned dicts that validated exactly one dict per expected "key" was present.

ietf/iesg/utils.py
rjsparks: That's not helpful for the way _these_ dicts are being used. The property that an access for an unknown key returns a 0 is important and is used until the last mention of each of those dictionaries.

jennifer-richards: Agree - I misread the code and thought the defaultdicts were included in the return value.

jennifer-richards (approved):

PR #9505, opened 2025-09-11
fix: normalize 3gpp groups and resolve duplication

rjsparks:


ietf/group/tasks.py
```diff
@@ -113,3 +115,82 @@ def generate_wg_summary_files_task():
...
+        for old_acronym, new_acronym, new_name in (
+            ("o3gpptsgran3", "3gpp-tsg-ran-wg3", "3GPP TSG RAN WG3"),
+            ("3gpp-tsgct-ct1", "3gpp-tsg-ct-wg1", "3GPP TSG CT WG1"),
+            ("3gpp-tsgct-ct3", "3gpp-tsg-ct-wg3", "3GPP TSG CT WG3"),
+            ("3gpp-tsgct-ct4", "3gpp-tsg-ct-wg4", "3GPP TSG CT WG4"),
+            ("3gpp-tsgran", "3gpp-tsg-ran", "3GPP TSG RAN"),
+            ("3gpp-tsgran-ran2", "3gpp-tsg-ran-wg2", "3GPP TSG RAN WG2"),
+            ("3gpp-tsgsa-sa2", "3gpp-tsg-sa-wg2", "3GPP TSG SA WG2"),
+            ("3gpp-tsgsa-sa3", "3gpp-tsg-sa-wg3", "3GPP TSG SA WG3"),
+            ("3gpp-tsgsa-sa4", "3gpp-tsg-sa-wg4", "3GPP TSG SA WG4"),
```
jennifer-richards, line 182: Add a paranoid precondition check that there are no groups using any of the `new_acronym` values?

jennifer-richards (changes requested):
I think one more precondition check would be a good idea to ensure this doesn't fail.

From the email thread on the subject, I think we also need a fixup of the name for 3gpp-tgst-wg2 to remove hyphens and of 3gpp-tsg-ct for capitalization if we want to be completely consistent. Those could easily be done through the admin though.

jennifer-richards (dismissed):

jennifer-richards (approved):

PR #9538, opened 2025-09-15
fix: Rev Fix Option 2 - Htmlized url regex

WhippsP:
Removing single value revision numbers as that is against the naming standard (https://authors.ietf.org/naming-your-internet-draft#version) and causes issues with htmlized documents with -1 in the name (eg draft-ietf-oauth-v2-1).

Second Option - Update htmlized URL Ref directly

### DIFF
--- a/ietf/doc/urls.py
+++ b/ietf/doc/urls.py
@@ -75,7 +75,7 @@ urlpatterns = [
 # This block should really all be at the idealized docs.ietf.org service
     url(r'^html/(?P<name>bcp[0-9]+?)(\.txt|\.html)?/?$', RedirectView.as_view(url=settings.RFC_EDITOR_INFO_BASE_URL+"%(name)s", permanent=False)), 
     url(r'^html/(?P<name>std[0-9]+?)(\.txt|\.html)?/?$', RedirectView.as_view(url=settings.RFC_EDITOR_INFO_BASE_URL+"%(name)s", permanent=False)), 
-    url(r'^html/%(name)s(?:-%(rev)s)?(\.txt|\.html)?/?$' % settings.URL_REGEXPS, views_doc.document_html),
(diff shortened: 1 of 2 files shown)


rjsparks:
This patch looks like it will improve the use case that motivated it without making the other edges worse. The branch survived a test crawl (a sample of a full crawl, but it's served us well in the past to expose unexpected behavior early).

One point to note: In `fuzzy_find_documents`:
https://github.com/ietf-tools/datatracker/blob/c71871855769d9c2980cad853cf92a9ec25cb50a/ietf/doc/utils.py#L1212-L1253

There is one bit of code that will never be executed when called from `document_html` (it will still be reached when called from `document_main` and other views):
https://github.com/ietf-tools/datatracker/blob/c71871855769d9c2980cad853cf92a9ec25cb50a/ietf/doc/utils.py#L1229-L1231

Arguably the real bug (or one facet of it) is those 3 lines, but the heuristic logic to deal with determining whether a string ending with a single or double digit is an attempt to reference the unversioned name of a document or a specific version of a document is currently spread between views that call each other and workarounds in those views compete with each other to introduce new unintended behavior. (If we could send messages back to the past, I think we'd strongly discourage overloading URL path segments with "this could be a name, or it could be a name with a version" entirely, but that expectation has been pushed into so many places, changing it is unfeasible).

But as noted in the issue, the workaround in this PR is making one case better without obviously making other cases worse, and unless a second set of eyes on it turns up something, I expect to merge it.

When we rebuild these views, further moving these heuristics into a place where the effect of changes can be tested more easily than they can be now will happen.

rjsparks (approved):

jennifer-richards (approved):
Agree that this fixes the corner case without much risk of breaking in unexpected places.

Issue #8271, opened 2024-11-25
CC'ing when sending LSs over datatracker


mirjak:
### Description

If people send LSs through the datatracker, the email generated comes from statements@ietf.org and the person listed in the From flied doesn't get a copy. It would be good to also cc that person, especially for cases where the LS was sent to statements@ietf.org and is uploaded by the secretariat manually. The original sender will get a note when the secretariat ticket is closed (and processed), however, it would be way nicer to have that person directly cc'ed in the send-out of the LS itself. 

As a side note, if a LS is sent to a wg, the chairs show up in the To field and the mailing list only in the cc. That seems also unexpected. I would expect to have the wg list in To, either even just the list address or both the list address and the chairs directly. Sorry if this should be a separate issue, but it seemed easier for me to mention this here as well; also this part is not really a problem that needs fixing, it's more a note.

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
The first change above (cc-ing the from contact) can be realized with a data adjustments. The mailtrigger concept was intended to make this kind of change easy.

We would need a new mailtrigger/recipient of `liaison_from_contact` expanding to `{{liaison.from_contact}}` and then that recipient would be added to the cc: recipients in https://datatracker.ietf.org/mailtrigger/name/liaison_statement_posted/.

I can make those changes as an incremental improvement through the admin.

The second (moving the group list (if the target is a group) to To vs CC requires adjusting code here: https://github.com/ietf-tools/datatracker/blob/b1cfa7082f60343210b8116668f182e3c67207bf/ietf/liaisons/views.py#L100-L158. We could do that now as incremental improvement, or wait until we rework the application and move all of that code into mailtriggers to ease future maintenance.

rjsparks:
> I can make those changes as an incremental improvement through the admin.

I've done this - see https://datatracker.ietf.org/mailtrigger/name/liaison_statement_posted/

rjsparks:
Moved the second observation to its own issue.

mirjak:
Thanks! Actually this is one of the issues where the discussion move on a bit. (Sorry I know this is a bit of a mess currently but that's also why a proposed to reorg the current issues maybe).

[...]

PR #9553, opened 2025-09-18
feat: split liaison_statement_posted mailtrigger into outgoing and incoming

### LINKED ISSUE #9434: When sending an incoming liaison statement, move the list address for any groups from the `Cc` to the `To` field.
### Description

See the comment in https://github.com/ietf-tools/datatracker/issues/8271.

This should be implemented by moving the code at https://github.com/ietf-tools/datatracker/blob/b1cfa7082f60343210b8116668f182e3c67207bf/ietf/liaisons/views.py#L100-L158 into mailtriggers, and adjusting the triggers to move the lists from cc to to.

### Code of Conduct

ghost:
Fixes [#9434](https://github.com/ietf-tools/datatracker/issues/9434)

### DIFF
--- a/ietf/mailtrigger/migrations/0008_liaison_statement_incoming_and_outgoing_posted.py
+++ b/ietf/mailtrigger/migrations/0008_liaison_statement_incoming_and_outgoing_posted.py
@@ -1,0 +1,63 @@
+# Copyright The IETF Trust 2023, All Rights Reserved
+
+from django.db import migrations
+
+
+def forward(apps, schema_editor):
+    Mailtrigger = apps.get_model("mailtrigger", "MailTrigger")
+    Recipient = apps.get_model("mailtrigger", "Recipient")
+    recipients_to = Recipient.objects.get(pk="liaison_to_contacts")
+    recipients_cc = list(
+        Recipient.objects.filter(
+            slug__in=(
+                "liaison_cc",
+                "liaison_coordinators",
+                "liaison_response_contacts",
+                "liaison_technical_contacts",
+            )
+        )
+    )
+    recipient_from = Recipient.objects.get(pk="liaison_from_contact")
+
+
+    liaison_posted_outgoing = Mailtrigger.objects.create(
+        slug="liaison_statement_posted_outgoing",
+        desc="Recipients for a message when a new outgoing liaison statement is posted", 
+    )
+    liaison_posted_outgoing.to.add(recipients_to)
+    liaison_posted_outgoing.cc.add(*recipients_cc, recipient_from)
+
+
+
+    liaison_posted_incoming = Mailtrigger.objects.create(
+        slug="liaison_statement_posted_incoming",
+        desc="Recipients for a message when a new incoming liaison statement is posted", 
+    )
+    liaison_posted_incoming.to.add(recipients_to, recipient_from)
+    liaison_posted_incoming.cc.add(*recipients_cc)
+
+
+    Mailtrigger.objects.filter(slug__in=("liaison_statement_posted")).delete() 
+
+ 
+def reverse(apps, schema_editor):
+    Mailtrigger = apps.get_model("mailtrigger", "MailTrigger")
+    
+    Mailtrigger.objects.filter(
+        slug__in=("liaison_statement_posted_outgoing", "liaison_statement_posted_incoming")
+    ).delete()   
+
+    Mailtrigger.objects.get(
+        slug="liaison_statement_posted", 
+        desc="Recipients for a message when a new liaison statement is posted", 
+    )
+
+    
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("mailtrigger", "0007_historicalrecipient_historicalmailtrigger")
+    ]
+
+    operations = [migrations.RunPython(forward, reverse)]
+
--- a/ietf/mailtrigger/migrations/0008_liaison_statement_incoming_and_outgoing_posted.py
+++ b/ietf/mailtrigger/migrations/0008_liaison_statement_incoming_and_outgoing_posted.py
@@ -39,20 +39,33 @@ def forward(apps, schema_editor):
 
     Mailtrigger.objects.filter(slug__in=("liaison_statement_posted")).delete() 
 
- 
+
 def reverse(apps, schema_editor):
     Mailtrigger = apps.get_model("mailtrigger", "MailTrigger")
+    Recipient = apps.get_model("mailtrigger", "Recipient")
     
     Mailtrigger.objects.filter(
         slug__in=("liaison_statement_posted_outgoing", "liaison_statement_posted_incoming")
     ).delete()   
 
-    Mailtrigger.objects.get(
+    liaison_statement_posted = Mailtrigger.objects.create(
         slug="liaison_statement_posted", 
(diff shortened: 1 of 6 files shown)


rjsparks:
Please put `playwright/tests/meeting/floor-plan.spec.js` back. If you're seeing playwright failures, they aren't related to what you're working on - call them out to us and we'll identiffy what's happening.

Then make sure migrations ran correctly in your dev instance and regenerate the names fixture with
```
ietf/manage.py generate_name_fixture --stdout | jq --sort-keys 'sort_by(.model, .pk)' > ietf/name/fixtures/names.json
```
(You can find that using `ietf/manage.py generate_name_fixture --help`).
and commit the updated result.

ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py
```diff
@@ -0,0 +1,48 @@
...
+    liaison_posted_outgoing.cc.add(*recipients_cc)
+
+
+    liaison_posted_incoming = Mailtrigger.objects.create(
+        slug="liaison_statement_posted_incoming",
+        desc="Recipients for a message when a new incoming liaison statement is posted",
+    )
+    liaison_posted_incoming.to.add(recipients_to)
+    liaison_posted_incoming.to.add(*recipients_cc)
+
```
rjsparks, line 38: This should remove `liaison_statement_posted`

ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py
```diff
@@ -0,0 +1,48 @@
...
+
+def reverse(apps, schema_editor):
+    MailTrigger = apps.get_model("mailtrigger", "MailTrigger")
+    MailTrigger.objects.filter(
+        slug__in=("liaison_statement_posted_outgoing", "liaison_statement_posted_incoming")
+    ).delete()   
+
```
rjsparks, line 50: and this should add `liaison_statement_posted` back

ietf/mailtrigger/utils.py
```diff
@@ -66,6 +74,68 @@ def get_mailtrigger(slug, create_from_slug_if_not_exists, desc_if_not_exists):
     return mailtrigger
 
 
+def get_cc(group):
```
rjsparks: This method name used to rely (too much) on the file it was in for context. With this move, it should be something like `get_contacts_for_liaison_messages_for_group`. 

This and the function after it should be merged into one that pays attention to outgoing/incoming, or perhaps one should be named ..._primary_... and the other ..._secondary_...

Whether these contacts land in the to or the cc should be up to how the mailtriggers use them, so getting the notion of cc or to out of the name is important. (See the title of #9434).

ietf/mailtrigger/utils.py
```diff
@@ -66,6 +74,68 @@ def get_mailtrigger(slug, create_from_slug_if_not_exists, desc_if_not_exists):
...
+        liaiman_roles = group.role_set.filter(name='liaiman')
+        emails.extend([ contact_email_from_role(r) for r in liaiman_roles ])
+
+    # explicit CCs
+    liaison_cc_roles = group.role_set.filter(name='liaison_cc_contact')
+    emails.extend([ contact_email_from_role(r) for r in liaison_cc_roles ])
+
+    return emails
+
+def get_contacts_for_group(group):
```
rjsparks: Similarly, this should be something like `get_contacts_for_liaison_messages_for_group` though that's getting a bit long.

ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py
```diff
@@ -0,0 +1,48 @@
...
+                "liaison_technical_contacts",
+            )
+        )
+    )
+    liaison_posted_outgoing = Mailtrigger.objects.create(
+        slug="liaison_statement_posted_outgoing",
+        desc="Recipients for a message when a new outgoing liaison statement is posted",
+    )
+    liaison_posted_outgoing.to.add(recipients_to)
+    liaison_posted_outgoing.cc.add(*recipients_cc)
```
rjsparks: See #8271 - lets add the new Recipient liaison_from_contact to the cc for outgoing.

rjsparks: Also see the title of #9434 - lets see if we can make that move from `cc` to `to` with this migration.

rjsparks (changes requested):
Thanks again for helping. Comments inline.

Please let me know if the next set of changes aren't obvious.

ietf/mailtrigger/utils.py
rjsparks: (should have edited that before posting it to match the comments above. What's said at L77 is what's intended.)

ietf/mailtrigger/utils.py
ghost: I've renamed the functions, would they still need outgoing/incoming separate from the mails logic if merged into one inside utils? CC comments need to be revised too.

ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py
ghost: Have attempted to remove with 

````
Mailtrigger.objects.filter(slug__in=("liaison_statement_posted")).delete() 
````

File: ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py (hunk 2 above)
Comment (ghost) [reply ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py:50]: How would you recommend adding this back inside reverse? NB fixture might still be marking `liaison_statement_posted` as active in this branch

File: ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py (hunk 5 above)
Comment (ghost) [reply ietf/mailtrigger/migrations/0008_liaison_outgoing_and_incoming_statement_posted.py]: Have added this to cc for outgoing and to for incoming, ty for noting parent issue

File: ietf/mailtrigger/migrations/0008_liaison_statement_incoming_and_outgoing_posted.py [hunk 6]
```diff
@@ -0,0 +1,63 @@
...
+
+ 
+def reverse(apps, schema_editor):
+    Mailtrigger = apps.get_model("mailtrigger", "MailTrigger")
+    
+    Mailtrigger.objects.filter(
+        slug__in=("liaison_statement_posted_outgoing", "liaison_statement_posted_incoming")
+    ).delete()   
+
+    Mailtrigger.objects.get(
```
rjsparks: This needs to be a create and the Recipient m2ms will have to be repopulated.

rjsparks (changes requested):
If you have time, please make this change, otherwise, we'll finish this out in the next few days.

rjsparks:
I've reworked the migration a bit.

rjsparks (approved):

jennifer-richards (approved):

PR #9605, opened 2025-09-24
feat: add consensus on FullDraftSerializer

rudimatz:


jennifer-richards (changes requested):
This code is fine, but the `default_consensus()` method is badly broken. It should probably be fixed in this PR. Discussing this elsewhere.

rudimatz:
@jennifer-richards I think a refactoring of `default_consensus()` should happen outside (and independent) from this PR, do you agree?

jennifer-richards:
> @jennifer-richards I think a refactoring of `default_consensus()` should happen outside (and independent) from this PR, do you agree?

That's fine. It's so strangely broken it'd be ok with me to fix it here, but either way.

PR #9607, opened 2025-09-24
feat: editable wglc messages

rjsparks:


rjsparks:
~~missed committing tests for the new form - will be added soon.~~ Done

ietf/doc/views_draft.py
```diff
@@ -1857,3 +1883,122 @@ def set_intended_status_level(request, doc, new_level, old_level, comment):
...
+        form = IssueWorkingGroupLastCallForm(request.POST)
+        if form.is_valid():
+            # Intentionally not changing tags or adding a comment
+            # those things can be done with other workflows
+            by = request.user.person
+            prev_state = doc.get_state("draft-stream-ietf")
+            events = []
+            wglc_state = State.objects.get(type="draft-stream-ietf", slug="wg-lc")
+            doc.set_state(wglc_state)
+            e = add_state_change_event(doc, by, prev_state, wglc_state)
```
jennifer-richards, line 1969: I think this point can now be reached with `prev_state == wglc_state`, in which case `e` will be `None`. Is that intended (or is something preventing it)?

rjsparks: good catch - not allowing tags/comment in this form makes allowing the view questionable at all if the document is already currently in the wg-lc state. I'll work through whether being able to use this view to update WGLC in progress is the right thing to or not.

rjsparks: the change_stream_state function in main doesn't update wg-lc in progress - it only allows tweaking tags/comments (and it still will in this branch), so I think preserving the current behavior is best for this bit of incremental improvement. I'll adjust things so that this view is not available if the doc is already in wg-lc.

jennifer-richards (approved):

PR #9611, opened 2025-09-25
feat: more RfcAuthor support

jennifer-richards:
Completes a few TODO items, fixes the admin handling of RfcAuthor records, and fixes various minor things.

The "BIG TODO" for RFC Editor sync is removed without implementation. That's a big enough task we'll track it out of band rather than via a code TODO.

### DIFF
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -895,7 +895,11 @@ class RfcAuthor(models.Model):
     ignored.
     """
 
-    document = ForeignKey("Document", on_delete=models.CASCADE)
+    document = ForeignKey(
+        "Document",
+        on_delete=models.CASCADE,
+        limit_choices_to={"type_id": "rfc"},  # only affects ModelForms (e.g., admin)
+    )
     titlepage_name = models.CharField(max_length=128, blank=False)
     is_editor = models.BooleanField(default=False)
     person = ForeignKey(Person, null=True, blank=True, on_delete=models.PROTECT)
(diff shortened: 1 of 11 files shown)


rudimatz (approved):

ietf/doc/models.py
```diff
@@ -410,16 +409,20 @@ def friendly_state(self):
 
     def author_names(self):
         names = []
-        if self.type_id=="rfc" and self.rfcauthor_set.exists():
+        if self.type_id == "rfc" and self.rfcauthor_set.exists():
```
rjsparks, line 412: Checking to see if we should make things more consistent. This returns names in favor of titlepage_names. Further down in the PR, we return titlepage_names in favor of names.

ietf/doc/views_doc.py
```diff
@@ -1643,12 +1643,19 @@ def extract_name(s):
...
+        else doc.documentauthor_set
+    ).select_related("person", "email").order_by("order")
     data["authors"] = [
-        dict(name=author.person.name,
-             email=author.email.address if author.email else None,
-             affiliation=author.affiliation)
-        for author in doc.documentauthor_set.all().select_related("person", "email").order_by("order")
-    ] # TODO: report differntly if we have RfcAuthors
+        {
+            "name": author.titlepage_name if hasattr(author, "titlepage_name") else author.person.name,
```
rjsparks, line 1653: here we favor titlepage_name

rjsparks (approved):
Approving but with one question to discuss pre or post merge.

jennifer-richards:
Merging as-is, but good point re: reconciling priority of titlepage name vs person name.

PR #9617, opened 2025-09-26
refactor: move session request tool to ietf.meeting and restyle

rpcross:
restyle to match standard Datatracker


ietf/secr/urls.py
```diff
@@ -6,6 +6,5 @@
     re_path(r'^announcement/', include('ietf.secr.announcement.urls')),
     re_path(r'^meetings/', include('ietf.secr.meetings.urls')),
     re_path(r'^rolodex/', include('ietf.secr.rolodex.urls')),
-    re_path(r'^sreq/', include('ietf.secr.sreq.urls')),
```
rjsparks, line 9: This should be a permanent RedirectView to the new location for at least one meeting cycle

ietf/templates/meeting/session_request_view.html
```diff
@@ -0,0 +1,84 @@
...
+      <button type="button" class="btn btn-primary me-2" onclick="if (window.confirm('Do you really want to cancel this session?')) { window.location='{% url 'ietf.meeting.views_session_request.cancel_request' acronym=group.acronym %}' };" {% if not can_cancel %} disabled{% endif %}>Cancel this Request</button>
+      <a href="{% url 'ietf.meeting.views_session_request.list_view' %}" class="btn btn-secondary float-end">Back</a>
+  
+    </div> <!-- view-container -->
+
+{% endblock %}
+
+{% block js %}
+
+    <script>
```
rjsparks: Should we move this out to a js file as part of this pr?

ietf/templates/meeting/session_request_confirm.html
```diff
@@ -0,0 +1,63 @@
...
+
+        </form>
+
+    </div> <!-- view-container -->
+
+{% endblock %}
+
+{% block js %}
+
+    <script>
```
rjsparks: and another script we may want to move to a js file now?

rjsparks (changes requested):
Please review and update all the copyright lines for modified files.

ietf/secr/urls.py
rpcross: fixed in commit c612a2a11cf0984a9f3238d21d20178338f2b5c5

ietf/templates/meeting/session_request_view.html
rpcross: fixed in commit 38c007a917249f7ecebe5d491cc059b19166831a

ietf/templates/meeting/session_request_confirm.html
rpcross: fixed in 38c007a917249f7ecebe5d491cc059b19166831a

rjsparks (dismissed):

ietf/secr/urls.py
```diff
@@ -1,11 +1,12 @@
...
 from django.views.generic import TemplateView
+from django.views.generic.base import RedirectView
 
 urlpatterns = [
     re_path(r'^$', TemplateView.as_view(template_name='index.html'), name='ietf.secr'),
     re_path(r'^announcement/', include('ietf.secr.announcement.urls')),
     re_path(r'^meetings/', include('ietf.secr.meetings.urls')),
     re_path(r'^rolodex/', include('ietf.secr.rolodex.urls')),
-    re_path(r'^sreq/', include('ietf.secr.sreq.urls')),
+    re_path(r'^sreq/', RedirectView.as_view(url='/meeting/session/request/', permanent=True)),
```
jennifer-richards: Do any of the sub-paths need redirects? E.g., `sreq/acronym/` or `sreq/acronym/new/`. I don't think we have promoted any of them as stable, but people might have bookmarks for them.

package.json
```diff
@@ -148,6 +148,9 @@
         "ietf/static/js/password_strength.js",
         "ietf/static/js/select2.js",
         "ietf/static/js/session_details_form.js",
+        "ietf/static/js/session_form.js",
+        "ietf/static/js/session_utils.js",
+        "ietf/static/js/sessions.js",
```
jennifer-richards: I think `session_request.js` would be a better name for `session.js`. We have other pages that deal with sessions in other ways that are not likely to use this JS and it may cause confusion later.

The `session_utils.js` name doesn't seem quite appropriate given that its use is for an immediate side effect rather than to access utility methods. I feel less strongly about this because `utils` files are always a mishmash, but it seems more like `custom_striped.js` or something that better reflects what it's actually doing?

jennifer-richards (changes requested):
Looks ok. I think a couple of js files should be renamed to avoid confusion and have a question about whether the single redirect for the main sreq page is adequate.

Not for this PR, but several of these views have side effects on GET requests. That's an anti-pattern and we should refactor in the nearish future to move these to POSTs instead.

ietf/secr/urls.py
rjsparks: probably a good idea

rpcross: Added redirects in commit 27a5389765abd78b0274b8bd7efa5bf8fda1ce8d

package.json
rpcross: Changed javascript filenames in commit 27a5389765abd78b0274b8bd7efa5bf8fda1ce8d

jennifer-richards (approved):
Thanks!

jennifer-richards:
Merging to a feature branch for sandbox deployment. Test failures are coverage reduction only, so ignoring.

PR #9629, opened 2025-09-29
feat: use rfcauthors for nomcom eligibility

jennifer-richards:
Refactors computation of the authorship path for nomcom eligibility to fix up some stale logic. Uses `RfcAuthor` records instead of `DocumentAuthor` for RFCs that have the former.

### DIFF
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -7,6 +7,7 @@ import datetime
 import logging
 import os
 
+import django.db
 import rfc2html
 
 from io import BufferedReader
--- a/ietf/nomcom/tests.py
+++ b/ietf/nomcom/tests.py
@@ -27,8 +27,14 @@ import debug                            # pyflakes:ignore
 from ietf.api.views import EmailIngestionError
 from ietf.dbtemplate.factories import DBTemplateFactory
 from ietf.dbtemplate.models import DBTemplate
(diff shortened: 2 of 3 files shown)


rjsparks (approved):
This will work best if the creation of rfcauthors via merged rfc and datatracker data preserves datatracker person links for authors on the last draft that remained authors on the RFC. I believe that merge has that property, but if it doesn't we should look to fix it.

jennifer-richards:
> This will work best if the creation of rfcauthors via merged rfc and datatracker data preserves datatracker person links for authors on the last draft that remained authors on the RFC. I believe that merge has that property, but if it doesn't we should look to fix it.

Assuming you mean the place that `DocumentAuthor` records are created for the `Document(type="rfc")`, then yes, those are exact copies of the records that were tied to the final datatracker draft at the time of publication.

When we start creating `RfcAuthor` data we'll need to be careful to maintain the `Person` links, but afaik we have not done that yet. If I've missed something, please let me know.

rjsparks:
> When we start creating RfcAuthor data we'll need to be careful to maintain the Person links

Is what I was talking about. I believe the merge we're doing creating purple's database has this property - it is _intended_ to, but we should verify that nothing is perturbing it.

rjsparks (approved):

rjsparks (approved):

PR #9631, opened 2025-09-29
feat: RPC modernization APIs

jennifer-richards:


dev/deploy-to-container/settings_local.py
```diff
@@ -71,11 +71,10 @@
 
 DE_GFM_BINARY = '/usr/local/bin/de-gfm'
 
-# No real secrets here, these are public testing values _only_
 APP_API_TOKENS = {
-  "ietf.api.views.ingest_email_test": ["ingestion-test-token"]
+  "ietf.api.views.ingest_email_test": ["ingestion-test-token"],  # Not a real secret
+  "ietf.api.views_rpc" : ["devtoken"],  # For RPC dev work only, remove before merging to main!!!
```
rjsparks: Replace this with "not a real secret"?

docker/configs/settings_local.py
```diff
@@ -100,3 +100,7 @@
             bucket_name=f"{storagename}",
         ),
     }
+
+APP_API_TOKENS = {
+    "ietf.api.views_rpc" : ["devtoken"],
```
rjsparks: not a real secret?

ietf/api/urls_rpc.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+urlpatterns = [
+    url(r"^doc/drafts_by_names/", views_rpc.DraftsByNamesView.as_view()),
+    url(r"^persons/search/", views_rpc.RpcPersonSearch.as_view()),
+    path(r"subject/<str:subject_id>/person/", views_rpc.SubjectPersonView.as_view()),
+]
+
+# add routers at the end so individual routes can steal parts of their address
+# space (specifically, ^person/ routes so far)
```
rjsparks: verify that this is current?

ietf/api/views_rpc_demo.py
```diff
@@ -0,0 +1,98 @@
+# Copyright The IETF Trust 2023-2025, All Rights Reserved
```
rjsparks: Should this capability come to main? I'm thinking it shouldn't.

ietf/doc/models.py
```diff
@@ -714,14 +724,22 @@ def contains(self):
     
     def part_of(self):
         return self.related_that("contains")
-
+
```
rjsparks: whitespace

ietf/doc/models.py
```diff
@@ -714,14 +724,22 @@ def contains(self):
...
             refs_to |= self.came_from_draft().referenced_by_rfcs()
         return refs_to
-    
+
+    def sent_to_rfc_editor_event(self):
+        if self.stream_id == "ietf":
+            return self.docevent_set.filter(type="iesg_approved").order_by("-time").first()
+        elif self.stream_id in ["iab", "irtf", "ise"]:
+            return self.docevent_set.filter(type="requested_publication").order_by("-time").first()
+        #elif self.stream_id == "editorial": #TODO
```
rjsparks: we have one now to look at and then add this branch to the logic

rjsparks (commented):
First pass inspecting "can this safely come into main".
I haven't done the "are the apis all sufficiently access guarded" read yet.

ietf/api/routers.py
```diff
@@ -3,14 +3,29 @@
...
+            raise TypeError("Must mixin with superclass that has register() method")
+        super_register(prefix, viewset, basename=f"{self.name_prefix}.{basename}")
+
+
+class PrefixedSimpleRouter(PrefixedBasenameMixin, routers.SimpleRouter):
+    """SimpleRouter that adds a dot-separated prefix to its basename"""
+
+
+class PrefixedDefaultRouter(PrefixedBasenameMixin, routers.DefaultRouter):
+    """SimpleRouter that adds a dot-separated prefix to its basename"""
```
jennifer-richards:
```suggestion
    """DefaultRouter that adds a dot-separated prefix to its basename"""
```

docker/configs/settings_local.py
jennifer-richards:
```suggestion
    "ietf.api.views_rpc" : ["devtoken"],  # Not a real secret
```

dev/deploy-to-container/settings_local.py
jennifer-richards:
```suggestion
  "ietf.api.views_rpc" : ["devtoken"],  # Not a real secret
```

ietf/api/urls_rpc.py
jennifer-richards: The specific example was outdated. This will use a current example:
```suggestion
# space (e.g., ^rfc/publish/ superseding the ^rfc/ routes of RfcViewSet)
```

ietf/api/views_rpc_demo.py
jennifer-richards: absolutely not

ietf/person/models.py
```diff
@@ -203,7 +203,7 @@ def has_drafts(self):
 
     def rfcs(self):
         from ietf.doc.models import Document
-        rfcs = list(Document.objects.filter(documentauthor__person=self, type='rfc'))
+        rfcs = list(Document.objects.filter(type="rfc").filter(models.Q(documentauthor__person=self)|models.Q(rfcauthor__person=self)).distinct())
```
rjsparks, line 209: We've shifted to checking if rfcauthor exists and only using that in other places. IIRC we decided to use the union here to be sure we over-, rather than under-, qualify people for nomcom. Given the evolution of the rest of the code, is it still the right choice? I can see leaving it this way, but perhaps we leave a comment for future us?

rjsparks (dismissed):

ietf/person/models.py
jennifer-richards: I agree this overcounts and I believe you're right re: nomcom counting. We should fix this, but it's rare enough not to be urgent. I added a comment here and will create a ticket.

rjsparks (approved):

rjsparks:
The test failures are only the expected changes in coverage (since the denominator moved).

PR #9646, opened 2025-10-02
feat: preview ballot email before save

rjsparks:


ietf/doc/views_ballot.py
```diff
@@ -301,21 +322,98 @@ def err(code, text):
...
+                "position",
+                "balloter",
+                "docname",
+                "cc_choices",
+                "additional_cc",
+            ]:
+                if key not in post_data:
+                    errors.append(f"{key} not found in post_data")
+    if len(errors) == 0:
+        person = Person.objects.filter(pk=post_data.get("balloter")).first()
```
jennifer-richards, line 352: Ideally should validate, at least for null characters, before using the parameters to build a query. Given the restricted access to this endpoint, it's not essential, but probably good practice. We have `ietf.utils.validators.validate_no_nulls()` for this purpose.

(Could also validate things through a form using data from the JSON body, but that is probably a distraction at this point)

ietf/templates/doc/ballot/edit_position.html
```diff
@@ -78,4 +119,44 @@ <h1>
 
         discussToggle($("input[name=position]:checked").val());
     </script>
+
+    <script>
```
jennifer-richards, line 123: Should probably refactor away from inline scripts

ietf/doc/views_ballot.py
```diff
@@ -301,21 +322,98 @@ def err(code, text):
...
+        wanted = dict()  # consider named tuple instead
+        wanted["discuss"] = post_data.get("discuss")
+        wanted["comment"] = post_data.get("comment")
+        wanted["position_name"] = post_data.get("position")
+        wanted["balloter"] = person
+        wanted["doc"] = doc
+        addrs, frm, subject, body = build_position_email_from_dict(wanted)
+
+        recipient_slugs = post_data.get("cc_choices")
+        # Consider refactoring gather_address_lists so this isn't duplicated from there
```
jennifer-richards, line 373: Seems worthwhile - it's just separating the "to" and "cc" gathering, I think? If that's all, it looks minor enough it could be included in this PR.

ietf/doc/views_ballot.py
```diff
@@ -301,21 +322,98 @@ def err(code, text):
...
+        doc = Document.objects.filter(name=post_data.get("docname")).first()
+        if doc is None:
+            errors.append("No document found matching docname")
+    if len(errors) > 0:
+        response = {
+            "success": False,
+            "errors": errors,
+        }
+    else:
+        wanted = dict()  # consider named tuple instead
```
jennifer-richards, line 364: Don't object to this, but I think a dict is fine

jennifer-richards (approved):
Some minor refactoring / additional validation is suggested inline. No objection to accepting this as-is and cleaning up in a future PR, though.

rjsparks:
I'd like to move this forward as is then - we can split _some_ of the improvements into separate PRs, but the bigger energy should go into replacing the UX for balloting with Vue/Nuxt.

rjsparks:
I'm going to retarget this to a feature branch and deploy that so the IESG can preview the change to their workflow.

PR #9677, opened 2025-10-07
feat: add a sort filter for django for the lisdexylic of us (like me)

elear:
# WARNING

**Eliot is playing with code he may not understand.**

Not quite sure how to test this in a sandbox, but happy to learn.

# Intent

The intent is to sort he list of RFCs that appear in the "obsoleted by" and "updated by" lists.  This makes it easier to spot RFCs in long lists, like that of RFC 2026, for instance.

# Method

Register a new function "sort" and then invoke it in the template.

### DIFF
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -129,8 +129,8 @@ def bracketpos(pos,posslug):
         return "[   ]"
 
 @register.filter(name='sort')
-def sortlist(li)
-    return li.sort()
+def sortlist(value):
+    return value.sort()
 
 register.filter('fill', fill)
 
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -130,7 +130,7 @@ def bracketpos(pos,posslug):
 
 @register.filter(name='sort')
 def sortlist(value):
-    return value.sort()
+    return sorted(value)
 
 register.filter('fill', fill)
 
--- a/ietf/doc/templatetags/ietf_filters.py
+++ b/ietf/doc/templatetags/ietf_filters.py
@@ -128,6 +128,10 @@ def bracketpos(pos,posslug):
     else:
         return "[   ]"
 
+@register.filter(name='sort')
+def sortlist(li)
+    return li.sort()
+
 register.filter('fill', fill)
 
 @register.filter
--- a/ietf/templates/doc/search/status_columns.html
+++ b/ietf/templates/doc/search/status_columns.html
@@ -87,11 +87,11 @@
         {{ doc.std_level|safe }} RFC
         {% if doc.obsoleted_by_list %}
             <br>
-            <span class="text-body-secondary">Obsoleted by {{ doc.obsoleted_by_list|join:", "|urlize_ietf_docs }}</span>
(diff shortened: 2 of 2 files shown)


ietf/doc/templatetags/ietf_filters.py
```diff
@@ -128,6 +128,10 @@ def bracketpos(pos,posslug):
     else:
         return "[   ]"
 
+@register.filter(name='sort')
+def sortlist(value):
+    return value.sort()
```
jennifer-richards:
```suggestion
    return sorted(value)
```
(`list.sorted()` returns None and does an in-place sort)

elear:
@jennifer-richards I also don't know if this is the only place that a change needs to be made.

rjsparks:
There's more that would need to happen I think. The comparisons need to operate on rfc _numbers_. The sort is probably more easily done in the view rather than the template, and the work needs to affect both this view and the /doc/html view.

elear:
@rjsparks there is fill_in_document_table_attributes in doc/utils_search.py. That seems to be where these values get filled in.  One could add a second loop after this code:

```
    for rel in xed_by:
        d = doc_dict[rel.target.id]
        if rel.relationship_id == "obs":
            d.obsoleted_by_list.append(rel.source)
        elif rel.relationship_id == "updates":
            d.updated_by_list.append(rel.source)
```

Is that what you are referring to?

rjsparks:
Yes, and it's only a couple of sort statements that need to be added, and I think we'll need to pass an explicit definition of the sort key to the sorting function (to pull the rfc number out of the doc object as an integer).

I'd also look for a similar place that gathers updates and obsoletes.

elear:
If that doc list is strings, of "rfc1234", then a lexical sort should be fine, right?

It seems to me like we are looking at something like this:
```
for d in doc_dict.keys():
        doc_dict[d].obsoleted_by_list.sort() # lexical
        doc_dict[d].updated_by_list.sort()
```

This presumes that doc_dict  doesn't include all RFCs/drafts.  Otherwise we have to go through the same game as above.

jennifer-richards:
> If that doc list is strings, of "rfc1234", then a lexical sort should be fine, right?

The rfc numbers don't have leading zeros, so lexical sort won't match numerical sort.

rjsparks:
Right - Think of rfc20.

But I may have reacted to where you were pointing in the code incorrectly - are you trying to solve the sorting on pages like https://datatracker.ietf.org/doc/rfc3261/? If so, then the code you're looking at in utils_search isn't relevant. That's used for pages like  https://datatracker.ietf.org/group/sip/documents/ (which looks sorted to me already).

If it's the first kind of link you're after, you need to be looking at this view: https://github.com/ietf-tools/datatracker/blob/24101bb8ca85cfb3a5c47d7f9ed283cc6fb5bc0e/ietf/doc/views_doc.py#L197, and the fix is going to be to either pull the calls in the construction of the render dict at the end forward and sort them, or maybe sort in the interesting_relations* utilities. See https://github.com/ietf-tools/datatracker/blob/24101bb8ca85cfb3a5c47d7f9ed283cc6fb5bc0e/ietf/doc/views_doc.py#L338-L372

rjsparks:
Eliot - do you want to continue pushing on this to drive learning, or do you want us to pick it up?

elear:
Hi Robert- I won't have time to dink this until the IETF.  If you want to move faster, please feel free to do so.

PR #9678, opened 2025-10-07
fix: don't show inactive constraints label if there are none

rpcross:


rjsparks:
Please rework a new PR based off a clone of feat/sreq instead of the original rpcross:sreq2 branch.

PR #9702, opened 2025-10-10
feat: allow editing call for adoption messages

rjsparks:


rjsparks:
Still WIP - several other workflows need intervention.

jennifer-richards (approved):

jennifer-richards:
(approved what's here if you want to do the additional work in a separate PR, but feel free to continue on this one if you prefer)

ietf/doc/templatetags/ietf_filters.py
```diff
@@ -1022,7 +1022,7 @@ def is_in_stream(doc):
 def can_issue_ietf_call_for_adoption(doc):
     return all(
         [
-            doc.stream_id in ["ietf", None],
```
jennifer-richards: Isn't the `doc.stream_id is None` case needed to allow calls for adopting individual submissions?

rjsparks: that's why it had wandered in and back out of this code - but no, it doesn't. The workflow won't go _through_ the view that uses this filter when working with an individual document. The filter's name may need updating to not be an attractive nuisance. We could change it to `can_issue_ietf_call_for_adoption_of_doc_already_in_ietf_stream` but...

rjsparks: This is why I compared these workflows to a moog patch panel earlier - This is a bandaid on a bandaid to work around the original stream state editing view doing so much of its work by side-effect.

jennifer-richards (approved):

PR #9728, opened 2025-10-14
feat: materials blob retrieval API

jennifer-richards:
This adds the blob retrieval API which, given a blob spec (bucket / name), returns the blob contents. If the blob does not exist, looks for a meeting-related Document of the expected type, creates a `StoredObject` and blob, and returns the contents.

The endpoint is `/api/meeting/blob/<bucket>/<name>` with no trailing slash. The values should be the ones returned from `/api/meeting/<num>/materials/<name>/resolve/`.

Known limitations:
* Does not create a `StoredObject` if the blob is in the blobdb but no `StoredObject` exists. This is an unexpected corner case.
* Does not yet render markdown into HTML - that'll be another PR
* Content type determination is simplistic
* no tests :-(


hulfo6:
هذه حالة استثنائية غير متوقعة.
لم يتم تحويل Markdown إلى HTML بعد - سيكون هذا طلب سحب آخر
تحديد نوع المحتوى أمر بسيط
لا اختبارات :-(

PR #9760, opened 2025-10-20
feat: robustness + date range for resolve materials task

jennifer-richards:
Limits resolution to meeting materials types (specifically to exclude charters, which get picked up via `session__meeting` filters) and sometimes have a revision like `00-00` that breaks assumptions in the resolver.

Logs failure to resolve a meeting if something goes wrong but continues. Outputs the date of each meeting as it's processed to assist in re-running if the task does fail.

Adds `meetings_until` task parameter in addition to `meetings_since` so an arbitrary range can be processed.

### DIFF
--- a/ietf/meeting/tasks.py
+++ b/ietf/meeting/tasks.py
@@ -79,8 +79,11 @@ def resolve_meeting_materials_task(*, meetings=None, meetings_since=None):
         if meetings_since is not None:
             log.log("Ignoring meetings_since because specific meetings were requested.")
         meetings = Meeting.objects.filter(number__in=meetings)
-    for meeting in meetings:
-        log.log(f"Resolving materials for {meeting.type_id} meeting {meeting.number}...")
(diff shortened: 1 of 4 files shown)


ietf/meeting/tasks.py
```diff
@@ -66,21 +66,57 @@ def fetch_meeting_attendance_task():
...
+    """
+    # IETF-1 = 1986-01-16
+    EARLIEST_MEETING_DATE = datetime.datetime(1986, 1, 1)
+    if meetings_since == "zero":
+        meetings_since = EARLIEST_MEETING_DATE
+    elif meetings_since is not None:
         meetings_since = datetime.datetime.fromisoformat(meetings_since)
+
+    if meetings_until is not None:
+        meetings_until = datetime.datetime.fromisoformat(meetings_until)
```
rjsparks: consider catching if a bogus string comes in.

rjsparks (approved):

PR #9763, opened 2025-10-20
feat: tag subseries API endpoints for purple

rudimatz:
goal: make subseries API endpoints available to purple client (in addition to red), so that xfer script can import subseries info into purple


jennifer-richards (approved):
Looks fine - did you confirm that this leads to the call appearing in both API clients in our workflow?

PR #9836, opened 2025-11-01
fix: AD order in IESG ballots (and other places) does not handle UTF-8 collation

### LINKED ISSUE #9483: AD order in IESG ballots does not handle UTF-8 collation
### Describe the issue

When listing the ADs' ballots for an I-D under IESG evaluation, the alphabetical sort does not take the UTF-8 correct collation, e.g., "É" should be sorted as a "E".

See https://datatracker.ietf.org/doc/draft-ietf-emu-eap-arpa/ballot/ where "Éric" appears after "Paul".

Mostly cosmetic ;-)

### Code of Conduct

evyncke:
Fix #9483 

Using a new ietf/utils/unicodenormalize.py to basically transform "Éric" into "Eric" and using this ancillary function for ballots, WG AD dashboard, and milestones under review.


evyncke:
Working on fixing ietf.iesg.tests.IESGTests.test_working_groups as the test was not expecting a sorted table ;-)

jennifer-richards (approved):

rjsparks:
This has the same issue interacting with tablesorter - clicking on the tables won't get the same sort result as this PR passes in now.

evyncke:
> This has the same issue interacting with tablesorter - clicking on the tables won't get the same sort result as this PR passes in now.

AFAIK, this was taken into consideration and tested with the current IESG for "..., Éric, Erik, ..." whose order is unchanged between Python generated first list and then after applying JS lists.js sorting.

Issue #9838, opened 2025-11-01
Tables in IESG dashboard are not sorted


evyncke:
### Description

The tables in /doc/ad are not sorted, AD names are listed in a random order.

Suggest to sort them by alphabetical order.

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

evyncke:
This issue can be closed as #9839 is merged and in production.

PR #9839, opened 2025-11-01
feat: Tables in IESG dashboard are not sorted

evyncke:
Tables in IESG dashboard are not sorted, fixing #9838 

Note: this PR also contains the new ietf/utils/unicodenormalize.py as in #9836


jennifer-richards (approved):

rjsparks:
There's a missed consideration here - the tables use tablesorter (the arrows in the column headings), and that's in javascript and won't have this new python function available to it.

In other words, the first time someone hits the "sort this column" they're likely to get a different order than was first rendered.

evyncke:
> There's a missed consideration here - the tables use tablesorter (the arrows in the column headings), and that's in javascript and won't have this new python function available to it.

Actually, this was considered and tested ;) at least for the current IESG. Python new code put the order as "..., Éric, Erik, ..." and using the lists.js on the arrows keeps the order "..., Éric, Erik, ..." (of course when sorting from A to Z).

PR #9841, opened 2025-11-01
feat: Add tab for IESG view of working groups

### LINKED ISSUE #9730: Add IESG WG dashboard to the DT menu
### Description

Finding https://datatracker.ietf.org/iesg/working-groups/ is rather difficult...

Unsure whether it is visible only for the IESG or is public (preferred) but it should appear in the main menu bar (either under Group/IESG -- preferred -- or in Documents/IESG -- a little weird --)

Having a link in the AD dashboard would be a plus

### Code of Conduct

russhousley:
fixes #9730 

Add a tab for Working Groups to http://datatracker.ietf.org/group/iesg/

### DIFF
--- a/ietf/group/utils.py
+++ b/ietf/group/utils.py
@@ -237,9 +237,10 @@ def construct_group_menu_context(request, group, selected, group_type, others):
         entries.append(("Review requests", urlreverse(ietf.group.views.review_requests, kwargs=kwargs)))
         entries.append(("Reviewers", urlreverse(ietf.group.views.reviewer_overview, kwargs=kwargs)))
         entries.append(("Reviews History", urlreverse(ietf.group.views.review_requests_history, kwargs=kwargs)))
-
     if group.features.has_meetings:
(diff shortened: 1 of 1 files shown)


ietf/group/utils.py
```diff
@@ -237,9 +237,10 @@ def construct_group_menu_context(request, group, selected, group_type, others):
         entries.append(("Review requests", urlreverse(ietf.group.views.review_requests, kwargs=kwargs)))
         entries.append(("Reviewers", urlreverse(ietf.group.views.reviewer_overview, kwargs=kwargs)))
         entries.append(("Reviews History", urlreverse(ietf.group.views.review_requests_history, kwargs=kwargs)))
-
     if group.features.has_meetings:
         entries.append(("Meetings", urlreverse("ietf.group.views.meetings", kwargs=kwargs)))
+    if group.acronym in ["iab", "iesg"]:
```
jennifer-richards: Should only be "iesg" ?

jennifer-richards (approved):

PR #9845, opened 2025-11-01
refactor: use ical for upcoming meetings

rudimatz:


ietf/meeting/views.py
```diff
@@ -4417,22 +4417,74 @@ def upcoming_ical(request):
...
+            description_lines.append(f"{item.timeslot.name}")
+        if item.session.agenda_note:
+            description_lines.append(f"Note: {item.session.agenda_note}")
+        
+        for material in item.session.materials.all():
+            description_lines.append(f"{material.type}")
+            if material.type.name != "Agenda":
+                description_lines.append(f"{material.title}")
+
+            description_lines.append(f"{material.get_href()}")
```
jennifer-richards: I think this section is causing problematic newline changes. The original looked like:
```
DESCRIPTION:
 \nAgenda:
 https://[...]\n
 \nNotAgenda
 This is the Title:
 https://[...]\n
 Remote instructions: [...]\n
```
which renders in the calendar like
> 
> Agenda: https://[...]
>
> NotAgenda This Is the Title: https://[...]
> Remote instructions: [...]
>

I think the new code is going to come out like
> Agenda:
> https://[...]
> NotAgenda
> This is the Title:
> https://[...]
> Remote instructions: [...]

jennifer-richards (changes requested):
Needs some attention to the newlines. Don't necessarily need to exactly replicate the original, but I think some of the changes here will negatively impact readability of the description in the calendar entry.

ietf/meeting/views.py
rudimatz: ok changed it to appear in one line

jennifer-richards (approved):

PR #9849, opened 2025-11-01
refactor: use ical for important-dates

rudimatz:


rjsparks (dismissed):

ietf/meeting/views.py
```diff
@@ -4432,6 +4432,73 @@ def upcoming_ical(request):
...
+                    "Register here: https://www.ietf.org/how/meetings/register/")
+            if important_date.name.slug == 'opensched':
+                description_lines.append("To request a Working Group session, use the "
+                                         "IETF Meeting Session Request Tool:")
+                description_lines.append(f"{request.scheme}://{request.get_host()}"
+                    f"{reverse('ietf.meeting.views_session_request.list_view')}")
+                description_lines.append("If you are working on a BOF request, it is "
+                                         "highly recommended to tell the IESG")
+                description_lines.append("now by sending an email to iesg@ietf.org")
+                description_lines.append("to get advance help with the request.")
```
jennifer-richards: This was a single line in the template - it's now broken in two. Is that intentional?

jennifer-richards (changes requested):
See inline

jennifer-richards (approved):

PR #9850, opened 2025-11-01
feat: add option to email users about duplicate accounts. Fixes #8174.

rpcross:


ietf/person/views.py
```diff
@@ -144,3 +146,71 @@ def merge(request):
...
     })
+
+
+@role_required("Secretariat")
+def send_merge_request(request):
+    if request.method == 'GET':
+        source_id = request.GET.get('source')
+        target_id = request.GET.get('target')
+        source = Person.objects.get(id=source_id)
+        target = Person.objects.get(id=target_id)
```
jennifer-richards: Suggest using a form here for validation

ietf/person/forms.py
```diff
@@ -21,3 +21,19 @@ def get_person(self, pk):
...
+class MergeRequestForm(forms.Form):
+    to = forms.CharField()
+    frm = forms.CharField()
+    reply_to = forms.CharField()
+    subject = forms.CharField()
+    body = forms.CharField(widget=forms.Textarea)
+
+    def clean_to(self):
+        addresses = self.cleaned_data['to']
+        return addresses.split(',')
```
jennifer-richards: Not ideal to be "cleaning" a CharField into a list in the `clean_*` method. That sort of transformation needs to be in a `to_python()` method of a custom field class. We have `ietf.utils.fields.MultiEmailField` that might be close to what you need, though it is probably in need of some clean up.

If that's not doing the right things, it'd be worth making your own field class here.

ietf/templates/person/merge.html
```diff
@@ -1,5 +1,5 @@
+{# Copyright The IETF Trust 2025, All Rights Reserved #}
```
jennifer-richards: Usual practice is to preserve the original year (i.e., `2015-2025`)

ietf/templates/person/merge.html
```diff
@@ -37,16 +37,27 @@ <h1>Merge Person Records</h1>
...
-        </button>
+            </button>
+        {% else %}
+            <button type="submit" class="btn btn-warning" title="Get Person information">
+                Get Info
+            </button>
+        {% endif %}
+        {% if method == 'post' %}
+            <a class="btn btn-primary"
+               href="{% url 'ietf.person.views.send_merge_request' %}?source={{ source.pk }}&amp;target={{ target.pk }}"
```
jennifer-richards: Does the \&amp; entity do the right thing here? I think it should be a literal \& since it's in a URL

jennifer-richards (changes requested):
I have a few comments inline, but the merge persons view needs a refactoring. It's currently dangerous enough that IMO it needs to be part of this PR.

The workflow currently looks like:

1. GET the /person/merge/ page
2. fill in the from/to fields with person IDs
3. Submit with a GET, returning a page with the preview of the persons.
4. Review and submit with a POST

However, in step 3, the from/to fields are still present and editable. Editing them does not update the preview, and submitting will merge persons based on the edited values without much warning. Obviously "don't do that," but it's an opportunity for a serious error that would be difficult to roll back.

This should instead be two views, one that is _only_ selecting the persons to merge and another that is _only_ previewing and submitting. It can have the current "swap" button and the new "send email" button in addition to the "Merge" button, but shouldn't allow editing the IDs to be affected. Probably it should add a "go back" type button to return to the original view.

LMK if that doesn't make sense

ietf/templates/person/merge.html
rpcross: In this case there wasn't a prior copyright text so I just used current year. Are you saying I should determine the first year the file was in place and use that in text?  ie.  [start]-[current]

ietf/person/forms.py
rpcross: It looks like this would work, except for the fact that validate_email doesn't handle emails in the form "IETF Secretariat <ietf-secretariat-reply@ietf.org>" which is the settings default from address. I could just use "ietf-secretariat-reply@ietf.org". What clean up are you referring to?

ietf/person/views.py
rpcross: Fixed in commit bf52f632cd49f585df9da8962f1d61ec29c7d3d3

ietf/templates/person/merge.html
jennifer-richards: Ah, hadn't noticed it was entirely new. If you can easily figure it out it'd be nice but otherwise 🤷

ietf/person/forms.py
jennifer-richards: Didn't look hard at cleanup, but it's written in old style / dialect that caught my eye (specifically, not using `super().validate(...)`). Not asking you to refactor it, just a comment.

I agree it doesn't do what you need. The main reason to point at it was just the structure it has. I.e., rather than doing data conversion in a `clean()` method, it'd be preferable to do something like

```python
class MutliNameAddrField(forms.Field):
    def to_python(self, value):
        """Split on commas into a list of maybe-valid name-addr strings, or raise ValidationError if it can't do that"""
        # ...
    def validate(self, value):
        """Do more detailed validation on the value that came out of to_python"""
        # ...
```

jennifer-richards: https://docs.djangoproject.com/en/5.2/ref/forms/validation/#form-and-field-validation is a useful reference that it always takes me too long to find

ietf/templates/person/merge.html
rpcross: Fixed in commit 5e69a36a4e4655ba6994400a0cb32c97e70a6ea5

ietf/person/forms.py
rpcross: Ok. I mixed this up. For the send_mail_text() "to" and "reply-to", which expect lists, the existing MultiEmailField works fine. For the From address, which gets a name-addr style email address from settings.DEFAULT_FROM_EMAIL, I created a custom field NameAddrEmailField.

See commit 421252641d2eab5837ed5e321f713b4e18969437

jennifer-richards (approved):

PR #9851, opened 2025-11-01
fix: add most recent ballot hint to History tab. 

russhousley:
Fixes #9696.
Add a hint to the History tab for balloters that tells the revision of the document associated with the most recent ballot.

### DIFF
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -697,7 +697,7 @@ def last_ballot_doc_revision(doc, person):
     balloters = get_active_balloters(ballot.ballot_type)
     if person not in balloters:
         return None
-    position_queryset = BallotPositionDocEvent.objects.filter(type="changed_ballot_position", balloter=person, ballot=ballot)
+    position_queryset = BallotPositionDocEvent.objects.filter(type="changed_ballot_position", balloter=person, ballot=ballot).order_by("-time")
     if not position_queryset.exists():
         return None
     ballot_time = position_queryset.first().time
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -692,7 +692,7 @@ def last_ballot_doc_revision(doc, person):
     """ Return the document revision for the most recent ballot position
         by the provided user. """
     ballot = doc.active_ballot()
-    if ballot == None or person == None:
+    if ballot is None or person is None:
         return None    
     balloters = get_active_balloters(ballot.ballot_type)
     if person not in balloters:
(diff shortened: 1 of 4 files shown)


ietf/doc/utils.py
```diff
@@ -687,6 +688,22 @@ def nice_consensus(consensus):
...
     return mapping[consensus]
 
+def last_ballot_doc_revision(doc, person):
+    """ Return the document revision for the most recent ballot position
+        by the provided user. """
+    ballot = doc.active_ballot()
+    if ballot == None or person == None:
+        return None    
+    balloters = get_active_balloters(ballot.ballot_type)
+    if not person in balloters:
```
rjsparks:
```suggestion
    if person not in balloters:
```

ietf/doc/utils.py
```diff
@@ -687,6 +688,22 @@ def nice_consensus(consensus):
...
+        return None
+    position_queryset = BallotPositionDocEvent.objects.filter(type="changed_ballot_position", balloter=person, ballot=ballot)
```
rjsparks:
```suggestion
    position_queryset = BallotPositionDocEvent.objects.filter(type="changed_ballot_position", balloter=person, ballot=ballot).order_by("-time")
```

rjsparks: This makes this query continue to work if the default ordering for the model changes in the future.

ietf/templates/doc/document_history.html
```diff
@@ -20,6 +20,11 @@
         <h2 class="my-3">Revision differences</h2>
         {% include "doc/document_history_form.html" with doc=doc diff_revisions=diff_revisions action=rfcdiff_base_url snapshot=snapshot only %}
     {% endif %}
+    {% if ballot_doc_rev %}
+        <p class="alert alert-info my-3">
+            Your most recent ballot position was provided on revision {{ ballot_doc_rev }}.
```
rjsparks: nit: it would be more precise to say "was provided when the document was at version"

rjsparks (changes requested):

ietf/templates/doc/document_history.html
rjsparks:
```suggestion
            Your most recent ballot position was provided when the document was at version {{ ballot_doc_rev }}.
```

ietf/doc/utils.py
```diff
@@ -687,6 +688,22 @@ def nice_consensus(consensus):
         }
     return mapping[consensus]
 
+def last_ballot_doc_revision(doc, person):
+    """ Return the document revision for the most recent ballot position
+        by the provided user. """
+    ballot = doc.active_ballot()
+    if ballot == None or person == None:
```
jennifer-richards:
```suggestion
    if ballot is None or person is None:
```

jennifer-richards (commented):
Trivial nit suggested

jennifer-richards (approved):

rjsparks (approved):

PR #9854, opened 2025-11-02
fix: no autoescape for bluesheet template

jennifer-richards:
@kesara


jennifer-richards:
Closing, got mixed with unrelated work

PR #9856, opened 2025-11-02
feat: show available sdo groups to liaison coordinators and managers

rjsparks:


ietf/liaisons/views.py
```diff
@@ -525,3 +526,17 @@ def liaison_resend(request, object_id):
     messages.success(request,'Liaison Statement resent')
     return redirect('ietf.liaisons.views.liaison_list')
 
+
+@role_required("Secretariat", "IAB", "Liaison Coordinator", "Liaison Manager")
+def list_other_sdo(request):
+    def _sdo_order_key(obj:Group)-> Tuple[str,str]:
```
jennifer-richards: nit: better just to do `-> tuple[str, str]` (lower-case `t`, no import of `Tuple`)

ietf/templates/base/menu.html
```diff
@@ -402,6 +402,14 @@
             Liaison statements
         </a>
     </li>
+    {% if user|has_role:"Secretariat,IAB,Liaison Manager,Liaison Coordinator" %}
+    <li>
+        <a class="dropdown-item {% if flavor != 'top' %}text-wrap{% endif %}"
```
jennifer-richards: Ultra-tiny nit: avoid trailing space when `flavor == 'top'`

```suggestion
        <a class="dropdown-item{% if flavor != 'top' %} text-wrap{% endif %}"
```

jennifer-richards (changes requested):
Minor cleanup probably worth doing to maintain best practices: remove the `import Tuple` and use `tuple[str, str]`. Otherwise looks good.

ietf/templates/base/menu.html
rjsparks: this had been keeping the same format as the other dozen versions of it around it, but I added a commit that changed them all (expanding the scope of the PR a bit).

jennifer-richards (approved):

Issue #9822, opened 2025-10-29
Provide a summary of nominee known participation to the nomcom members


rjsparks:
### Description

cc: @MikeBishop

Provide a page summarizing known participation of accepted nominees to the nomcom members.
This page should show
* past meetings attended in the last 5 years
* drafts/rfcs
* roles held

The page should prominently call out the limitations of the datatracker to speak to things such as past roles.

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

evyncke:
Wondering why this page would be limited to NomCom members and not to everyone ?

I.e., adding this information to the plain IETF profile page.

MikeBishop:
I don't know that there's a compelling reason one way versus the other. It was requested for NomCom because we wanted the information; we had nominees who appeared to have very little IETF footprint, and we wanted to verify that impression / check that we weren't missing others in the same situation.

The information it's drawing on is public, so if there's a desire to expose it more generally in the future, that's probably fine.

rjsparks:
Please note the discussion about privacy in the PR.

PR #9857, opened 2025-11-02
feat: Provide a summary of nominee known participation to the nomcom members

evyncke:
Address #9822 

I did not dare to modify the /person/john.doe@example.com hence a new URL /person/john.doe@example.com/details

I.e., some pages in the NomCom part should probably be modified to point to this new /person/john.doe@example.com/details

### DIFF
--- a/ietf/group/templatetags/group_filters.py
+++ b/ietf/group/templatetags/group_filters.py
@@ -15,6 +15,10 @@ def active_roles(queryset):
     return queryset.filter(group__state_id__in=['active', 'bof']).exclude(group__acronym='secretariat')
     
 @register.filter
+def inactive_roles(queryset):
+    return queryset.filter(group__state_id__in=['conclude'])
+
+@register.filter
 def active_nomcoms(user):
     if not (user and hasattr(user, "is_authenticated") and user.is_authenticated):
         return []
--- a/ietf/nomcom/tests.py
+++ b/ietf/nomcom/tests.py
@@ -143,7 +143,7 @@ class NomcomViewsTest(TestCase):
 
         # Verify that nominee table has links to person and feedback pages
         nom_pos = self.create_nominee('accepted', COMMUNITY_USER, 'APP')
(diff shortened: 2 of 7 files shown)


evyncke:
Updated at least one page of the NomCom private site to link to this detailed page

evyncke:
As this issue was raised by @MikeBishop for this NomCom this might be interesting to run / deploy it on sandbox quickly once all the checks are completed successfully.

rjsparks:
I've given the nomcom a custom report for now.

There's a lot that's good in this PR but it should move to /person/ and it _can't_ deal with past roles correcty - the data the datatracker has about them is incomplete at best and misleading at worst. See the tools-discuss/chairs discussion about how poorly the datatracker can talk about the past chairs of a working group at the moment. We will need a data collection effort before we start reporting on anything more than what roles a person hold right now.

evyncke:
I was not aware of the report, so, nothing urgent anymore.

I agree with the past roles.

I can move the last 5 years of meetings in the normal /people/ page if you prefer and remove the people/details one

rjsparks:
> I can move the last 5 years of meetings in the normal /people/ page if you prefer

Just checking: Is this really something we want to summarize for any anonymous person to quickly see about any other person? The data _is_ public already, but it takes work to compile the summary. I would like to talk a bit more about the privacy implications before we add it here.

In the short term, a single page that concentrates this for nomcom members to see the values for all the nominees is something I expect nomcoms to ask for (rather than clicking through to each person's person page).

richsalz:
I share the privacy concern. Leave it nomcom only.

evyncke:
Up to the tools-team of course, but this smells like security by obfuscation as the data is easy to grab via the datatracker API anyway.

richsalz:
> Up to the tools-team of course, but this smells like security by obfuscation as the data is easy to grab via the datatracker API anyway.

Not sure it's *easy* but yes, it's adding a speed bump in favor of a nod to privacy.

PR #9858, opened 2025-11-02
fix: no autoescape for bluesheet template

jennifer-richards:
Replaces #9854 (minus the accidentally included work)

@kesara


kesara (approved):
LGTM
May be add a tiny test to confirm the behaviour?

rjsparks (approved):

PR #9886, opened 2025-11-05
feat: add OpenTelemetry integration

NGPixel:


dev/build/gunicorn.conf.py
```diff
@@ -1,5 +1,17 @@
 # Copyright The IETF Trust 2024, All Rights Reserved
```
jennifer-richards:
```suggestion
# Copyright The IETF Trust 2024-2025, All Rights Reserved
```

dev/build/gunicorn.conf.py
```diff
@@ -119,3 +131,25 @@ def post_request(worker, req, environ, resp):
...
+
+def post_fork(server, worker):
+    server.log.info("Worker spawned (pid: %s)", worker.pid)
+
+    resource = Resource.create(attributes={
+        "service.name": "datatracker",
+        "service.version": ietf.__version__,
+        "service.instance.id": worker.pid,
+        "service.namespace": "datatracker",
+        "deployment.environment.name": os.environ.get('DATATRACKER_SERVICE_ENV', 'dev')
```
jennifer-richards:
```suggestion
        "deployment.environment.name": os.environ.get("DATATRACKER_SERVICE_ENV", "dev")
```

dev/build/gunicorn.conf.py
```diff
@@ -119,3 +131,25 @@ def post_request(worker, req, environ, resp):
...
+    })
+
+    trace.set_tracer_provider(TracerProvider(resource=resource))
+    otlp_exporter = OTLPSpanExporter(endpoint='https://heimdall-otlp.ietf.org/v1/traces')
```
jennifer-richards:
```suggestion
    otlp_exporter = OTLPSpanExporter(endpoint="https://heimdall-otlp.ietf.org/v1/traces")
```

ietf/context_processors.py
```diff
@@ -51,3 +52,8 @@ def timezone_now(request):
     return {
         'timezone_now': timezone.now(),
     }
+
+def traceparent_id(request):
+    context_extras = {}
+    inject(context_extras)
+    return context_extras
```
jennifer-richards: What would you think about...
```suggestion
    return {"opentelemetry": context_extras}
```
and then the template would refer to `opentelemetry.traceparent`? Might help avoid collisions and make clearer where the values came from when looking at the context.

jennifer-richards (commented):
Very important style suggestions, plus a thought about the context injection...

ietf/context_processors.py
NGPixel: Wouldn't that create potential collisions with the "opentelemetry" own packages? We could change it to "otel" though.

jennifer-richards: That works

dev/build/gunicorn.conf.py
```diff
@@ -119,3 +131,25 @@ def post_request(worker, req, environ, resp):
     in_flight = in_flight_by_pid.get(worker.pid, [])
     if request_description in in_flight:
         in_flight.remove(request_description)
+
+def post_fork(server, worker):
```
jennifer-richards, line 135: 🍴

jennifer-richards (approved):

egyptianlove4444-dotcom (commented):
(spam removed)

PR #9928, opened 2025-11-13
ci: automate release merge

rjsparks:


jennifer-richards:
Is the "do not delete branches" step expected to be permanent, or is that a while-debugging/testing feature that will change back?

NGPixel:
The delete-branch is actually required as it's a temporary branch that we don't want to keep around.

PR #9940, opened 2025-11-14
feat: basic RfcAuthor API

jennifer-richards:
Adds basic RfcAuthor management endpoints. Serializer naming is a bit less than ideal because `RfcAuthorSerializer` is already in use for the red API. I'll clean that up when I add RfcAuthor support to that API and then rename the serializer.

This is similar to purple's `RfcAuthor` API, but does not provide the `set_order()` action. We could add that, but I think it's unlikely to be used on this interface. Instead, delete all the existing authors for an RFC and add the correct set from scratch.

### DIFF
--- a/ietf/api/serializers_rpc.py
+++ b/ietf/api/serializers_rpc.py
@@ -7,7 +7,7 @@ from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
-from ietf.doc.models import DocumentAuthor, Document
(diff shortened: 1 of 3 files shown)


rudimatz (approved):

ietf/api/serializers_rpc.py
```diff
@@ -86,7 +86,7 @@ class Meta:
 
     def get_plain_name(self, document_author: DocumentAuthor) -> str:
         return document_author.person.plain_name()
-
+
```
rjsparks: nit: stray whitespace

rjsparks (approved):

PR #9950, opened 2025-11-15
style: update text color in timeline bar for better visibility in dark mode

### LINKED ISSUE #9905: darkmode timeline colors are hard to see
### Description

Image
The colors could be less glaring and the numbers in a color that make them more visible.

### Code of Conduct

Sreyanth:
fix #9905 - show dark color text on timeline bar in both light & dark modes

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -319,8 +319,9 @@ th,
         dominant-baseline: central;
     }
 
+    $timeline-bar-text-color: $gray-900;
     .bar text {
-        fill: var(--bs-body-color);
+        fill: $timeline-bar-text-color;
         dominant-baseline: central;
         pointer-events: none;
     }


jennifer-richards:
@larseggert IIRC an aim of the bootstrap5 reskin was to avoid custom CSS as much as possible. Thoughts about adding the custom color here rather than finding something in the `--bs-*` variables (or some other fix)?

rjsparks:
These changes, so far, only affect the text color:
image
The colors of the bars themselves should also be reconsidered in this mode - they aren't painful in light-mode, but they are eye-stabbing in dark-mode.

The lightmode alternates between bootstrap's warning and info colors if I read the scss correctly, but I wonder if the way we're collecting them causes them to not follow changes to darkmode correctly. Re-appealing to @larseggert for a look.

Sreyanth:
Thanks for the comments! For this PR I intentionally kept the scope small and only updated the text colors (so the labels are at least readable), reusing the existing variables to keep things consistent with the rest of the dark-mode theme.

I agree that the timeline bar colors themselves could use a broader rethink for dark-mode. I’m happy to dig into that if that’s a direction you’d like to pursue. Looking forward to @larseggert's guidance.

PR #9964, opened 2025-11-18
fix: simplify call for adoption workflow

rjsparks:


jennifer-richards (approved):

ietf/doc/views_draft.py
```diff
@@ -2075,27 +2075,44 @@ def clean(self):
...
+    group = Group.objects.filter(acronym=acronym, type_id="wg").first()
+    if any(
+        [
+            doc.type_id != "draft",
+            group is None,
+            not is_doc_ietf_adoptable(doc),
+            doc.group.acronym == "none" and acronym is None,
+            doc.group.acronym == "none" and not can_adopt_draft(request.user, doc),
+            doc.group.acronym != "none"
+            and not is_authorized_in_doc_stream(request.user, doc),
```
jennifer-richards: The doc may not be in the "ietf" stream at this point - is there a check that the user is authorized to issue the adoption in that stream? It looks to me like it checks its current stream, but that might be a different stream. Am I missing something?

jennifer-richards (changes requested):
oops - I think I have a question about authorization.  See inline

ietf/doc/views_draft.py
rjsparks: If the document is not in any stream, it will be attached to the group with acronym 'none' and `can_adopt_draft` does the checking you are looking for.

If the document is in some stream that is not the ietf stream, the group will not be of type `wg` and the query at L2080 will return None, so L2084 will reject access.

If the document is in the ietf stream, the awkwardly named `is_authorized_in_doc_stream` will perform the check you are looking for.

rjsparks: I restructured the condition to hopefully make this easier to follow.

The differences between `can_adopt_draft` and `is_authorized_in_draft_stream` could probably use further refactoring, but I don't think that's a job for this PR.

jennifer-richards (approved):
thx for the explanation, refactor makes the logic a lot clearer

Issue #8739, opened 2025-03-26
Chat Log in Dark Mode is unreadable


dhruvdhody:
### Describe the issue

Example https://datatracker.ietf.org/doc/chatlog-122-srv6ops-202503211300/ seen below in dark mode - 

![Image](https://github.com/user-attachments/assets/55f29277-248a-4120-876f-c78ea09f0410)

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #9995, opened 2025-11-24
fix: make chatlog readable in dark mode

ghost:
Fixes #[8739](https://github.com/ietf-tools/datatracker/issues/8739)


jennifer-richards (approved):

rjsparks:
This results in
image
which is a vast improvement, but I wonder if we should be toning down the brightness (a future task - not asking for a change to this PR).

PR #10014, opened 2025-11-27
feat: use RfcAuthor for red API

jennifer-richards:
Changes the `RfcAuthor` schema to match the `RfcAuthor` model and returns data from `doc.rfcauthor_set` if that is not empty. Otherwise, it uses `doc.documentauthor_set` data. The same serialized schema is used regardless of the data source.

The new schema is below. Note that `titlepage_name` is the only guaranteed non-empty field. This is the only name field included - the old `name` field has been dropped. A new `datatracker_person_path` field is added. This is the URL  path for the author's "person page", relative to datatracker's root. If this is empty, no link should be rendered.

When representing a `DocumentAuthor`:
*  `titlepage_name` will be `person.plain_name()` (same value as the `name` field prior to this refactor)
*  `is_editor` is _always_ `False`
*  `person` is _always_ non-null

```yaml
    RfcAuthor:
      type: object
      description: Serializer for a DocumentAuthor in a response
      properties:
        titlepage_name:
          type: string
          maxLength: 128
        is_editor:
          type: boolean
        person:
          type: integer
          nullable: true
        email:
          type: string
          format: email
        affiliation:
          type: string
          description: Organization/company used by author for submission
          maxLength: 100
        country:
          type: string
          description: Country used by author for submission
          maxLength: 255
        datatracker_person_path:
          type: string
          format: uri
          description: URL for person link (relative to datatracker base URL)
      required:
      - titlepage_name
```

### DIFF
--- a/ietf/doc/serializers.py
+++ b/ietf/doc/serializers.py
@@ -10,26 +10,43 @@ from rest_framework import serializers, fields
 
 from ietf.group.serializers import GroupSerializer
 from ietf.name.serializers import StreamNameSerializer
-from .models import Document, DocumentAuthor
+from .models import Document, DocumentAuthor, RfcAuthor
 
 
 class RfcAuthorSerializer(serializers.ModelSerializer):
     """Serializer for a DocumentAuthor in a response"""
-
-    name = fields.CharField(source="person.plain_name")
-    titlepage_name = fields.CharField(default="")
-    email = fields.EmailField(source="email.address", required=False)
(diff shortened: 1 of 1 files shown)


holloway (approved):
Looks good, assuming that `get_absolute_url` does return a path and not a full url

jennifer-richards:
> Looks good, assuming that `get_absolute_url` does return a path and not a full url

Yes, it returns only the path. I don't know why it's named that way but it's Django's convention. (We supply the method so it won't change on us.)

PR #10016, opened 2025-11-27
feat: Add an option for /doc/html that renders as HTML

martinthomson:
This tries to make the HTML rendering look good, using the native bootstrap styling as much as I can be bothered to do for now.

I've retained the HTML-that-looks-like-text default.

Preview:

image

Dark:

image


martinthomson:
@larseggert, you might have some insight into whether this is complete enough to preview given your experience of how this was put together.

larseggert (approved):
LGTM in terms of the code. (Which wasn't pretty to begin with.) We should somehow remove display options ratehr than adding them, but 🤷.

Nit: Why is the metadata block not left-adjusted? Having it kinda-but-not-quite centered looks weird to me.

martinthomson:
The metadata block is left-adjusted, but I haven't spent a lot of time tweaking anything here, so some of the sizes are probably off, based on the choice of fonts and other things.  So the left padding, which needs to be long enough to handle the longest possible label in that block, is probably a little bit more generous than it needs to be.  I spotted a few minor errors like that; there will likely be more, but it might be OK as a non-default option to iterate that way.

martinthomson:
The test failures [here](https://github.com/ietf-tools/datatracker/actions/runs/19751042184/job/56594026130?pr=10016#step:6:1954) seem unrelated:
> FAIL: test_leadership_wg (ietf.group.tests.GroupLeadershipTests.test_leadership_wg)

kesara:
@martinthomson, Is your motivation behind this to add an option to display HTML similar to what `xml2rfc --html` output?

jennifer-richards (commented):
My comment about proliferation of styles on #10024 really belongs here, I guess. Another option is ok with me now, but wouldn't want the number to keep growing.

martinthomson:
@kesara, more or less.  Though it's not that simple.  The stylesheet needs to integrate with datatracker to pick up things like light/dark modes, colors, and fonts.

I acknowledge the style proliferation, but I think we have people who want non-txt-looking HTML, even if the "brand" is monospaced.

PR #10022, opened 2025-11-28
fix: Ensure more doc/html links stay in doc/html

martinthomson:
As far as I can tell, the code that was supposed to rewrite references was inoperative.  It just glued the same URL back together.  This version is effective.

The links to RFCs in the header of documents were being missed. I did not change the RFC link that appears in boilerplate.

I also update the text of the link if it is the same as the destination, to avoid surprises.

This also applies the fix unconditionally, so that the treatment is consistent regardless of what preference is chosen.

### DIFF
--- a/ietf/static/js/document_html.js
+++ b/ietf/static/js/document_html.js
@@ -87,17 +87,17 @@ document.addEventListener("DOMContentLoaded", function (event) {
     document.activeElement.blur();
 
     // keep RFC links within the datatracker
-    const rfclink = /^https?:\/\/[^\/\.]*.(?:rfc-editor|ietf).org\/.*\/rfc(\d+)$/;
-    const docbase = document.location.origin + "/doc/html/rfc";
+    const rfclink = /^https?:\/\/[^\/\.]*.(?:rfc-editor|ietf).org\/.*\/((?:rfc|std|bcp)\d+)$/;
+    const docbase = document.location.origin + "/doc/html/";
     [...document.querySelectorAll(".references dd a:not([href='']:last-of-type)"),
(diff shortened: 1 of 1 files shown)


martinthomson:
@larseggert, this is another one of your changes.  Maybe you can help tell if this makes sense.

larseggert (approved):
If it works now, it makes sense :-)

We really need unit tests for all this datatracker JS code.

ietf/static/js/document_html.js
```diff
@@ -86,6 +86,20 @@ document.addEventListener("DOMContentLoaded", function (event) {
     defpane.show();
     document.activeElement.blur();
 
+    // keep RFC links within the datatracker
+    const rfclink = /^https?:\/\/[^\/\.]*.(?:rfc-editor|ietf).org\/.*\/((?:rfc|std|bcp)\d+)$/;
```
jennifer-richards, line 90: Perhaps should limit the subdomains that are treated as RFC links here. A link to `https://notes.ietf.org/.../bcp7890` isn't an RFC link. It's also not a great example as it's not likely to appear at all, but the point is real RFC links are at a more constrained set of hosts.

martinthomson: I wasn't confident that I could enumerate the locations.  Is it just www.ietf.org and datatracker.ietf.org and www.rfc-editor.org?  The deployment architecture of this stuff (and its history) are a bit opaque to me.

jennifer-richards: I'll have a look, but imo it'd be better to start off conservatively

PR #10044, opened 2025-12-02
fix: remove all None authors, not just first

jennifer-richards:


rjsparks (approved):

ietf/doc/models.py
```diff
@@ -470,7 +470,7 @@ def authors(self):
                 log.log(
```
kesara, line 470: Remove the log message?

jennifer-richards: No - it still doesn't handle them properly, it just doesn't fail. Authors without a person are omitted from the return value.

jennifer-richards: Well.. I guess we _could_ remove it if it's being hit often enough to cause problems. Its main purpose was to be sure we didn't forget the todo on L467

kesara (approved):

PR #10084, opened 2025-12-11
feat: scripts to build and test recovering deleted objects

rjsparks:


extract_recovery.py
```diff
@@ -0,0 +1,83 @@
...
+            rows_in_progress = [line[2:]]
+        elif line.startswith("< COPY"):
+            if len(rows_in_progress) > 1:
+                tables[table_in_progress] = rows_in_progress
+            table_in_progress = line.split()[2]
+            rows_in_progress = [line[2:]]
+        else:
+            rows_in_progress.append(line[2:])
+    assert(len(TARGET_TABLES) == 38)
+    assert(set(TARGET_TABLES) == set(tables.keys()))
```
jennifer-richards: Danger! The `assert` statement _does not take parentheses_. You're ok in these uses, but it's a MASSIVE footgun if you add a message to the assert. (`assert(False, "this never fails")` doesn't fail - it's saying `assert tuple((False, "this never fails"))`, and the tuple is truthy.

rjsparks:
Closing - the PR will remain as a record of how we recovered this data.

PR #10135, opened 2025-12-18
fix: change title from 'IAB news we can use' to 'IESG Liaison News'

### LINKED ISSUE #10136: Rename "IAB news we can use" on the IESG Agenda
### Description

Currently, section 5 of the IESG agenda (https://datatracker.ietf.org/iesg/agenda/) is "IAB news we can use." Please change this to "IESG liaison news we can use."

### Code of Conduct

OR13:


jennifer-richards:
Fixes #10136 (with slight difference in text, which may or may not be acceptable)

rjsparks (approved):

PR #10223, opened 2026-01-07
feat: include tickets in RegistrationResource

jennifer-richards:
This allows filtering meeting registrations by registration type via the v1 API. Includes the ticket data in the registration response - this is verbose, but avoids having to do lots of additional lookups to fetch data. This is not hard to make terser, but I don't think the quantity of data will be prohibitively large.

@evyncke

### DIFF
--- a/ietf/meeting/resources.py
+++ b/ietf/meeting/resources.py
@@ -21,7 +21,13 @@ from ietf.meeting.models import (Meeting, ResourceAssociation, Constraint, Room,
                                  Attended,
                                  Registration, RegistrationTicket)
 
-from ietf.name.resources import MeetingTypeNameResource
+from ietf.name.resources import (
+    AttendanceTypeNameResource,
+    MeetingTypeNameResource,
+    RegistrationTicketTypeNameResource,
+)
+
+
 class MeetingResource(ModelResource):
     type             = ToOneField(MeetingTypeNameResource, 'type')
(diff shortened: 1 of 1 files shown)


rjsparks (approved):

evyncke:
Thank you both

evyncke:
And in case you wonder why I need this data, it is to build leadership, chairs, and authors expected attendance at the next IETF meeting. It was built in the COVID years so I still need to remove the COVID part in https://ehealth.vyncke.org/ietf/

Currently using a join/match (datatracker roles and registrations) based on full name but this quite error prone as opposed to datatracker ID.

rjsparks:
As being discussed on tools-discuss, this should feed from a purposed API about meeting registration rather than API/v1, but also, as this is important to the IESG, shouldn't it be built into the datatracker?

evyncke:
> As being discussed on tools-discuss, this should feed from a purposed API about meeting registration rather than API/v1, but also, as this is important to the IESG, shouldn't it be built into the datatracker?

About the completely new API, my concern is that this would take a long time.

About getting it in the datatracker (guessing that you mean something similar to my quick & dirty PHP/JS page and backend PY scripts), why not ? But, to be honest this would be useful mainly to check attendance in meetings where there is some concerns about onsite participations, i.e., rather an exception. But, I had also this idea in mind to be honest

PR #10233, opened 2026-01-08
feat: provide link to detailed submission status page for submission api

### LINKED ISSUE #6072: Submission API should provide link to detailed submission status page
### Description

The response to a submission via the API includes a URL for the status API, `status_url` in the JSON payload:
```json
{
"id": "12345",
"name": "draft-ietf-some-draft-name",
"rev": "00",
"status_url": "https://datatracker.ietf.org/api/submission/12345/status"
}
```

That API endpoint is intended for use in automation - it provides minimal details, currently just the state slug (and with a recently merged but not yet deployed change, the human-readable name of the state).

Detailed, human-readable status of the submission is available at `/submit/status/12345`, which is where a user is taken when searching for a submission by name through the UI. It would be nice to provide a direct link to that page in the API. My immediate thought is that it should be in the `/api/submission/12345/status` response, though perhaps it belongs in the submission API response.

### Code of Conduct

ghost:
fixes [#6072](https://github.com/ietf-tools/datatracker/issues/6072)

### DIFF
--- a/ietf/submit/views.py
+++ b/ietf/submit/views.py
@@ -182,6 +182,10 @@ def api_submission(request):
                             settings.IDTRACKER_BASE_URL,
                             urlreverse(api_submission_status, kwargs={'submission_id': submission.pk}),
                         ),
+                        'submission_status_url': urljoin(
+                            settings.IDTRACKER_BASE_URL,
+                            urlreverse("ietf.submit.views.submission_status", kwargs={'submission_id': submission.pk}),
+                        ),
                     }
                 )
             else:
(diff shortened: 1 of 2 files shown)


ietf/submit/views.py
```diff
@@ -182,6 +182,10 @@ def err(code, error, messages=None):
                             settings.IDTRACKER_BASE_URL,
                             urlreverse(api_submission_status, kwargs={'submission_id': submission.pk}),
                         ),
+                        'submission_status_url': urljoin(
```
jennifer-richards: I'm a little worried that having both `status_url` and `submission_status_url` as fields in the response is potentially confusing. My first reaction is to suggest `friendly_status_url` or just `submission_url`. Don't know if @rjsparks has any thoughts?

jennifer-richards (approved):

PR #10237, opened 2026-01-09
refactor: authors() -> author_persons()

jennifer-richards:
Rename the `DocumentInfo.authors()` method to `author_persons()` and formalize that it omits any authors that do not have a `Person`. Refactor code using this. In most cases this is just using the new method, but in a couple templates I've switched to the `author_persons_or_names()` method. That's likely a moot distinction for now, but looks ahead to when we eventually unify the author models between drafts and RFCs.

### DIFF
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -456,7 +456,6 @@ class DocumentInfo(models.Model):
             authors_qs = self.rfcauthor_set.filter(person__isnull=False)
         else:
             authors_qs = self.documentauthor_set.all()
-        print(f"nyah: {authors_qs.count()}")
         return [a.person for a in authors_qs.select_related("person")]
 
     def author_list(self):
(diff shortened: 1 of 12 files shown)


ietf/doc/models.py
```diff
@@ -435,16 +436,28 @@ def author_names(self):
...
+    def author_persons(self):
+        """Authors as a list of Persons
+        
+        Omits any RfcAuthors with a null person field.
+        """
+        if self.type_id == "rfc" and self.rfcauthor_set.exists():
+            authors_qs = self.rfcauthor_set.filter(person__isnull=False)
+        else:
+            authors_qs = self.documentauthor_set.all()
+        print(f"nyah: {authors_qs.count()}")
```
rjsparks: leftover debug

rjsparks (changes requested):
One nit

ietf/doc/models.py
jennifer-richards: nyah!

rjsparks (approved):

PR #10250, opened 2026-01-12
fix: doc property authors needs refactoring

rudimatz:
relates to https://github.com/ietf-tools/datatracker/pull/10237

currently, endpoint `POST /api/purple/rfc/bulk_authors/` is broken

### DIFF
--- a/ietf/api/serializers_rpc.py
+++ b/ietf/api/serializers_rpc.py
@@ -78,16 +78,16 @@ class AuthorPersonSerializer(serializers.ModelSerializer):
 
 
 class RfcWithAuthorsSerializer(serializers.ModelSerializer):
-    author_persons = AuthorPersonSerializer(many=True)
+    authors = AuthorPersonSerializer(many=True, source="author_persons")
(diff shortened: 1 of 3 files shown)


rjsparks (approved):

ietf/api/serializers_rpc.py
```diff
@@ -78,11 +78,11 @@ class Meta:
 
 
 class RfcWithAuthorsSerializer(serializers.ModelSerializer):
-    authors = AuthorPersonSerializer(many=True)
+    author_persons = AuthorPersonSerializer(many=True)
```
jennifer-richards: This change will impact the API and require coordinated changes with red. I suspect we want to do this by setting `source` instead. See https://github.com/rudimatz/datatracker/pull/1 (which targets the source branch for this PR)

jennifer-richards (changes requested):
I think there may be more changes needed. See inline comment also.

rjsparks (approved):

PR #10255, opened 2026-01-13
feat: rsync rfc content, store in blob, rebuild references

rjsparks:


ietf/doc/tasks.py
```diff
@@ -128,3 +130,23 @@ def investigate_fragment_task(name_fragment: str):
         "name_fragment": name_fragment,
         "results": investigate_fragment(name_fragment),
     }
+
+@shared_task
+def rebuild_reference_relations_task(doc_names:List[str]):
```
jennifer-richards:
```suggestion
def rebuild_reference_relations_task(doc_names:list[str]):
```

Just use list for annotation, List is no longer needed. Same goes for tuple and dict.

ietf/doc/tasks.py
```diff
@@ -128,3 +130,23 @@ def investigate_fragment_task(name_fragment: str):
...
+    log.log("Task: Rebuilding reference relations for"+str(doc_names))
```
jennifer-richards: string formatting - `for"+` -> `for " + ` or change to an f-string

(missing space before quote is the important bit)

ietf/doc/utils.py
```diff
@@ -816,50 +816,66 @@ def rebuild_reference_relations(doc, filenames):
...
         try:
             refs = draft.PlaintextDraft.from_file(filename).get_refs()
         except IOError as e:
-            return { 'errors': ["%s :%s" %  (e.strerror, filename)] }
+            return {"errors": ["%s :%s" % (e.strerror, filename)]}
     else:
-        return {'errors': ['No Internet-Draft text available for rebuilding reference relations. Need XML or plaintext.']}
+        return {
+            "errors": [
+                "No Internet-Draft text available for rebuilding reference relations. Need XML or plaintext."
```
jennifer-richards: nit: for RFCs, the error should technically not say "Internet-Draft text..."
```suggestion
                "No file available for rebuilding reference relations. Need XML or plaintext."
```

ietf/doc/utils.py
```diff
@@ -816,50 +816,66 @@ def rebuild_reference_relations(doc, filenames):
...
-        filename = filenames['txt']
+    if "xml" in filenames:
+        refs = XMLDraft(filenames["xml"]).get_refs()
+    elif "txt" in filenames:
+        filename = filenames["txt"]
         try:
             refs = draft.PlaintextDraft.from_file(filename).get_refs()
         except IOError as e:
-            return { 'errors': ["%s :%s" %  (e.strerror, filename)] }
+            return {"errors": ["%s :%s" % (e.strerror, filename)]}
```
jennifer-richards:
```suggestion
            return {"errors": ["%s: %s" % (e.strerror, filename)]}
```
?

ietf/sync/tasks.py
```diff
@@ -222,3 +232,80 @@ def fix_subseries_docevents_task():
         DocEvent.objects.filter(type="sync_from_rfc_editor", desc=desc).update(
             time=obsoleting_time
         )
+
+@shared_task
+def rsync_rfcs_from_rfceditor(rfc_numbers: List[int]):
```
jennifer-richards:
```suggestion
def rsync_rfcs_from_rfceditor(rfc_numbers: list[int]):
```

ietf/sync/tasks.py
```diff
@@ -222,3 +232,80 @@ def fix_subseries_docevents_task():
...
+    log.log("Rsyncing rfcs from rfc-editor: " + str(rfc_numbers))
+    from_file = None
+    with NamedTemporaryFile(mode="w", delete=False) as fp:
```
jennifer-richards: We're at python 3.12, so we now have `delete_on_close=False`. Using that instead of `delete=False` will do a better job at cleanup than manually unlinking after the loop.

ietf/sync/tasks.py
```diff
@@ -222,3 +232,80 @@ def fix_subseries_docevents_task():
...
+        )
+    if from_file is not None:
+        from_file.unlink()
+    for num in rfc_numbers:
+        for ext in settings.RFC_FILE_TYPES:
+            fs_path = Path(settings.RFC_PATH) / f"rfc{num}.{ext}"
+            if fs_path.is_file():
+                with fs_path.open("rb") as f:
+                    bytes = f.read()
+                m_time = fs_path.stat().st_mtime
```
jennifer-richards: Peculiar spelling of `m_time` (rather than `mtime`) intentional?

ietf/sync/tasks.py
```diff
@@ -222,3 +232,80 @@ def fix_subseries_docevents_task():
...
+                f"--include-from={str(from_file)}",
+                "--exclude=*",
+                "rsync.rfc-editor.org::rfcs/",
+                f"{settings.RFC_PATH}",
+            ]
+        )
+    if from_file is not None:
+        from_file.unlink()
+    for num in rfc_numbers:
+        for ext in settings.RFC_FILE_TYPES:
```
jennifer-richards: Should also grab `.json`, at least to put in the blob store

jennifer-richards (changes requested):
A few minor nits. One high-level comment: I assume it's intentional that the `prerelease/` files are stashed in the blob store but not added to datatracker's own filesystem. The RPC also publishes a `.json` that datatracker ignores afaik, but should be treated as a publication artifact.

rjsparks:
~Yes, it's intentional that notprepped isn't landing in the dt filesystem.~
~We'll be working on where the computation of the .json happens.~
I was wrong on both counts - the current system has both the notprepped and the json in the ::rfc rsync module.
I'll put them both in the fs as I'm doing the other files.

ietf/sync/tests_utils.py
```diff
@@ -0,0 +1,82 @@
...
+rfc54321.json
+prerelease/rfc54321.notprepped.xml
+""",
+        )
+
+
+class RfcBlobUploadTests(TestCase):
+    def test_load_rfcs_into_blobdb(self):
+        with TemporaryDirectory() as faux_rfc_path:
+            with override_settings(RFC_PATH=faux_rfc_path):
```
jennifer-richards, line 58: Overriding with a temp dir here is ok but redundant - `RFC_PATH` is always a per-test temp directory via the `settings_temp_path_overrides` mechanism in `ietf.utils.test_utils.TestCase`.

ietf/sync/tasks.py
```diff
@@ -234,78 +233,42 @@ def fix_subseries_docevents_task():
         )
 
 @shared_task
-def rsync_rfcs_from_rfceditor(rfc_numbers: List[int]):
-    log.log("Rsyncing rfcs from rfc-editor: " + str(rfc_numbers))
+def rsync_rfcs_from_rfceditor(rfc_numbers: list[int]):
```
jennifer-richards: Conventionally we have a `_task` suffix on task method names - mind adding it?

ietf/sync/tests.py
```diff
@@ -1134,3 +1166,51 @@ def test_iana_protocols_update_task(
...
+    @mock.patch("ietf.sync.tasks.load_rfcs_into_blobdb")
+    @mock.patch("ietf.sync.tasks.rebuild_reference_relations_task.delay")
+    def test_rsync_rfcs_from_rfceditor(
+        self,
+        rebuild_relations_mock,
+        load_blobs_mock,
+        rsync_helper_mock,
+    ):
+        tasks.rsync_rfcs_from_rfceditor([12345, 54321])
+        self.assertTrue(rsync_helper_mock.called)
```
jennifer-richards, line 1180: There's a minor gap in coverage here: this doesn't check that the from-file passed to `rsync_helper` had the expected contents.

It looks like that will be tricky to validate without a bunch of effort. I won't push back against deciding it's not worth the effort in this case.

jennifer-richards (changes requested):
A couple nits and a place where the test could be tightened (but, as noted inline, it's not clear to me it's worth the effort on that)

ietf/sync/tests.py
rjsparks: That's why I factored out the content generation and tested it separately. An additional test will cover "did we write the bytes into the file", and given the risk/benefit tradeoff (and the expected lifetime of this code), I don't think it's worth the squeeze.

ietf/sync/tests_utils.py
rjsparks: something's not right then, because it's _not_ getting a per-test temp dir. I'll leave this for now and we can remove the extra `with` when we figure that out.

jennifer-richards (approved):

Issue #10120, opened 2025-12-17
Linkifying URLs with an email address in them fails


thomwiggers:
### Describe the issue

If you are using the Mailman3 URL in a datatracker field, e.g., `https://mailman3.ietf.org/mailman3/lists/plants@ietf.org/`, then the `bleach.linkify` call (with the options specified) will parse it as a funky email address and render it wrong:

```py
>>> _bleach_linker.linkify("https://mailman3.ietf.org/mailman3/lists/tls@ietf.org/")
'https:<a href="mailto://mailman3.ietf.org/mailman3/lists/tls@ietf.org">//mailman3.ietf.org/mailman3/lists/tls@ietf.org</a>/'
```

This is probably a bug in the upstream library Bleach, but [Bleach is unmaintained](https://github.com/mozilla/bleach/issues/698) and unlikely to fix this. 

(A separate issue is probably to move away from bleach).

## Related code
https://github.com/ietf-tools/datatracker/blob/4f0102d4f48e59ab65e0545959315f060a9c3a11/ietf/utils/text.py#L54-L69

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

---
Aggregated with agent assistance on 2026-08-18

jennifer-richards:
> (A separate issue is probably to move away from bleach).

Yep

PR #10264, opened 2026-01-15
feat: use nh3 for HTML sanitization

### LINKED ISSUE #10138: Replace datatracker's use of bleach
### Description

Bleach is long deprecated. Replace it with something being maintained. nh3 is a good candidate.

### Code of Conduct

---

ghost:
fixes [#10138](https://github.com/ietf-tools/datatracker/issues/10138)


ietf/utils/html.py
```diff
@@ -15,63 +15,60 @@
...
 
 
-# Allow the protocols/tags/attributes we specifically want, plus anything that bleach declares
-# to be safe. As of 2025-01-27, the explicit lists for protocols and tags are a strict superset
-# of bleach's defaults.
-acceptable_protocols = bleach.sanitizer.ALLOWED_PROTOCOLS.union(
+# Allow the protocols/tags/attributes we specifically want, plus anything that nh3 declares
+# to be safe.
+
+acceptable_protocols = nh3.ALLOWED_URL_SCHEMES.union(
```
ghost: TODO check. this is more permissive than bleach's strict superset. however, unsure if this adds extra vulnerabilities since nh3 uses a better maintained parser https://github.com/rust-ammonia/ammonia

ghost: @jennifer-richards @rjsparks  any thoughts?

rjsparks: I'm going to try to get a few other eyes on this

ietf/utils/html.py
```diff
@@ -15,63 +15,61 @@
...
     {
         # fmt: off
         "a", "abbr", "acronym", "address", "b", "big",
         "blockquote", "body", "br", "caption", "center", "cite", "code", "col",
         "colgroup", "dd", "del", "dfn", "dir", "div", "dl", "dt", "em", "font",
         "h1", "h2", "h3", "h4", "h5", "h6", "head", "hr", "html", "i", "ins", "kbd",
-        "li", "ol", "p", "pre", "q", "s", "samp", "small", "span", "strike", "style",
+        "li", "ol", "p", "pre", "q", "s", "samp", "small", "span", "strike",
         "strong", "sub", "sup", "table", "title", "tbody", "td", "tfoot", "th", "thead",
         "tr", "tt", "u", "ul", "var"
```
martinthomson: xmp is also good
```suggestion
        "tr", "tt", "u", "ul", "var", "xmp"
```

ietf/utils/html.py
```diff
@@ -15,63 +15,61 @@
...
 
-# Allow the protocols/tags/attributes we specifically want, plus anything that bleach declares
-# to be safe. As of 2025-01-27, the explicit lists for protocols and tags are a strict superset
-# of bleach's defaults.
-acceptable_protocols = bleach.sanitizer.ALLOWED_PROTOCOLS.union(
+# Allow the protocols/tags/attributes we specifically want, plus anything that nh3 declares
+# to be safe.
+
+acceptable_protocols = nh3.ALLOWED_URL_SCHEMES.union(
     {"http", "https", "mailto", "ftp", "xmpp"}
```
martinthomson:
```suggestion
    {"ftp", "http", "https", "mailto", "tel", "xmpp"}
```
Sort.  Add "tel".

(If performance is dictated by order, move "https" to the top.)

ietf/utils/html.py
```diff
@@ -15,63 +15,61 @@
...
 
 # Instantiate sanitizer classes
-_bleach_cleaner = bleach.sanitizer.Cleaner(
+_nh3_cleaner = nh3.Cleaner(
     tags=acceptable_tags,
     attributes=acceptable_attributes,
-    protocols=acceptable_protocols,
-    strip=True,
+    url_schemes=acceptable_protocols,
+    link_rel=None
```
martinthomson: This doesn't seem right to me.  The default value is safer.  Are there cases where links opened will need `window.opener`?  I can't imagine that being necessary for user-generated content.

Same below.

ietf/utils/html.py
```diff
@@ -15,63 +15,61 @@
...
 )
 
 
-_liberal_bleach_cleaner = bleach.sanitizer.Cleaner(
-    tags=acceptable_tags.union({"img", "figure", "figcaption"}),
-    attributes=acceptable_attributes | {"img": ["src", "alt"]},
-    protocols=acceptable_protocols,
-    strip=True,
+_liberal_nh3_cleaner = nh3.Cleaner(
+    tags=acceptable_tags.union({"mg", "figure", "figcaption"}),
```
martinthomson: typo?

```suggestion
    tags=acceptable_tags.union({"img", "figure", "figcaption"}),
```

ietf/utils/html.py
```diff
@@ -15,63 +15,61 @@
...
-    return _liberal_bleach_cleaner.clean(text)
+    return _liberal_nh3_cleaner.clean(text)
 
 
 @keep_lazy(str)
 def remove_tags(html, tags):
     """Returns the given HTML sanitized, and with the given tags removed."""
     allowed = acceptable_tags - set(t.lower() for t in tags)
-    return bleach.clean(html, tags=allowed, strip=True)
+    return nh3.clean(html, tags=allowed)
```
martinthomson: You might want to audit invocations of this function.

ietf/utils/html.py
ghost: cool, changed it here as there were no rel attributes before

Jason-123-cyber (commented):
``kk

ietf/utils/html.py
larseggert: I think adding `tel:` and `ftp:` (really?) are fine from a security perspective.

martinthomson: We don't really need ftp, but tel might be useful.

rjsparks:
Hi @nouralmaa - Thanks for continuing to work on this.

One point on something that's not obvious - commits bf72278 and  9736b74 touch an old data migration (see the filename) - those will never run again, and its better to not touch these migration files as they should show what happened when they actually ran on the production database. If it's easy, please revert those commits. If it's not, we'll do it when this gets towards final review.

When we have major changes to django (we _may_ have one coming with django5) we tend to squash the migrations and that one would go away at that time.

ietf/utils/templatetags/tests.py
```diff
@@ -53,6 +53,10 @@ def test_linkify(self):
             linkify("https://www.ietf.org"),
             '<a href="https://www.ietf.org" rel="nofollow">https://www.ietf.org</a>',
         )
+        self.assertEqual(
+            linkify("https://mailman3.ietf.org/mailman3/lists/tls@ietf.org/"),
```
ghost: fixes [#10120](https://github.com/ietf-tools/datatracker/issues/10120)

Image

ietf/utils/templatetags/textfilters.py
```diff
@@ -76,9 +76,8 @@ def texescape_filter(value):
 @stringfilter
 def linkify(value, autoescape=True):
     if autoescape:
-        # Escape unless the input was already a SafeString
-        value = conditional_escape(value)
-    text = mark_safe(_linkify(value))  # _linkify is a safe operation
+        value = urlize(value, autoescape=True)  # _linkify is a safe operation
+    text = mark_safe(value)
```
jennifer-richards: This won't work as written - it needs to call `urlize()` in all cases, not only when `autoescape` is true. However, 

1. The commit log around `ietf.doc.templatetags.ietf_filters.urlize` mentions issues with Django's `urlize` and its handling of adjacent parentheses. Need to investigate that and whether it's been fixed.
2. If it is ok to switch to `urlize`, we should just do away with `linkify` as it's redundant.

ietf/group/tests_info.py
```diff
@@ -2136,12 +2135,10 @@ def test_see_status_update(self):
 
     def test_view_status_update(self):
         chair = RoleFactory(name_id='chair',group__type_id='wg')
-        event = GroupEventFactory(type='status_update',group=chair.group)
```
jennifer-richards, line 2139: Unexpected change. Unless there's a reason not to test this behavior, need to fix the assertion below rather than drop the test.

ietf/utils/html.py
jennifer-richards: Not sure if @martinthomson did so himself, but it looks to me like the only use is via the `removetags` filter in `htmlfilters.py`, which is not used anywhere (I also checked the `DBTemplates`)

If so, we can just lose this method entirely.

If not, I think this changes `remove_...` to `escape_...` because nh3 doesn't have a `strip` option (but that's just based on reading docs)

ghost:
> Hi @nouralmaa - Thanks for continuing to work on this.
> 
> One point on something that's not obvious - commits [bf72278](https://github.com/ietf-tools/datatracker/commit/bf72278de808ea65e738a19144ea8e3f9094d08f) and [9736b74](https://github.com/ietf-tools/datatracker/commit/9736b7416db2c9f74f99bfa420bc1dfc268df4d7) touch an old data migration (see the filename) - those will never run again, and its better to not touch these migration files as they should show what happened when they actually ran on the production database. If it's easy, please revert those commits. If it's not, we'll do it when this gets towards final review.
> 
> When we have major changes to django (we _may_ have one coming with django5) we tend to squash the migrations and that one would go away at that time.

re these commits, I've reverted the changes and also made a change to use html instead of having the links hardcoded into the slugs for that template. I imagine that may be necessary down the road in any case, but feel free to revert if not.

ietf/utils/html.py
ghost: Doesn't seem to raise any validation errors

ietf/utils/templatetags/textfilters.py
ghost: have adjusted so this can be called in both cases. 
1. re the handling of adjacent parentheses and this [commit](https://github.com/ietf-tools/datatracker/commit/2c0348cc7a31746f465182a119d06d5ffc61288e), I can understand why django's urlizer might not be ideal for linkification like `[REF](http://example.com/foo)`, and the issue still stands. however, I think the markdown filter inside `ietf.utils.templatetags.htmlfilters` can be used in such cases with links intact
2. yes, but it would require changing lots of internal html templates and tests to do away with the linkify filter completely. maybe can be raised in a separate issue? can also try to have a look soon.

PR #10289, opened 2026-01-20
fix: restrict rfc bulkload to those with a Document object

rjsparks:


ietf/sync/utils.py
```diff
@@ -26,17 +26,42 @@ def build_from_file_content(rfc_numbers: list[int]) -> str:
...
+                        bytes = f.read()
+                    mtime = fs_path.stat().st_mtime
+                    try:
+                        store_bytes(
+                            kind="rfc",
+                            name=f"{ext}/rfc{num}.{ext}",
+                            content=bytes,
+                            allow_overwrite=False,  # Intentionally not allowing overwrite.
+                            doc_name=f"rfc{num}",
+                            doc_rev=None,
```
jennifer-richards, line 45: If you wanted to be paranoid, you could iterate over `Documents` instead of numbers and use `doc.name` and `doc.rev` here. I think the data are clean in this regard, though.

jennifer-richards (approved):
lgtm

One thought inline in case it resonates, but if the doc names are corrupted in a way that it matters then we're in deeper trouble already.

PR #10356, opened 2026-02-03
feat: speed up agenda.ics; cache more agenda data

### LINKED ISSUE #10355: caching for get_href / get_versionless_href mixes data between these calls
### Describe the issue

The `DocumentInfo_get_href()` helper caches its result, but it does not segregate the cache between different values of its `meeting_doc_refs` parameter. As a result, `DocumentInfo.get_href()` and `DocumentInfo.get_versionless_href()` return incorrect values if calls to them are interleaved.

E.g., using an arbitrary agenda:
```python
>>> from ietf.doc.models import *
>>> ag = Document.objects.filter(type_id="agenda", group__acronym="asdf").last()
>>> ag.get_href()
'https://datatracker.ietf.org/meeting/interim-2026-asdf-02/materials/agenda-interim-2026-asdf-02-asdf-01-00'
>>> ag.get_versionless_href()
'https://datatracker.ietf.org/meeting/interim-2026-asdf-02/materials/agenda-interim-2026-asdf-02-asdf-01-00'
>>>
```
Note that these are both the same and end with `-01-00`. Now restart the interpreter and...
```python
>>> from ietf.doc.models import *
>>> ag = Document.objects.filter(type_id="agenda", group__acronym="asdf").last()
>>> ag.get_versionless_href()
'https://datatracker.ietf.org/meeting/interim-2026-asdf-02/materials/agenda-interim-2026-asdf-02-asdf-01'
>>> ag.get_href()
'https://datatracker.ietf.org/meeting/interim-2026-asdf-02/materials/agenda-interim-2026-asdf-02-asdf-01'
>>>
```
again both the same, but without the final `-00`!

### Code of Conduct

jennifer-richards:
Speeds up agenda.ics endpoints for full IETF meetings to near instant. Assuming we keep the caches warm as intended, it will also speed up all agenda page loads. The updates do not affect interim meetings, only `type_id="ietf"`.

* Refactors the `agenda_ical()` and its helpers to use the same precomputed data that is supplied to the agenda-neue frontend.
* Adds agenda caching for all meetings, not just the current meeting, with a week-scale cache lifetime for non-current meetings.
* Changes the ical links to use versionless agenda hrefs so they'll always point to the current revision.
* Adds a task to populate/refresh the cache for all meetings so it will always be warm.
* Renames `agenda_data_refresh` to `agenda_data_refresh_task` (keeps/deprecates old task until we deploy and update our config to use the new name)
* Fixes #10355

I propose that we continue refreshing the agenda data for the "current" meeting every 5 minutes but only weekly for other meetings. That was assumed in choosing the new constants in `settings.py`. This does not add a mechanism for invalidating the cache dynamically if the agenda is changed. A change like that will need a manual cache refresh if waiting for the next regular refresh is not acceptable.

Note: The cache is kept in memcache. Data for each meeting is 200-300 kB, so it fits well within memcache's size limit. Restarting memcached will lose cached data, though. I think this is acceptable because memcached restarts are rare and the cache will recover on its own as requests come in, usually with at most one slow request per meeting.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,10 +62,6 @@ services:
         image: ghcr.io/ietf-tools/static:latest
         restart: unless-stopped
 
-    memcache:
-        image: memcached:1.6-alpine
-        restart: unless-stopped
-
     mq:
         image: rabbitmq:3-alpine
         restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,6 +62,10 @@ services:
         image: ghcr.io/ietf-tools/static:latest
         restart: unless-stopped
 
+    memcache:
+        image: memcached:1.6-alpine
+        restart: unless-stopped
+
     mq:
         image: rabbitmq:3-alpine
         restart: unless-stopped
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -239,14 +239,14 @@ class DocumentInfo(models.Model):
         return revisions
 
     def get_href(self, meeting=None):
-        return self._get_ref(meeting=meeting,meeting_doc_refs=settings.MEETING_DOC_HREFS)
+        return self._get_ref(meeting=meeting, versioned=True)
 
 
     def get_versionless_href(self, meeting=None):
-        return self._get_ref(meeting=meeting,meeting_doc_refs=settings.MEETING_DOC_GREFS)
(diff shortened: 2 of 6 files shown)


jennifer-richards:
n.b., this was merged without review into the feat/ical feature branch so it could be deployed to a dev instance. Review / actual merge to main will be via #10362

PR #10411, opened 2026-02-13
fix: remove conditional formatting for liaison bodies

ghost:
fixes [#10129](https://github.com/ietf-tools/datatracker/issues/10129)

### DIFF
--- a/ietf/templates/liaisons/detail.html
+++ b/ietf/templates/liaisons/detail.html
@@ -203,7 +203,7 @@
                     Body
                 </th>
                 <td>
-                    <pre>{{ liaison.body|maybewordwrap:"80"|urlize_ietf_docs|linkify }}</pre>
+                    <pre>{{ liaison.body|urlize_ietf_docs|linkify }}</pre>
                 </td>
             </tr>
         {% endif %}


rjsparks:
Simple code fix, but it's going to need a companion data fix, or at least an inspection to verify that the effect on the display of the existing statements is acceptable.

The entire liaison app is up for reimplementation (that effort is in a design phase that will accelerate after we deploy the rpc modernrization work). It would be best to check whether changes in that app are likely to just be replaced soon anyhow.

Rather than do the data display inspection now, lets hold it for the reimplementation.

ghost:
> Simple code fix, but it's going to need a companion data fix, or at least an inspection to verify that the effect on the display of the existing statements is acceptable.
> 
> The entire liaison app is up for reimplementation (that effort is in a design phase that will accelerate after we deploy the rpc modernrization work). It would be best to check whether changes in that app are likely to just be replaced soon anyhow.
> 
> Rather than do the data display inspection now, lets hold it for the reimplementation.

okk, happy to try the markdown preview option. just simple fix as i was editing the template

rjsparks:
Just in case - I think I'm still not shining the light on what I am trying to point out. This last line from the issue:

> Either will require an effort to adjust existing bodies to display correctly after the change.

It's the _existing_ data that will have to be inspected and possibly reconditioned when we make this change, not newly entered statements.

And again, this has really become input into the requirements for the re-implementation rather than requirements for point fixes.

PR #10412, opened 2026-02-13
feat: cancel interim meeting at materials page

ghost:
fixes [#9887](https://github.com/ietf-tools/datatracker/issues/9887)
Screenshot 2026-02-16 at 14 52 12

### DIFF
--- a/ietf/meeting/tests_views.py
+++ b/ietf/meeting/tests_views.py
@@ -5889,8 +5889,37 @@ class InterimTests(TestCase):
         q = PyQuery(r.content)
         self.assertEqual(len(q("a.btn:contains('nnounce')")),2)
 
-    def test_interim_request_details_cancel(self):
-        """Test access to cancel meeting / session features"""
+    def test_interim_meeting_cancel(self):
+        make_interim_test_data()
+        mars_sessions = Session.objects.with_current_status(
+        ).filter(
+            meeting__type='interim',
+            group__acronym='mars',
+            current_status="sched"
+        )
+        meeting = mars_sessions.first().meeting
+        # All these roles should have access to cancel the request
+        usernames_and_passwords = (
+            ('marschairman', 'marschairman+password'),
+            ('secretary', 'secretary+password')
+        )
+        url = urlreverse('ietf.meeting.views.session_details',
+                         kwargs={'num': meeting.number, 'acronym': 'mars'})
+        
+        for username, password in usernames_and_passwords:
+            self.client.login(username=username, password=password)
+            r = self.client.get(url)
+            self.assertEqual(r.status_code, 200)
+            q = PyQuery(r.content)
+            cancel_meeting_btns = q("a.btn:contains('Cancel meeting')")
+            self.assertEqual(len(cancel_meeting_btns), 1,
+                'Should be exactly one cancel meeting button for user %s' % username)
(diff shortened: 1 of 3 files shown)


rjsparks:
Thanks for pushing this forward.

This one is going to need additional tests.

The code/template might need additional logic - the view needs to account for whether the meeting has already passed. Look to see if that's already available. Right now, it's gating on whether the materials cutoff has passed for chairs, and ignores whether the meeting is in the past if the user is part of the secretariat.

On these views, the button to cancel should not be presented to anyone for a meeting that has already passed.

ghost:
> Thanks for pushing this forward.
> 
> 
> 
> This one is going to need additional tests.
> 
> 
> 
> The code/template might need additional logic - the view needs to account for whether the meeting has already passed. Look to see if that's already available. Right now, it's gating on whether the materials cutoff has passed for chairs, and ignores whether the meeting is in the past if the user is part of the secretariat.
> 
> 
> 
> On these views, the button to cancel should not be presented to anyone for a meeting that has already passed.

have adjusted!

ietf/meeting/tests_views.py
```diff
@@ -5889,8 +5889,37 @@ def test_interim_request_details_announcement(self):
...
+            self.assertEqual(r.status_code, 200)
+            q = PyQuery(r.content)
+            cancel_meeting_btns = q("a.btn:contains('Cancel meeting')")
+            self.assertEqual(len(cancel_meeting_btns), 1,
+                'Should be exactly one cancel meeting button for user %s' % username)
+            self.assertEqual(cancel_meeting_btns.eq(0).attr('href'),
+                urlreverse('ietf.meeting.views.interim_request_cancel',
+                           kwargs={'number': meeting.number}),
+                                'Cancel meeting points to wrong URL')
+
```
rjsparks, line 5921: Please add tests confirming that the anonymous user, a user with no roles (see other tests for `nobody`), and a chair of a different group are _not_ presented with the link.

ietf/templates/meeting/session_details_panel.html
```diff
@@ -38,6 +38,16 @@ <h3 class="" id="session_{{ session.pk }}">
                             Meeting details
                         </a>
                     {% endif %}
+                    {%  with meeting_status.slug as status_slug %}
```
rjsparks, line 41: This with is now not used.

ietf/meeting/tests_views.py
```diff
@@ -5889,8 +5889,37 @@ def test_interim_request_details_announcement(self):
         q = PyQuery(r.content)
         self.assertEqual(len(q("a.btn:contains('nnounce')")),2)
 
-    def test_interim_request_details_cancel(self):
-        """Test access to cancel meeting / session features"""
+    def test_interim_meeting_cancel(self):
```
rjsparks, line 5892: (This note is just a comment not a request for change).
This move to call out the scope of the cancel is not incorrect, but it is somewhat subtle. It reflects what the current code assumes, but there is an outstanding bug (or at least misfeature) in the current code. We are slowly returning to having multi-day in-person interims. Right now the misfeature makes the groups that have them schedule them as multiple interim meetings. When we address that problem, the scope of this cancellation view will likely need to shift back to session (or we will need _both_ cancel session and cancel meeting). Again, nothing to do about it right now other than be aware.

rjsparks (changes requested):

PR #10436, opened 2026-02-20
fix: update WG State descriptions as per IESG decision 12-02-2025

paulwouters:


rjsparks:
This PR is changing a derived product - the `names.json` fixture is created from the content of the database and is only used by the tests. I'll create a migration that will change the data in the database to these values instead.

rjsparks:
Closing this in favor of #10437

PR #10465, opened 2026-02-24
feat: notify errata system of new rfc

rjsparks:


ietf/doc/utils_errata.py
```diff
@@ -0,0 +1,29 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
```
jennifer-richards, line 1: Run ruff on this module? It has some quote / indentation deviations from the preferred style.

jennifer-richards (dismissed):
Maybe run ruff on utils_errata.py but ok to leave that for later if you'd rather merge and move on.

jennifer-richards (approved):

PR #10481, opened 2026-02-28
feat: add attendance summary and pie chart to meeting attendees page

larseggert:
For IETF meetings ≥ 118, the attendees proceedings page now shows an Onsite / Remote / Total summary row matching the counts displayed on registration.ietf.org, together with a "View chart" button that opens a Bootstrap modal containing a Highcharts pie chart.

### DIFF
--- a/ietf/meeting/tests_views.py
+++ b/ietf/meeting/tests_views.py
@@ -9007,6 +9007,8 @@ class ProceedingsTests(BaseMeetingTestCase):
            - assert onsite checkedin=True appears, not onsite checkedin=False
            - assert remote attended appears, not remote not attended
            - prefer onsite checkedin=True to remote attended when same person has both
+           - summary stats row shows correct counts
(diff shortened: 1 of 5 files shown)


jennifer-richards (changes requested):
The code looks reasonable to me.

Technically speaking: we're trying to get rid of inline JS as much as we can. We should move the inline json into a js file in `ietf/static/js`, add it to the packaging list in `package.json`, and then reference it via `{% static 'js/whatever.js' %}` in the template. That can still be done conditionally using the template tags. I can help with that if you don't want to get into those details.

We've had lots of requests for this sort of data, so it's a great addition. We should confirm that we trust the data and aren't going to induce firestorms by adding it though. @rjsparks , opinions about that?

rjsparks:
The data being shown here is good.

jennifer-richards (approved):

PR #10526, opened 2026-03-09
feat: rfc-index generation

jennifer-richards:
This is based on @kesara's index code from purple.

Generates rfc-index.xml. Relies on three files existing in the "red bucket": 
* `unusable-rfc-numbers.json`
    - list of `{"number": 123, "comment": "comment or empty string"}`
* `april-first-rfc-numbers.json`
    - list of integers
* `publication-std-levels.json`
    - list of `{"number": 123, "publication_std_level": "some slug"}`
These should be generated by purple and deposited in the "red" bucket. A directory prefix within the bucket can be set via `settings.RFCINDEX_INPUT_PATH`, which defaults to `""`.

Indexes are written to the same bucket using `RFCINDEX_OUTPUT_PATH` (also defaults to `""`)

Expects to share the same credentials as for the blobdb replication storages.

Adds support for `Document.keywords` and adds API support to allow these to be set from purple.

Does not yet have triggering a / task wrapper.

For development, defaults to using an `InMemoryStorage`, which does not leave traces on the filesystem. Without the JSON files, the RFC index will be wrong: there will be no unusable RFC numbers listed, no RFCs will be treated as April 1, and all publication-status values will be UNKNOWN. This situation will raise an exception in production. In development, a log entry will be emitted as a reminder. The tests fill in their own JSON data. Comments in `settings_local.py` suggest a way to create a persistent `FileSystemStorage` which is useful for development.

### Todo
- [x] Tests
- [x] Real `Storage` configuration for `red_bucket`
- [x] Extract settings from `rfcindex.py` into the settings module

### DIFF
--- a/docker/configs/settings_local.py
+++ b/docker/configs/settings_local.py
@@ -101,6 +101,17 @@ for storagename in ARTIFACT_STORAGE_NAMES:
         ),
     }
 
+# For dev on rfc-index generation, create a red_bucket/ directory in the project root
+# and uncomment these settings. Generated files will appear in this directory. To
+# generate an accurate index, put up-to-date copies of unusable-rfc-numbers.json,
+# april-first-rfc-numbers.json, and publication-std-levels.json in this directory
+# before generating the index.
+#
+# STORAGES["red_bucket"] = {
+#     "BACKEND": "django.core.files.storage.FileSystemStorage",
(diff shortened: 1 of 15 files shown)


garethkirman4-web:
> _ No description provided. _

kesara (dismissed):
LGTM!

rjsparks (approved):

PR #10538, opened 2026-03-12
fix: Rewrite upper right document search box

ekr:
This is a fix to the problem where the first item in the dropdown is auto-selected and then when you hit return you go to that rather than searching for what's in the text field.

It appears to be challenging to get this behavior with select2, so this is actually a rewrite of the box with explicit behavior. As a side effect, the draft names actually render a bit better.

Fixes #10358

### DIFF
--- a/ietf/templates/base.html
+++ b/ietf/templates/base.html
@@ -75,8 +75,7 @@
                                autocomplete="off"
                                data-ajax-url="{% url 'ietf.doc.views_search.ajax_select2_search_docs' model_name='document' doc_type='all' %}"
                                aria-label="Document search">
-                        <ul class="dropdown-menu dropdown-menu-end" id="navbar-doc-search-results"
-                            style="max-height: 400px; overflow-y: auto; min-width: 350px;">
+                        <ul class="dropdown-menu" id="navbar-doc-search-results">
                         </ul>
                     </div>
                 </div>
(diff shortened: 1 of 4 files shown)


jennifer-richards (changes requested):
Thanks for this. I think it needs a couple tweaks before we can merge it.

The draft names are all cut off by the right-hand side of the window, showing in many cases the least informative part of draft names. The old wrapped names were a little awkward, but this seems less usable to me. 

Image

Also, we should extract the JS code into a module rather than inlining it in `base.html`. [This commit shows what's needed](https://github.com/ietf-tools/datatracker/pull/10481/changes/cce530c1611d16fe83344968c8f41333697edc60), but I'd be happy to do that refactor if you prefer.

ietf/templates/base.html
```diff
@@ -67,13 +67,18 @@
...
+                    <div class="d-none d-md-block dropdown" id="navbar-doc-search-wrapper">
+                        <input class="form-control"
+                               id="navbar-doc-search"
+                               type="text"
+                               placeholder="Document search"
+                               autocomplete="off"
+                               data-ajax-url="{% url 'ietf.doc.views_search.ajax_select2_search_docs' model_name='document' doc_type='all' %}"
+                               aria-label="Document search">
+                        <ul class="dropdown-menu dropdown-menu-end" id="navbar-doc-search-results"
+                            style="max-height: 400px; overflow-y: auto; min-width: 350px;">
```
jennifer-richards: Also - the style needs to be pulled out into a CSS class. Our validation won't accept inline styles.

ekr:
Updated per review comments.

jennifer-richards (approved):

PR #10549, opened 2026-03-14
feat: add author affiliation in serializer

rudimatz:


ietf/api/serializers_rpc.py
```diff
@@ -100,14 +100,18 @@ class DocumentAuthorSerializer(serializers.ModelSerializer):
...
 
     class Meta:
         model = DocumentAuthor
-        fields = ["person", "plain_name"]
+        fields = ["person", "plain_name", "affiliation"]
 
     def get_plain_name(self, document_author: DocumentAuthor) -> str:
         return document_author.person.plain_name()
 
+    def get_affiliation(self, document_author: DocumentAuthor) -> str:
```
jennifer-richards: This should not be needed, just adding `affiliation` to `fields` ought to deal with all the workings. Did you try that?

jennifer-richards (approved):

PR #10550, opened 2026-03-14
feat: add some per country/affiliation meeting registration statistics

evyncke:
Very simple (read trivial) attempt to present some registration statistics using pie-chart (from charts.js).

Missing is some percentage data (only participants count for now), expand country code to country name, be able to change the meeting number (beside fiddling with URL).

Alas, no data is kept on the ticket creation date, i.e., cannot have a timeline of registration for a specific meeting.


evyncke:
Could also do stats per affiliation ;-)

evyncke:
Meeting statistics are done, so, this PR could be merged (I won't work on it unless tools-team requests so to fix issues).

Next codesprint will handle the documents statistics but this should not prevent already merging this PR.

jennifer-richards (changes requested):
Looks really nice. Before merging, we need to move code from the inline `<script>` tag into a module. [This commit](https://github.com/ietf-tools/datatracker/pull/10481/changes/cce530c1611d16fe83344968c8f41333697edc60) illustrates what's involved, but I'm happy to handle that if you'd prefer.

We may want/need to add some caching to the views as well.

evyncke:
If you do not mind, then please go forward with the move to a specific JS script.

And the caching would be useful as it takes 2-3 seconds on my laptop to generate the data (and the data does not really change hourly). Could generate a big JSON every day ?

jennifer-richards:
> If you do not mind, then please go forward with the move to a specific JS script. 

Ok, will do.

> And the caching would be useful as it takes 2-3 seconds on my laptop to generate the data (and the data does not really change hourly). Could generate a big JSON every day ?

I'll see about adding at least basic caching before we bring this live.

Thanks!

evyncke:
> > If you do not mind, then please go forward with the move to a specific JS script.
> 
> Ok, will do.
> 
> > And the caching would be useful as it takes 2-3 seconds on my laptop to generate the data (and the data does not really change hourly). Could generate a big JSON every day ?
> 
> I'll see about adding at least basic caching before we bring this live.
> 
> Thanks!

@jennifer-richards any news ? I understand that it is low priority for you and if you cannot do it, then I will give the 'moving JS out of template' part a try.

Unsure about the caching, any pointer/reference/example on how to do it ?

Thanks again

jennifer-richards:
@evyncke 

My apologies. I expect that the next datatracker release will be next week and do plan to get this wrapped up for that. If you want to work on it a bit I don't object, but it's just been slow, not forgotten.

jennifer-richards:
I've put the refactoring + caching on a PR here: https://github.com/evyncke/datatracker/pull/1 - if you merge that in your fork, it should appear in this PR

jennifer-richards (approved):

PR #10552, opened 2026-03-14
feat: show parents on list of teams with grouping (#8635)

### LINKED ISSUE #8635: Add Area/Parent group to list of teams
### Description

The list of teams at https://datatracker.ietf.org/team/ would benefit from having a new column for the parent group / area.

### Code of Conduct

breakertt:
Show parents of teams on the list of active teams in a grouping way, resolves #8635)

### DIFF
--- a/ietf/group/views.py
+++ b/ietf/group/views.py
@@ -245,7 +245,7 @@ def active_review_dirs(request):
     return render(request, 'group/active_review_dirs.html', {'dirs' : dirs })
 
 def active_teams(request):
-    teams = Group.objects.filter(type="team", state="active").order_by("name")
+    teams = Group.objects.filter(type="team", state="active").order_by("parent__name", "name").select_related("parent")
     for group in teams:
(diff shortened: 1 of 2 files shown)


rjsparks:
Thanks - this shows up some data grooming that will need to be performed before this is deployed, and I think points out we need a custom sort (rather than parent name alphabetical).
image

Several of the "other" things really need different parents. I'll work through those directly with the secretariat.

But we should sort the things that are areas first, then the things that are really representing stream bodies, then the LLC.

breakertt:
I have pushed a new commit where parents are sorted as area first, then stream bodies and other parents, then administrative, then teams with parents as None to Others.

jennifer-richards (commented):
Looks ok to me, but waiting to hear whether @rjsparks would like the parent group sections ordered differently

rjsparks:
@jennifer-richards : This ordering is fine to start with.

jennifer-richards (approved):

PR #10558, opened 2026-03-16
feat: UUIDs as person identifiers

jennifer-richards:
Creates a one-to-many relationship from `Person` to UUID values, which permits URLs like `https://datatracker.ietf.org/person/18dd451b-66df-43d6-b8b9-b22ada29e985` to identify a unique person. Reassigns UUIDs as part of a person merge. (To be clear: this means a `Person` will have, in general, multiple such URLs.)

PR as draft while we consider the ramifications of a feature like this...


jennifer-richards:
Superseded by (and included in) #11415

PR #10563, opened 2026-03-16
fix: reduce db churn and log noise for rebuilding references

rjsparks:


ietf/doc/utils.py
```diff
@@ -1014,6 +1035,13 @@ def rebuild_reference_relations(doc, filenames):
     if unfound:
         ret["unfound"] = list(unfound)
 
+    logmsg = f"rebuild_reference_relations for {doc.name}: "
+    logmsg += "changed referrences" if changed_references else "references unchanged"
```
jennifer-richards: typo
```suggestion
    logmsg += "changed references" if changed_references else "references unchanged"
```

jennifer-richards (changes requested):
One typo

jennifer-richards (approved):

PR #10567, opened 2026-03-17
feat: trigger red recomputation on RFC publication or metadata update

rjsparks:


ietf/doc/utils_red.py
```diff
@@ -0,0 +1,34 @@
...
+
+import requests
+
+from django.conf import settings
+
+from ietf.utils.log import log
+
+
+def trigger_red_precomputer(rfc_number_list=()):
+    # TODO load the following into settings from the env
```
jennifer-richards: todo -> todone?

ietf/doc/utils_red.py
```diff
@@ -0,0 +1,34 @@
...
+    url = getattr(settings, "TRIGGER_RED_PRECOMPUTE_MULTIPLE_URL", None)
+    if url is not None:
+        headers = {"Content-Type": "application/json"}
```
jennifer-richards: `"application/json"` should be set automatically when using the `json` parameter rather than the `data` parameter

jennifer-richards (approved):
ok, but you may want to address a nit about a todo that I think is completed and a place where you could (but don't have to) drop a redundant line

ietf/api/serializers_rpc.py
```diff
@@ -682,6 +683,7 @@ def update(self, instance, validated_data):
                 stale_subseries_relations.delete()
             if len(rfc_events) > 0:
                 rfc.save_with_history(rfc_events)
+            trigger_red_precomputer_task.delay(rfc_number_list=[rfc.rfc_number])
```
jennifer-richards: Just realized that this is happening in a transaction (started on l.565). Need to move it outside the transaction or it will have sync issues.

jennifer-richards (changes requested):
Noticed a must-fix

ietf/api/serializers_rpc.py
```diff
@@ -683,7 +683,9 @@ def update(self, instance, validated_data):
                 stale_subseries_relations.delete()
             if len(rfc_events) > 0:
                 rfc.save_with_history(rfc_events)
-            trigger_red_precomputer_task.delay(rfc_number_list=[rfc.rfc_number])
+        # Intentionally triggering red even if the above transaction is rolled back.
+        # The unnecessary computation red would do is small.
```
jennifer-richards: fwiw, unnecessary triggers will be unlikely. Normally a rollback will be in response to an exception in the `atomic()` block, which will be re-raised outside the block. I think even a commit failure would result in an exception, but I can't find documentation of that at the moment.

jennifer-richards (approved):

ietf/api/tests_serializers_rpc.py
```diff
@@ -83,6 +87,10 @@ def test_update(self):
             result.part_of(),
             [Document.objects.get(name="fyi999")],
         )
+        self.assertTrue(mock_task_delay.called)
+        _, mock_kwargs = mock_task_delay.call_args
+        self.assertIn("rfc_number_list", mock_kwargs)
+        self.assertEqual(mock_kwargs["rfc_number_list"], [rfc.rfc_number])
 
     def test_partial_update(self):
```
jennifer-richards: This test needs the `mock_task_delay` treatment

jennifer-richards (changes requested):

jennifer-richards (approved):

PR #10575, opened 2026-03-18
feat: update typesense search index on rfc pub/update

jennifer-richards:
Does an `upsert()` with data intended to match what the initial JS index build tool generated. Has been tested against a local typesense instance, but have not done a systematic comparison at all.

~~Draft while I work on the tests.~~ Ready to go, I think

Things to think about for ~~this or~~ later work: (edit: I think best to take this as a complete feature and leave these for future improvements)
* The `_sanitize_text()` method in `searchindex.py` is based on the JS treatment of the content field. I don't completely understand the goals, but it would be good to review this for reasonableness.
* We'll want to add a full rebuild action using typesense's bulk update API. That's not in this PR.
* Datatracker should probably be in charge of the Typesense schema for this collection. That's also not in this PR. For now, it's using what's already set up.

### DIFF
--- a/ietf/api/serializers_rpc.py
+++ b/ietf/api/serializers_rpc.py
@@ -1,4 +1,4 @@
-# Copyright The IETF Trust 2025, All Rights Reserved
+# Copyright The IETF Trust 2025-2026, All Rights Reserved
 import datetime
 from pathlib import Path
 from typing import Literal, Optional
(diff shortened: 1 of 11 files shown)


NGPixel (dismissed):

rjsparks (approved):

evyncke:
Should requirements.txt be updated to include the "httpx" module ?

jennifer-richards:
> Should requirements.txt be updated to include the "httpx" module ?

Yes, since we refer to it we should include it. I think it should be pulled in automatically as a dependency of the typesense package, though, so it shouldn't actually be required for this to work as intended.

jennifer-richards:
> Should requirements.txt be updated to include the "httpx" module ?

https://github.com/ietf-tools/datatracker/pull/10678

PR #10631, opened 2026-03-31
feat: Generate bcp-index.txt

kesara:
Fixes #10533

Sample: [`bcp-index.txt`](https://t4.fq.nz/dt.10631/bcp-index.txt) _(Updated with authors.)_

TODO:
- [x] Update header description. But need more eyes on that.
- [x] Make sure RFCs are sorted.

### DIFF
--- a/ietf/sync/rfcindex.py
+++ b/ietf/sync/rfcindex.py
@@ -24,6 +24,8 @@ from ietf.name.models import StdLevelName
 from ietf.utils.log import log
 
 FORMATS_FOR_INDEX = ["txt", "html", "pdf", "xml", "ps"]
+BCP_TXT_MARGIN = 3
+BCP_TXT_CUE_COL_WIDTH = 14
 
 
 def format_rfc_number(n):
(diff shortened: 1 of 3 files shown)


richsalz:
The format description at the beginning of the sample is not correct. No opinion from me on which is better :)

Since this is plain-text, remove the `<` and `>` around the URL's to make it easier to cut and paste them.

kesara:
@richsalz, Thanks for the feedback. `<` and `>` was added because that's how RFC text files show BCPs in the reference. The new `bcp-index.txt` follows that format.

richsalz:
Yes, it follows the RFC text format, but it's still less useful.  Will there be HTML versions of the index?

kesara:
> Yes, it follows the RFC text format, but it's still less useful. Will there be HTML versions of the index?

No, at the moment only HTML index like view is https://www.rfc-editor.org/bcps which redirects to a search result view.
New rfc-editor website will have the similar functionality with a search results view.

HTML index would be a feature request for future updates on new rfc-editor.org website.

richsalz:
Thanks for the answer.  I still think the angle brackets is wrong

ajeanmahoney:
Hi @richsalz, if you would like to push back on angle brackets, please open an issue at https://github.com/rfc-editor/draft-rpc-rfc7322bis/issues, thanks!

richsalz:
I don't care about the angle brackets in RFC text format.  If I want clickable links, I will go to the HTML version. Since there is no HTML version of this index, users have to cut and paste any URLs in the document. Rendering them with angle brackets make that cut and paste a little harder. That is my point.

rjsparks:
@richsalz part of what we were asked to do with the subseries indices was to match the format that other reference generation, hence Jean's push of your feedback towards the style guide. (And I think it would be good to change that in the style guide).

rjsparks (dismissed):

rjsparks (approved):

PR #10636, opened 2026-03-31
fix: look in target if draft name not found in anchor

rjsparks:
Fixes #10634


rjsparks:
@larseggert you were here not to long ago - would appreciate your input.

ietf/utils/xmldraft.py
```diff
@@ -102,6 +102,17 @@ def _document_name(self, ref):
             number = int(maybe_number)
             return f"{label}{number}"
 
+        target = ref.get("target")
+        if isinstance(target, str):
+            target = target.lower()
+            if target.startswith("https://datatracker.ietf.org/doc/"):
```
larseggert, line 108: Depending on this not changing is a bit ugly. Can we at least make `https://datatracker.ietf.org/` and maybe `doc` tool-wide constants?

jennifer-richards: I agree and would be inclined to split out a helper like `targetsDatatrackerDoc()` (or some not terrible name. It could probably wait until we see a second use for it, though.

jennifer-richards (approved):
Looks like an improvement to me. Could perhaps be refactored, but I don't think it's essential.

rjsparks:
I guess we should bring this in and go for incremental improvement later - moving it out of draft.

PR #10692, opened 2026-04-10
feat: serve notprepped rfc xml from datatracker

rjsparks:
Adds the "Get Editor Source" button here:
image
Which leads to:
image

Claude was used while generating this PR.


ietf/doc/storage_utils.py
```diff
@@ -178,8 +178,7 @@ def retrieve_bytes(kind: str, name: str) -> bytes:
                     content = f.read()
         except Exception as err:
             log(f"Blobstore Error: Failed to read bytes from {kind}:{name}: {repr(err)}")
-            if settings.SERVER_MODE == "development":
-                raise
+            raise
```
rjsparks, line 181: n.b. this is a major change

jennifer-richards: Indeed! #10704 separates it out so it'll appear more clearly in the release history

jennifer-richards (approved):

PR #10703, opened 2026-04-13
fix: Stop interim edit from clobbering Meetecho remote_instructions (#10689)

### LINKED ISSUE #10689: Changing an interim agenda puts the interim in an unrecoverable state
### Describe the issue

As WG chair, I was working on preparing the WEBBOTAUTH interim `interim-2026-webbotauth-01`. I went in to [update the agenda for the meeting](https://datatracker.ietf.org/meeting/interim/request/interim-2026-webbotauth-01) and the datatracker replied:
> An error occurred while creating a Meetecho conference. The interim meeting request has been created without complete remote participation information. Please edit the request to add this or contact the secretariat if you require assistance.

Before I made that change, the [interim page](https://datatracker.ietf.org/doc/agenda-interim-2026-webbotauth-01-webbotauth-01/) would say:
> Remote instructions:  https://meetings.conf.meetecho.com/interim/?session=35206

But once I changed the agenda the [interim page](https://datatracker.ietf.org/doc/agenda-interim-2026-webbotauth-01-webbotauth-01/) started showing the Remote instructions as empty.

I had to contact the Secretariat to resolve the issue. They then had to delete the original session and create a new one. This all caused a lot of spam on the mailing list: (emails [one](https://mailarchive.ietf.org/arch/msg/web-bot-auth/IWcvj8Os5C0zapYM4THAjfP_K0M/) [two](https://mailarchive.ietf.org/arch/msg/web-bot-auth/Epx-XhjfWoY3WobP0fjfx6_syac/) [three](https://mailarchive.ietf.org/arch/msg/web-bot-auth/j0CTp1WkAKOMS72MI9qrLLEcVtE/)) which led to confusion amongst participants. Now we also have [two interims listed instead of one](https://datatracker.ietf.org/group/webbotauth/meetings/), as `interim-2026-webbotauth-01` was replaced with `interim-2026-webbotauth-02`.

I recommend disabling the ability to edit agendas until the datatracker stops putting meetings in such an unrecoverable state.

### Code of Conduct

ekr:
This PR was written by @claude but reviewed by @ekr 

Editing an interim meeting backed by an auto-created Meetecho conference used to lose the Remote instructions URL whenever the subsequent Meetecho API call failed (and silently orphan the old conference when it succeeded). The WG chair was left with an interim that pointed nowhere and had to escalate to the secretariat to manually recover — losing uploaded slides in the process.

Root cause: `InterimSessionModelForm.clean()` unconditionally blanked `remote_instructions` whenever the form was submitted with `remote_participation=='meetecho'`, on the assumption that the post-save helper would then fill in a fresh Meetecho URL. That assumption held for the create flow but falls apart on edit: if the helper's API call to Meetecho fails for any reason (network, auth, rate limit), the URL is already gone from the in-memory form data *and* has been persisted by `ModelForm.save()`, leaving nothing to fall back on.

Fixes:

- `forms.py`: stop blanking `remote_instructions` in `clean()`. The helper layer decides whether to overwrite the URL and only does so on successful Meetecho creation. Also initialize the `remote_participation` radio from the stored URL on edit so a no-op resubmit doesn't flip modes.

- `helpers.py`: only (re)create a Meetecho conference when one of the scheduling fields (`date`, `time`, `requested_duration`, `end_time`) actually changed. Merely editing the agenda now leaves the existing conference alone. On a successful recreate, the old conference is fetched and deleted so it is not orphaned on the Meetecho side.

- Tests in `tests_forms.py` and `tests_helpers.py` cover:
  - form-level: clean() preserves the URL across a Meetecho-mode edit;
  - helper-level: failure path (exception and empty result) preserves the URL; success path deletes the previous conference;
  - post-save filter: skip when only non-timing fields changed, recreate when any timing field changed, always create when no URL exists yet.

Note: the "is this a Meetecho URL" check currently uses a substring match on "meetecho"; a dedicated flag on Session would be cleaner but requires a migration and is out of scope for this bugfix.

Fixes #10689.

### DIFF
--- a/ietf/meeting/helpers.py
+++ b/ietf/meeting/helpers.py
@@ -1082,10 +1082,11 @@ def sessions_post_save(request, forms):
             form.save_agenda()
 
         try:
-            create_interim_session_conferences(
+            create_interim_session_conferences([
                 form.instance for form in forms
                 if form.cleaned_data.get('remote_participation', None) == 'meetecho'
-            )
+                and _needs_meetecho_refresh(form)
+            ])
         except RuntimeError:
             messages.warning(
                 request,
@@ -1095,11 +1096,30 @@ def sessions_post_save(request, forms):
             )
 
 
+# Form fields that, when changed, require a Meetecho conference to be
+# (re)created because they alter the scheduled time or duration.
+_MEETECHO_TIMING_FIELDS = frozenset(('date', 'time', 'requested_duration', 'end_time'))
+
+
+def _needs_meetecho_refresh(form):
+    """Does this session's Meetecho conference need to be (re)created?
+
+    If the session already has a Meetecho URL and no timing-related field
+    changed, we leave the existing conference alone — recreating it would
+    orphan the old one and is unnecessary.
+    """
+    existing = form.instance.remote_instructions or ''
+    if 'meetecho' not in existing:
+        return True  # no Meetecho conference yet (empty or manual URL)
+    return bool(_MEETECHO_TIMING_FIELDS.intersection(form.changed_data))
+
+
 def create_interim_session_conferences(sessions):
     error_occurred = False
     if hasattr(settings, 'MEETECHO_API_CONFIG'):  # do nothing if not configured
         meetecho_manager = meetecho.ConferenceManager(settings.MEETECHO_API_CONFIG)
         for session in sessions:
+            previous_url = session.remote_instructions or ''
             ts = session.official_timeslotassignment().timeslot
             try:
                 confs = meetecho_manager.create(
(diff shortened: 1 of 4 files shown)


ekr:
Hmm.. This didn't fail locally. Debugging.

ekr:
Pretty hard to see how this PR could have created this bug. I can't seem to re-kick off the CI, but maybe try doing that and see if it's an intermittent somehow?

ietf/meeting/helpers.py
alexamirante, line 1131: If a Meetecho session already exists with that `session_id`, you cannot use the `create` api to update it: it will always fail.

We may need to implement an `update` api on the Meetecho side if we want to safely change timing-related parameters. 
Otherwise you may call the `delete` api first and then `create`, but in this case a guard is needed to avoid deleting a session that has already started (which would kick everybody out of the room).

ietf/meeting/helpers.py
```diff
@@ -1116,7 +1136,23 @@ def create_interim_session_conferences(sessions):
...
+                    try:
+                        for conference in meetecho_manager.fetch(session.group):
+                            if conference.url == previous_url:
+                                conference.delete()
+                                break
+                    except Exception as err:
+                        log.log(
+                            f'Exception deleting old Meetecho conference {previous_url} '
+                            f'for {session}: {err}'
+                        )
```
alexamirante, line 1152: I don't think you can never get to this code because it's not possible to have multiple meetecho sessions with the same `session_id=session.pk` (see previous comment).

ietf/meeting/helpers.py
ekr: `update` seems better. Do you have a sense of how long that would take?

rjsparks: Having an update API (as it's easy to implement) is fine, but we need to be careful about when/how to use it. There's been a lot of pressure to never remove a scheduled session from a published calendar (like /meeting/upcoming, or the main meeting agenda for `ietf` meetings). People that saw the original announcement/made plans have asked that there be a clear "canceled" or "moved" notification at the place they originally expected the session to be.

But this is wandering away from this _particular_ issue as David wasn't trying to change the time parameters.

PR #10708, opened 2026-04-14
feat: debounce rfc index refresh

jennifer-richards:
Adds the now ubiquitous `DirtyBits` model for the RFC index. The API call purple used to trigger the task now just marks the index as dirty.

Renames the `create_rfc_index_task` to `refresh_rfc_index_task`. This should be schedule to run periodically (5-10 minutes, depending on how long index generation takes in practice).

(n.b., this does not yet import errata.json updates, that will come separately)

@kesara, this touches the task you're working on in #10706 but that should only be superficial and should be easy to merge.

@rjsparks, note that the third commit (931e9bc) removes some ca 2011 code that is unused as far as I can tell. I searched for imports fairly thoroughly, but if you think there's a chance it's still live I can revert that.


ietf/api/views_rpc.py
```diff
@@ -545,5 +545,5 @@ class RfcIndexView(APIView):
         request=None,
     )
     def post(self, request):
-        create_rfc_index_task.delay()
+        mark_rfcindex_as_dirty()
```
kesara, line 551: I don't see any process calling `mark_rfcindex_as_processed()`.

kesara (dismissed):

rjsparks:
>@rjsparks, note that the third commit (https://github.com/ietf-tools/datatracker/commit/931e9bc1963d736c8a7d85cf73b524cf9233b341) removes some ca 2011 code that is unused as far as I can tell. I searched for imports fairly thoroughly, but if you think there's a chance it's still live I can revert that.

I agree that this is unreachable code and it is appropriate to remove it.

ietf/api/views_rpc.py
jennifer-richards: good catch, thanks!

rjsparks (approved):

PR #10732, opened 2026-04-17
chore: remove blobdb profiling logs

jennifer-richards:
These are not useful any more, blobdb is fast


rjsparks (approved):
Are we also already providing DATATRACKER_BLOB_STORE_ENABLE_PROFILING on our current k8s deployments?

jennifer-richards:
> Are we also already providing DATATRACKER_BLOB_STORE_ENABLE_PROFILING on our current k8s deployments?

Yes, it is set to true on the production / staging deployments

PR #10779, opened 2026-04-29
fix: Disable calt font feature

### LINKED ISSUE #10778: Disable Contextual alternates (calt) font feature in doc view
### Describe the issue

Doc view[^1] uses Inter font with Contextual alternates (calt)[^2] which displays strings like `sntrup761x25519` as `sntrup761×25519`.

Therefor `calt` font feature should be disabled in the doc view.

[^1]: https://datatracker.ietf.org/doc/rfc9941/
[^2]: https://luisangelmaciel.github.io/inter/#features/calt

### Code of Conduct

kesara:
Fixes #10778 

This disables `calt` font feature site wide.
I couldn't find a way to disable this via bootstrap settings.

Screenshot:
Screenshot 2026-04-29 at 15 55 14

### DIFF
--- a/ietf/static/css/ietf.scss
+++ b/ietf/static/css/ietf.scss
@@ -46,6 +46,11 @@ $bootstrap-icons-font-src: url("npm:bootstrap-icons/font/fonts/bootstrap-icons.w
 url("npm:bootstrap-icons/font/fonts/bootstrap-icons.woff") format("woff");
 @import "bootstrap-icons/font/bootstrap-icons";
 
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):

jennifer-richards:
> I couldn't find a way to disable this via bootstrap settings.

I think the more selective approach would be to adjust the `@font-face` rules for the Inter font to add the setting. I don't think the other fonts we have will be affected by this flag, though, so we can at least run with this for now.

PR #10782, opened 2026-04-29
feat: Check if any files exists before publishing RFC

### LINKED ISSUE #10757: Additional checks before publishing
Add additional checks before publishing to fail early rather than failing midway of publishing.
These checks can:
 * Verify if published RFC output files doesn't existing in the blobstore.
 * Datatracker can accept the published RFC output files.

kesara:
Fixes #10757

### DIFF
--- a/ietf/api/tests_views_rpc.py
+++ b/ietf/api/tests_views_rpc.py
@@ -151,6 +151,16 @@ class RpcApiTests(APITestCase):
         self.assertEqual(r.status_code, 409)  # conflict
         file_in_the_way.unlink()
 
+        # Put a blob in the way. Post should fail because replace = False
+        blob_in_the_way = Blob.objects.create(
+            bucket="rfc", name=f"txt/rfc{unused_rfc_number}.txt", content=b""
+        )
+        r = self.client.post(
+            url, data=post_data, format="json", headers={"X-Api-Key": "valid-token"}
+        )
+        self.assertEqual(r.status_code, 409)  # conflict
+        blob_in_the_way.delete()
+
         r = self.client.post(
             url, data=post_data, format="json", headers={"X-Api-Key": "valid-token"}
         )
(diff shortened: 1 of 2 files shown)


ietf/api/tests_views_rpc.py
```diff
@@ -135,6 +140,19 @@ def test_notify_rfc_published(self, mock_task_delay):
         r = self.client.post(url, data=post_data, format="json")
         self.assertEqual(r.status_code, 403)
 
+        # Put a file in the way. Post should fail because files exists
+        with TemporaryDirectory() as rfc_dir:
+            settings.RFC_PATH = rfc_dir  # affects overridden settings
```
jennifer-richards: I think you copied a pattern I used in the other test that's unnecessary - tests always replace `RFC_PATH` with a temporary directory that gets replaced between test cases and cleaned up at the end. It's harmless, so ok to keep it consistent and clean up both instances later.

ietf/api/views_rpc.py
```diff
@@ -368,7 +368,38 @@ def get_queryset(self):
         )
 
 
-class RfcPubNotificationView(APIView):
+class EnhancedAPIView(APIView):
```
jennifer-richards: Since the additions in this class don't (that I see) involve any of the `APIView` workings, suggest making it a mixin. E.g., `class DestinationHelperMixin:...` and then the view classes become `class WhateverView(APIView, DestinationHelperMixin):...`

jennifer-richards (changes requested):
Nice! Two things (plus a comment inline that is just a comment, not a request for change). See inline for thing one. Thing two: test should cover the "blob-exists" case in addition to the "file-exists" case.

ietf/api/tests_views_rpc.py
kesara: Addressed with 503821f3641f8eef52aa85bbe60f23350bfd8b4c

ietf/api/views_rpc.py
kesara: Addressed with 58a51728c5c9068b8b7fca64c16cf0b23e4646d5

jennifer-richards (approved):

PR #10785, opened 2026-04-29
chore: add DATATRACKER_BROKER_URL config option

jennifer-richards:
Adds `DATATRACKER_BROKER_URL` as an alternative for configuring the rabbitmq connection. This is used when connecting to a rabbitmq cluster managed by the kubernetes operator.

When not present, falls back to the old `CELERY_PASSWORD` mechanism and assumes it will be using the custom rabbitmq StatefulSet. (Regardless of the settings, the rabbitmq StatefulSet is still deployed.)


kesara (approved):
LGTM.

PR #10800, opened 2026-05-01
fix: ipv4-only for internal traffic

jennifer-richards:
Listening on ipv6 interfaces may have contributed to increased CPU usage during heavy traffic. This switches back to ipv4 only and configures nginx to use 127.0.0.1 instead of localhost for proxying.

Does not affect external access, this is purely internal to the k8s cluster.


NGPixel (approved):

sarashed731-prog (commented):
{
  "version": "v0.9",
  "action": {
    "name": "submit_reservation",
    "surfaceId": "booking-surface",
    "sourceComponentId": "submit-btn",
    "timestamp": "2026-02-25T10:40:00Z",
    "context": {
      "time": "7:00 PM",
      "size": 4
    }
  }
}

PR #10815, opened 2026-05-04
feat: add documents and authors statistics 

evyncke:
Beside minor update to the existing meetings statistics, this PR adds several statistics related to documents (streams, publication status) and to authors (country and affiliation).

Need to add the caching system though (not enabled now to facilitate testing), hence it is still a draft.


evyncke:
Only one remaining thing to code: the cache of all timelines as they are heavy

evyncke:
All the features I wanted to add are in, caching is enabled (and verified working), tests cover the new codes/templates/URLs.

So ready to be reviewed before merging.

ietf/stats/utils.py
```diff
@@ -38,25 +60,22 @@ def get_aliased_affiliations(affiliations):
...
+      "Examplar International" -> "Examplar" """
 
     res = {}
 
     ending_re = compile_affiliation_ending_stripping_regexp()
 
     known_aliases = { alias.lower(): name for alias, name in AffiliationAlias.objects.values_list("alias", "name") }
+    # Let's prepare a dict for things like "Google Inc." -> "Google" adding a space to the end of the main name 
+    # so we only match it at the beginning of the affiliation and not in the middle of it, e.g. "Google Analytics"
+    affiliation_main_names = [(main_name.lower() + ' ', main_name) for main_name in AffiliationMainName.objects.values_list("main_name", flat=True)]
```
jennifer-richards: I'm not understanding the comment/aim of this. The `main_name` values are things like `"Google"`, so I think adding a space would match both `"Google Inc."` and `"Google Analytics"`. The comment reads to me like it says it _won't_ match the latter. Am I misreading?

ietf/stats/factories.py
```diff
@@ -2,10 +2,20 @@
...
+from ietf.stats.models import AffiliationIgnoredEnding, AffiliationMainName, MeetingRegistration
 from ietf.meeting.factories import MeetingFactory
 from ietf.person.factories import PersonFactory
 
+class AffiliationIgnoredEndingFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = AffiliationIgnoredEnding
+
+    ending = ''
+class AffiliationMainNameFactory(factory.django.DjangoModelFactory):
```
jennifer-richards, line 16: nit (style): Should have a couple newlines between class definitions

ietf/stats/models.py
```diff
@@ -42,6 +42,18 @@ class AffiliationIgnoredEnding(models.Model):
...
+    main_name = models.CharField(
+        max_length=255, 
+        unique=True,
+        help_text="Main leading part of an affiliation, the remaing part can be ignored.")
+
+    class Meta:
+        verbose_name_plural = 'affiliation main names'
+
+    def __str__(self):
+        return self.main_name
```
jennifer-richards: nit (style): should have two newlines between classes

ietf/stats/models.py
```diff
@@ -42,6 +42,18 @@ class AffiliationIgnoredEnding(models.Model):
     def __str__(self):
         return self.ending
 
+class AffiliationMainName(models.Model):
+    """Records that this start of an affiliation is what matters (for statistical purposes)."""
+    main_name = models.CharField(
+        max_length=255, 
+        unique=True,
+        help_text="Main leading part of an affiliation, the remaing part can be ignored.")
```
jennifer-richards: typo: `remaing` -> `remaining`

ietf/stats/models.py
```diff
@@ -42,6 +42,18 @@ class AffiliationIgnoredEnding(models.Model):
     def __str__(self):
         return self.ending
 
+class AffiliationMainName(models.Model):
+    """Records that this start of an affiliation is what matters (for statistical purposes)."""
```
jennifer-richards, line 43: High level question: how many affiliation aliases would we need to add in order not to need this additional model / matching mechanism? I gather that it works, but it will need to be maintained and adds complexity to the process.

ietf/stats/urls.py
```diff
@@ -1,18 +1,21 @@
...
 from ietf.utils.urls  import url
+from ietf.stats import views_authors, views_documents, views_meetings, views_reviews
 
 urlpatterns = [
     url(r"^$", views.stats_index),
-    url(r"^document/(?:(?P<stats_type>authors|pages|words|format|formlang|author/(?:documents|affiliation|country|continent|citations|hindex)|yearly/(?:affiliation|country|continent))/)?$", views.document_stats),
+    url(r"^authors/(?P<doc_type>all|draft|wg-draft|rfc)/(?P<stats_type>affiliation|country)/$", views_authors.authors_timeline),
+    url(r"^total/authors/(?P<doc_type>all|draft|wg-draft|rfc)/(?P<stats_type>affiliation|country)/$", views_authors.authors_total),
+    url(r"^documents/(?P<doc_type>draft|rfc)/(?P<stats_type>level|stream|wg)/$", views_documents.documents_timeline),
+    url(r"^total/documents/(?P<doc_type>draft|rfc)/(?P<stats_type>level|stream|wg)/$", views_documents.documents_total),
```
jennifer-richards: Putting the `total/` as a prefix to the stat type seems to break the hierarchical nature of the url namespace. I think it'd be cleaner as a suffix. E.g., `documents/rfc/wg/` would give the timeline and `documents/rfc/wg/total/` would give the total. Would that make sense to you?

ietf/stats/urls.py
```diff
@@ -1,18 +1,21 @@
...
     url(r"^knowncountries/$", views.known_countries_list),
-    url(r"^meeting/$", views.meetings_timeline),
-    url(r"^meeting/(?P<meeting_number>\d+)/(?P<stats_type>affiliation|country)/$", views.meeting_stats),
-    url(r"^meeting/(?:(?P<stats_type>affiliation|country|total)/)?$", views.meetings_timeline),
-    url(r"^review/(?:(?P<stats_type>completion|results|states|time)/)?(?:%(acronym)s/)?$" % settings.URL_REGEXPS, views.review_stats),
+    url(r"^meeting/(?P<meeting_number>\d+)/(?P<stats_type>affiliation|country)/$", views_meetings.meeting_stats),
+    url(r"^meetings/(?:(?P<stats_type>affiliation|country|total)/)?$", views_meetings.meetings_timeline),
```
jennifer-richards: Both the meeting-related URLs should have the same prefix. I.e., either singular or plural. I'd be inclined to make them both plural to match `authors` and `documents`.

I also wonder if the `total` should be made into a suffix to match the `authors` and `documents` cases.

ietf/stats/views_authors.py
```diff
@@ -0,0 +1,298 @@
...
+        top_n: Maximum number of top groups to include before aggregating into Other.
+
+    Returns:
+        A Chart.js-compatible data dictionary.
+    """
+    # Build a dynamic query set filter
+    filters = Q()    
+    if doc_type != 'all' and doc_type  != 'wg-draft':
+        filters &= Q(document__type_id=doc_type)
+    if doc_type == 'wg-draft':
```
jennifer-richards: nit: use `elif` instead

ietf/stats/views_authors.py
```diff
@@ -0,0 +1,298 @@
...
+        doc_type: Document category to filter on.
+        group_by: Field used to group authors.
+        top_n: Maximum number of top groups to include before aggregating into Other.
+
+    Returns:
+        A Chart.js-compatible data dictionary.
+    """
+    # Build a dynamic query set filter
+    filters = Q()    
+    if doc_type != 'all' and doc_type  != 'wg-draft':
```
jennifer-richards: The `all` case should probably filter to `draft` and `rfc` as a precaution. We don't currently attach `DocumentAuthor` records (or `RfcAuthor`, see other comment) to other doc types, but it'd be better to guard.

ietf/stats/views_authors.py
```diff
@@ -0,0 +1,298 @@
...
+        filters &= Q(document__type_id=doc_type)
+    if doc_type == 'wg-draft':
+        filters &= Q(document__type_id= 'draft')
+        filters &= Q(document__name__startswith='draft-ietf')
+    queryset = (
+        DocumentAuthor.objects
```
jennifer-richards: So - fun complication. For `rfc` type documents, we no longer use `DocumentAuthor`. We still have them for pre-RPC tool-cutover `Document` instances, but they should not be used. I think you'll need to add a second query over `RfcAuthor` when gathering stats that include the `rfc` type, then combine the two for `all` stats.

ietf/stats/views_authors.py
```diff
@@ -0,0 +1,298 @@
...
+    Returns:
+        A Chart.js-compatible data dictionary.
+    """
+    # Build a dynamic query set filter
+    filters = Q()    
+    if doc_type != 'all' and doc_type  != 'wg-draft':
+        filters &= Q(document__type_id=doc_type)
+    if doc_type == 'wg-draft':
+        filters &= Q(document__type_id= 'draft')
+        filters &= Q(document__name__startswith='draft-ietf')
```
jennifer-richards: An alternative would be `document__group__type_id="wg"`. That would more explicitly check the group type. The results should be similar but there may be edge case / outlier documents where they don't agree.

ietf/stats/views_authors.py
```diff
@@ -0,0 +1,298 @@
...
+        if doc_type != 'all' and doc_type  != 'wg-draft':
+            filters &= Q(document__type_id=doc_type)
+        if doc_type == 'wg-draft':
+            filters &= Q(document__type_id= 'draft')
+            filters &= Q(document__name__startswith='draft-ietf')
+        queryset = (
+            DocumentAuthor.objects
+            .select_related('document')
+            .filter(filters)
+        )
```
jennifer-richards: Same comments re: filters / `RfcAuthor` as for the `_total_data_` method above

ietf/stats/views_authors.py
```diff
@@ -0,0 +1,298 @@
...
+    # ── Step 3: Get top N and others ── must be outside of the cache
+    top_groups = sorted(
+        documents_totals.keys(),
+        key=lambda c: documents_totals[c],
+        reverse=True
+    )[:top_n]
+    non_top_groups = documents_totals.keys() - set(top_groups)
+    other_totals = defaultdict(int)
+    for y in years_list:
+        other_totals[y] = 0
```
jennifer-richards: don't need this line, defaultdict will handle the initialization for you

ietf/stats/views_documents.py
```diff
@@ -0,0 +1,296 @@
...
+        doc_type: Document type filter ('rfc', 'draft', 'all', 'wg-draft').
+        group_by: Field to group by (e.g., 'stream__name', 'group__name').
+        top_n: Number of top groups to display.
+
+    Returns:
+        Chart.js compatible data dictionary with labels and datasets.
+    """
+    # Build a dynamic query set filter
+    filters = Q()
+    if doc_type != 'all' and doc_type != 'wg-draft':
```
jennifer-richards: Here, you definitely want to filter by type even for `all`, as I assume you don't want to include things like slides, etc. (If you're getting the results you expect with this code, it's likely because we don't set the fields you're grouping by on these other `Document` types, but that's fragile.)

ietf/stats/views_documents.py
```diff
@@ -0,0 +1,296 @@
...
+        filters &= Q(type_id=doc_type)
+    if doc_type == 'wg-draft':
```
jennifer-richards: nit again: `elif` instead

ietf/stats/views_documents.py
```diff
@@ -0,0 +1,296 @@
...
+    top_groups = sorted(
+        documents_totals.keys(),
+        key=lambda c: documents_totals[c],
+        reverse=True
+    )[:top_n]
+    non_top_groups = set(documents_totals.keys()) - set(top_groups)
+    other_totals: Dict[int, int] = defaultdict(int)
+    other_bin_is_empty = True
+    for y in years_set:
+        other_totals[y] = 0
```
jennifer-richards: not needed since it's a `defaultdict(int)`

ietf/stats/migrations/0003_update_aliases.py
```diff
@@ -0,0 +1,133 @@
+
```
jennifer-richards, line 2: needs copyright boilerplate

ietf/stats/views_meetings.py
```diff
@@ -0,0 +1,518 @@
...
+def get_country_data_for_meetings(attendance_type: Optional[str] = None, top_n: int = 20) -> Tuple[List[str], List[Dict[str, Any]]]:
+    """Get country participation data for meetings timeline chart.
+
+    Args:
+        attendance_type: Optional filter for attendance type (e.g., 'onsite').
+
+    Returns:
+        Tuple of (sorted_meetings, datasets) for Chart.js.
+    """
+    cache_key = f'stats:get_country_data_for_meetings:{attendance_type}-{top_n}'
```
jennifer-richards: Here and elsewhere that the `top_n` parameter comes in, we should restrict choices to a small number of reasonable options. Otherwise a malicious cache-buster can create a lot of work by enumerating the integers...

k8s/settings_local.py
```diff
@@ -18,7 +18,7 @@
 
 
 def _multiline_to_list(s):
-    """Helper to split at newlines and convert to list"""
+    """Helper to split at newlines and conver to list"""
```
jennifer-richards: I think the changes in this file must be from a bad merge. Please revert. We should check whether there were other regressions.

.gitignore
```diff
@@ -1,7 +1,6 @@
 .DS_store
```
jennifer-richards, line 1: Again, did you change this file on purpose? These look like a bad merge.

jennifer-richards (changes requested):
Comments so far are from reading the code, I haven't yet tried running it but trust that it works. It looks good, but needs a few adjustments. At a high level:

* the authors stats need to be updated to handle `RfcAuthor` for `rfc` type docs
* the `main_name` heuristic for affiliations looks reasonable, but I'm curious about how it interacts with the existing aliases. I guess it's further coalescing similar names? Also a couple inline questions about some implementation details.
* the URL naming should probably be adjusted to be more hierarchical . There's room for discussion; I have a couple suggestions inline
* caching based on `top_n` exposes an easy cache-busting enumeration attack, so we should limit the choices.
* a few changes that look like they were inadvertent merge artifacts

Happy to discuss if any of my comments seem off mark.

ietf/stats/migrations/0003_update_aliases.py
evyncke: Done

.gitignore
evyncke: Tried to revert it back (this was indeed a mistake of mine)

ietf/stats/views_documents.py
evyncke: Fixed, i.e., L200 removed

ietf/stats/models.py
evyncke: Fixed

ietf/stats/views_meetings.py
evyncke: Good catch indeed !

I will add a drop-down choice of 5-10-20-50-100 in the templates and check the top_n values in the views*.py.

evyncke: Implemented now (mainly in stats/utils.py + relevant views & templates)

ietf/stats/utils.py
evyncke: I must admit that the comments were not really useful or readable :-( 

Updated now with more sensible comments (I hope)

ietf/stats/models.py
evyncke: Correct, as there is no 'logic' in actual affiliation and the list of affiliation keeps changing (new companies, merges, e.g., Juniper & HPE), I am afraid that this will be dynamic and will require maintenance (perhaps a handful of changes per year)

ietf/stats/views_documents.py
evyncke: oh good point indeed!

Let me think about it

ietf/stats/urls.py
evyncke: Good idea, implemented now.

ietf/stats/urls.py
evyncke: Good idea indeed, implemented.

ietf/stats/views_documents.py
evyncke: Done

ietf/stats/views_authors.py
evyncke: Good idea, code fixed 2 x views_authors and 1 x views_documents.

With this change, there are slighly more (about 5%) WG drafts counted, i.e., not all WG documents follow the naming draft-ietf-* :-O

ietf/stats/views_authors.py
evyncke: Done for the filtering, still working on RfcAuthor though

ietf/stats/views_authors.py
evyncke: RfcAuthor is now used for 'rfc' and 'all', @jennifer-richards one issue though as RfcAuthor.country is always the empty string. Was there an error during the import from RFCEditor ?

ietf/stats/views_authors.py
evyncke: Also done

ietf/stats/views_authors.py
jennifer-richards: As we discussed elsewhere, no, country data simply is not available for the RfcAuthor model 😢

evyncke: Changed the UI to avoid RFC/All by authors' countries

evyncke:
CI test again failing at `test_bash (ietf.secr.telechat.tests.SecrTelechatTestCase.test_bash) ... ERROR`

evyncke:
@rjsparks ' review has been addressed.

PR #10828, opened 2026-05-06
chore: drop unused config + unused k8s file

jennifer-richards:


jennifer-richards:
Converted to draft - apparently the secrets.yaml + readme in the `k8s/` directory are meant to be there as examples. Planning to add the readme back with the secrets.yaml in the readme instead of as a separate yaml file. That will help avoid the impression that the example file might be in use.

jennifer-richards:
Merged up to date and sample `secrets.yaml` restored (with a `.example` suffix to make it clear that it's not in use)

NGPixel (approved):

PR #10834, opened 2026-05-08
feat: Celery task to create BibXML for a RFC

kesara:


ietf/sync/tasks.py
```diff
@@ -192,12 +195,12 @@ def iana_protocols_update_task():
 
     rfc_numbers = iana.parse_protocol_page(response.text)
 
-    def batched(l, n):
-        """Split list l up in batches of max size n.
+    def batched(ls, n):
```
kesara: `ruff check` complained about this becasue `l` can be ambiguous[^1].

[^1]: https://docs.astral.sh/ruff/rules/ambiguous-variable-name/

ietf/sync/tasks.py
```diff
@@ -342,3 +345,11 @@ def refresh_rfc_index_task():
             pass
 
         mark_rfcindex_as_processed(new_processed_time)
+
+
+@shared_task
+def recreate_rfc_bibxml_task():
+    try:
+        recreate_rfc_bibxml()
+    except Exception as e:
```
jennifer-richards: Catching this exception will cause the celery task to succeed even if an exception is raised. Is that wanted?

If you let the exception escape here, celery will log a backtrace for us and the failure will be somewhat more visible in grafana. 

In other places where we're catching all the exceptions in tasks, we have some other mechanism for alerting the right people that something failed. I don't _think_ that's the case here.

ietf/sync/bibxml.py
```diff
@@ -0,0 +1,67 @@
...
+        ContentFile(content if isinstance(content, bytes) else content.encode("utf-8")),
+    )
+    log(f"Saved {bucket_path} in bibxml_bucket storage")
+
+
+def get_rfc_bibxml(rfc_number):
+    """Return BibXML entry for the given rfc"""
+
+    rfc = Document.objects.get(rfc_number=rfc_number)
+    link = f"{settings.RFC_EDITOR_INFO_BASE_URL}rfc{rfc_number}"
```
jennifer-richards: I'd suggest `urllib.parse.urljoin(settings.RFC_EDITOR_INFO_BASE_URL + "/", f"rfc{rfc_number}")`

The `+ "/"` makes sure that it won't act funny if the setting is missing a trailing slash. A doubled slash there will be de-duplicated
```
>>> urllib.parse.urljoin("https://test.example.com/frag//", "rfc123")
'https://test.example.com/frag/rfc123'
````

File: ietf/sync/tasks.py [hunk 4]
```diff
@@ -192,12 +195,12 @@ def iana_protocols_update_task():
 
     rfc_numbers = iana.parse_protocol_page(response.text)
 
-    def batched(l, n):
-        """Split list l up in batches of max size n.
+    def batched(ls, n):
+        """Split list ls up in batches of max size n.
 
         For Python 3.12 or later, replace this with itertools.batched()
```
jennifer-richards: I think we're at python 3.12 now - replace with `itertools.batched()`?

ietf/sync/tests_bibxml.py
```diff
@@ -0,0 +1,70 @@
...
+        ).doc
+
+    def test_get_rfc_bibxml(self):
+        bibxml = get_rfc_bibxml(self.rfc.rfc_number)
+        self.assertIsNotNone(ElementTree.fromstring(bibxml))
+        self.assertIn(f"RFC{self.rfc.rfc_number}", bibxml)
+        self.assertIn(
+            f"{settings.RFC_EDITOR_INFO_BASE_URL}rfc{self.rfc.rfc_number}", bibxml
+        )
+        self.assertIn('<date month="April" year="2021"/>', bibxml)
```
jennifer-richards, line 39: Is there a schema or other way to test the bibxml output more deeply?

jennifer-richards (commented):
See some questions / suggestions inline.

At a high level, how do you imagine this being triggered? I think it only has a "rebuild all" task now. Should there be a do-one mechanism?

ietf/sync/tasks.py
kesara: Addressed with c3c7779b6c09637e02935cfda47992ec3d7e323d

ietf/sync/bibxml.py
kesara: Good idea! But we don't really need to append an extra `/` here.
```
>>> urljoin("https://example.com/", "bar")
'https://example.com/bar'
>>> urljoin("https://example.com", "bar")
'https://example.com/bar'
```

kesara: Addressed with 613e21a6a8031ef479be94168c2fac4e3bb8b6d9

ietf/sync/tests_bibxml.py
kesara: No, bibxml doesn't have it's own schema, it's part of RFCXML schema.

ietf/sync/tasks.py
kesara: Addressed with 6943bf94e086ca66c8819121bb9bc2415a01bce6

ietf/sync/bibxml.py
jennifer-richards: > Good idea! But we don't really need to append an extra `/` here.
> 
> ```
> >>> urljoin("https://example.com/", "bar")
> 'https://example.com/bar'
> >>> urljoin("https://example.com", "bar")
> 'https://example.com/bar'
> ```

The case you have to be wary of is
```python
>>> from urllib.parse import urljoin
>>> urljoin("https://example.com/some-path", "bar")
'https://example.com/bar'
```
vs
```python
>>> urljoin("https://example.com/some-path" + "/", "bar")
'https://example.com/some-path/bar'
```

jennifer-richards (approved):
Approving (note that I re-pointed it to the new feat/bibgen feature branch).

See response inline regarding the `"/"` in the `urljoin()` call. The code is correct as is, though, and we aren't likely to have a path in the base URL. If we do, we just have to be sure we append the trailing `/` to the setting.

ietf/sync/bibxml.py
kesara: 4f977a16efadb1d633489bb974b45ff8553390e0 addresses this issue.

PR #10848, opened 2026-05-12
fix: allow empty keyword list for Document

jennifer-richards:
Without this change, the Document admin won't allow saving a document without at least one keyword in its array. ModelForms involving Document would likely also have this issue.


rjsparks (approved):

amartinezjr0310310-cyber (commented):
Thanks🫠

PR #10851, opened 2026-05-12
feat: add updates and obsoletes for RFCs

rudimatz:
fixes part of https://github.com/ietf-tools/purple/issues/1167

### DIFF
--- a/ietf/api/serializers_rpc.py
+++ b/ietf/api/serializers_rpc.py
@@ -600,6 +600,24 @@ class EditableRfcSerializer(serializers.ModelSerializer):
             "obsoletes",
         ]
 
+    def _validate_rfc_number_list(self, field_name, rfc_numbers):
+        """Raise ValidationError if any RFC numbers in the list don't exist."""
+        unknown = [
+            n for n in rfc_numbers
+            if not Document.objects.filter(rfc_number=n, type_id="rfc").exists()
+        ]
+        if unknown:
+            raise serializers.ValidationError(
+                {field_name: [f"Unknown RFC number: {n}" for n in unknown]}
+            )
+        return rfc_numbers
+
+    def validate_updates(self, value):
+        return self._validate_rfc_number_list("updates", value)
+
+    def validate_obsoletes(self, value):
+        return self._validate_rfc_number_list("obsoletes", value)
+
     def create(self, validated_data):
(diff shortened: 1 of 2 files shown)


ietf/api/serializers_rpc.py
```diff
@@ -673,6 +689,30 @@ def update(self, instance, validated_data):
                                 )
                             )
                         )
+            if updates is not omitted:
```
jennifer-richards, line 710: For both `updates` and `obsoletes`, we sometimes add `DocEvents` when they change. That's perhaps not needed for RFCs, where the document itself is an indication of when the older document became updated or obsoleted. @rjsparks ?

rjsparks: In the long run I think DocEvents on the older RFC when a newer one updates or obsoletes it would be a nice thing to have going forward just to make the history easier to read, but to keep things consistent we'd have to back-cast these events for the whole series, and that's not something I think we should bite off right now.

So, lets add making them a feature request and move on without adding them now.

jennifer-richards (changes requested):
You guard against unknown RFC numbers when adding relationships, which is great. I think it'd be better to fail if an unknown document is referenced, though. Otherwise datatracker may silently get out of sync. I suggest adding a validator (or maybe a transaction around the RelatedDocument manipulation that rolls back if any of the lookups fail)

ietf/api/serializers_rpc.py
```diff
@@ -600,6 +600,24 @@ class Meta:
...
 
+    def _validate_rfc_number_list(self, field_name, rfc_numbers):
+        """Raise ValidationError if any RFC numbers in the list don't exist."""
+        unknown = [
+            n for n in rfc_numbers
+            if not Document.objects.filter(rfc_number=n, type_id="rfc").exists()
+        ]
+        if unknown:
+            raise serializers.ValidationError(
+                {field_name: [f"Unknown RFC number: {n}" for n in unknown]}
```
jennifer-richards, line 611: This is creating a `ValidationError` that maps the field name to a `list` rather than a string. That might work, but I think it should be more like
```python
{field_name: f"Unknown RFC number(s): {", ".join(n for n in unknown)}"}
```

jennifer-richards (approved):

PR #10864, opened 2026-05-14
feat: disable editing DocumentAuthors when RfcAuthors exist

rjsparks:


ietf/doc/views_doc.py
```diff
@@ -1842,12 +1842,15 @@ def add_fields(self, form, index):
                 if fh in form.fields:
                     form.fields[fh].widget = forms.HiddenInput()
 
+    doc = get_object_or_404(Document, name=name)
+    if doc.rfcauthor_set.exists():
+        return HttpResponseForbidden("Contact the Rfc-Editor to change RFC Author information")
```
jennifer-richards: Nit, but just for general consistency:

```suggestion
        return HttpResponseForbidden("Contact the RFC Editor to change RFC Author information")
```

jennifer-richards (approved):

PR #10865, opened 2026-05-14
feat: only store real date changes in history

rudimatz:
avoid changes of published time recorded within one second


rudimatz:
keep as draft until bug is verified after cutover

PR #10873, opened 2026-05-15
fix: stop linking to pdfized txt at the rfc editor site

rjsparks:


ietf/doc/feeds.py
```diff
@@ -234,14 +234,6 @@ def item_extra_kwargs(self, item):
                             "is_format_of": self.item_link(item),
                         }
                     )
-            if item.rfc_number not in [571, 587]:
```
richsalz, line 237: of course, now I'm dying to know what is special about 571,587 :). Don't waste time answering me.

jennifer-richards: Obviously they're the 51st and 52nd Lesser Sexy Primes. https://oeis.org/A023201

jennifer-richards (approved):

ietf/doc/feeds.py
rjsparks: At some point in the past, the pdfizing tool did bad things (would hang IIRC) with these 1 page RFCs.

PR #10898, opened 2026-05-19
feat: create docevents for adding/removing obs/updates

rudimatz:
add log entries like this for obs and updates relations
image


rudimatz:
relates to https://github.com/ietf-tools/datatracker/pull/10851, which contains the main logic... this only adds DocEvents for the history tab in DT

rudimatz:
possibly OBE, will turn into draft

rudimatz:
see also https://github.com/ietf-tools/datatracker/pull/10851#pullrequestreview-4275587063 for post-MVP fix

PR #10935, opened 2026-05-26
fix: consistent counts at proceedings and proceedings/attendees

rjsparks:
This causes proceedings/attendees to show all attendees, not just those for whom the attendance is nomcom qualifying.

It refactors how attendance is calculated for proceedings, proceedings/attendees, and three_of_five_eligible9389

It fixes a bug (that had no consequence of previous nomcom choices) that counted hackathon only attendance towards nomcom eligibility.

I've independently checked that, except for the bug above which affected only two people, the results for three_of_five_eligible9389 remain unchanged for sets of 5 from [110...115] to [121...125].

Claude was used in creating this PR

### DIFF
--- a/ietf/meeting/models.py
+++ b/ietf/meeting/models.py
@@ -236,16 +236,24 @@ class Meeting(models.Model):
             document__states__slug='active', document__states__type_id='procmaterials'
         ).order_by('type__order')
 
-    def get_attendance(self):
-        """Get the meeting attendance from the Registrations
+    def get_attendees(self, sessions=None):
+        """Get the meeting attendees from the Registrations
 
-        Returns a NamedTuple with onsite and remote attributes. Returns None if the record is unavailable
-        for this meeting.
+        Returns a pair of sets containing Person objects for persons attending
+        the meeting either onsite or online (remote). Returns None if the record
+        is unavailable for this meeting.
+
+        sessions: optional queryset of Sessions to restrict which Attended records
+            count toward attendance. Defaults to all sessions at this meeting.
+            Note: restricting sessions does not affect the registration-based path
+            (checkedin / attended flags), only the Attended-record path.
         """
         number = self.get_number()
         if number is None or number < 110:
             return None
-        Attendance = namedtuple('Attendance', 'onsite remote')
+
+        if sessions is None:
+            sessions = self.session_set.all()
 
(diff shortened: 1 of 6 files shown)


ietf/meeting/views.py
```diff
@@ -4816,16 +4816,20 @@ def proceedings_attendees(request, num=None):
...
-        checked_in, attended = participants_for_meeting(meeting)
-        regs = list(Registration.objects.onsite().filter(meeting__number=num, checkedin=True))
-        onsite_count = len(regs)
-        regs += [
+        onsite, remote = meeting.get_attendees()
+        onsite_count = len(onsite)
+        remote_count = len(remote)
+        remote_pks = frozenset(p.pk for p in remote)
+
+        regs = list(
```
rpcross: Here registrations is being calculated separately and differently than stats so the totals may not match the table data. For example stats.onsite (from meeting.get_attendees().online) would include someone who registered onsite, did not checkin, and attended a session (Attended), whereas registrations would not.

rpcross (commented):
Possible proceedings/attendees/ view inconsistencies.

ietf/meeting/views.py
rjsparks: Ok - the reconstruction of Registration objects here is overly complicated anyhow. e1a84da makes it wholly dependent on the people returned by get_attendees.

ietf/meeting/models.py
```diff
@@ -254,41 +262,90 @@ def get_attendance(self):
...
 
-        attended_per_meeting_attended = (
-            Q(attended__session__meeting=self)
-            # Note that we are not filtering to plenary, wg, or rg sessions
-            # as we do for nomcom eligibility - if picking up a badge (see above)
-            # is good enough, just attending e.g. a training session is also good enough
-        )
         attendees_by_att = set(
-            Person.objects.filter(attended_per_meeting_attended).values_list(
+            Person.objects.filter(attended__session__in=sessions).values_list(
```
jennifer-richards, line 275: There's no guard that `sessions` are attached to this `Meeting`. May want to check that and fail with an assertion (or at least a logged error).

(You could also filter sessions to the current meeting, but I think a mix-up like this probably would indicate a more serious logic error somewhere.)

jennifer-richards (approved):
Approving. One minor comment inline, but not an essential change.

PR #10939, opened 2026-05-27
ci: pin xym until we can adapt to new required arguments

rjsparks:
Without the pin, we are getting:
```
'YangModuleExtractor.get_extracted_models() missing 2 required positional arguments: 'extract_semver' and 'semver_symlink'
```


rjsparks:
We should reapply #10127 as part of the steps taken to be able to unpin.
cc: @xorrkaz

xorrkaz:
Hmmm, let me see what I can do about that.  The YANG Semver stuff is my work.

jennifer-richards (approved):

PR #10951, opened 2026-05-29
feat: update rfc json

rjsparks:
RFC JSON files (`json/rfc{N}.json`) live in the `"rfc"` blob bucket. Purple generates the
initial file on publication and uploads it via `RfcPubFilesView`. After that, nothing updates
them. Datatracker must regenerate them — by combining data from its own DB (which purple keeps
current) and `errata.json` (produced by the errata system) — whenever any of three triggers fire.
* Publication of a new RFC affects metadata of other RFCs
* RFC Metadata is updated by purple
* an RFC gets an errata report (indicate by an entry in errata.json)

### DIFF
--- a/ietf/doc/utils_rfc_json.py
+++ b/ietf/doc/utils_rfc_json.py
@@ -1,5 +1,6 @@
 # Copyright The IETF Trust 2026, All Rights Reserved
 
+import datetime
 import json
 from pathlib import Path
 
@@ -7,6 +8,7 @@ from django.conf import settings
 from django.utils import timezone
 
 from ietf.doc.models import Document, RelatedDocument
+from ietf.name.models import StdLevelName
 from ietf.doc.storage_utils import exists_in_storage, store_bytes
 from ietf.sync.errata import errata_map_from_json, get_errata_data
 from ietf.sync.rfcindex import get_april1_rfc_numbers, get_publication_std_levels
@@ -42,6 +44,13 @@ def generate_rfc_json(rfc_number: int, *, pub_levels=None) -> None:
         log(f"generate_rfc_json: no RFC found for rfc_number={rfc_number}")
         return
 
+    if pub_levels is None:
+        try:
+            pub_levels = get_publication_std_levels()
+        except Exception as e:
+            log(f"generate_rfc_json: failed to get publication std levels: {e}")
+            return
+
     doc_id = f"RFC{rfc_number}"
 
     # draft name
@@ -66,13 +75,23 @@ def generate_rfc_json(rfc_number: int, *, pub_levels=None) -> None:
     # page_count
     page_count = str(rfc.pages) if rfc.pages is not None else ""
 
-    # pub_status from publication-std-levels.json in the red bucket
-    if pub_levels is None:
-        pub_levels = get_publication_std_levels()
-    pub_status = pub_levels[rfc_number].name.upper()
-
     # status: current std_level
     status = rfc.std_level.name.upper() if rfc.std_level else ""
+
+    # pub_status from publication-std-levels.json in the red bucket
+    # but guard against recent publication not having updated the bucket yet
+    pub_event = rfc.latest_event(type="published_rfc")
+    if rfc_number in pub_levels:
+        pub_status = pub_levels[rfc_number].name.upper()
+    else:
+        if (
+            pub_event is not None
+            and timezone.now() - pub_event.time < datetime.timedelta(days=2)
+        ):
+            pub_status = status
+        else:
+            log(f"Assuming an unknown publication status for rfc{rfc_number}")
+            pub_status = StdLevelName.objects.get(slug="unkn").name.upper()
 
     # source: adapted from errata system's display_source() logic
     stream_slug = rfc.stream.slug if rfc.stream else ""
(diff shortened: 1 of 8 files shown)


ietf/doc/utils_rfc_json.py
```diff
@@ -0,0 +1,197 @@
...
+        if exists_in_storage(kind="rfc", name=f"{ext}/rfc{rfc_number}.{ext}")
+    ]
+
+    # page_count
+    page_count = str(rfc.pages) if rfc.pages is not None else ""
+
+    # pub_status from publication-std-levels.json in the red bucket
+    if pub_levels is None:
+        pub_levels = get_publication_std_levels()
+    pub_status = pub_levels[rfc_number].name.upper()
```
jennifer-richards: Should perhaps handle the case where `rfc_number not in pub_levels`, although unless / until we retry on failure I'm not sure what to do about it other than fail with a clearer error message.

ietf/doc/utils_rfc_json.py
```diff
@@ -0,0 +1,197 @@
...
+    elif group_acronym not in ("none", ""):
+        source = group_acronym
+        if stream_slug == "ietf" and area_acronym:
+            source += f" ({area_acronym})"
+        elif stream_slug:
+            source += f" ({stream_slug})"
+    elif stream_slug:
+        source = "Legacy" if stream_slug == "legacy" else stream_slug.upper()
+    else:
+        source = ""
```
jennifer-richards, line 136: This differs substantially from the json production at purple. I believe this code is correct and purple's is incorrect for some non-IETF WG documents. We could fix purple to match, but long-term I think we should make datatracker responsible for producing these json files on initial publication as well as update. (The important things being that it's the same code generating the original and updated contents, and that it's the same place that the RFC index is generated)

(In the short term it _may_ be easier to fix purple rather than adjust the plumbing for document publication.)

@rudimatz fyi

ietf/doc/utils_rfc_json.py
```diff
@@ -0,0 +1,197 @@
...
+    if pub_levels is None:
+        pub_levels = get_publication_std_levels()
+    pub_status = pub_levels[rfc_number].name.upper()
+
+    # status: current std_level
+    status = rfc.std_level.name.upper() if rfc.std_level else ""
+
+    # source: adapted from errata system's display_source() logic
+    stream_slug = rfc.stream.slug if rfc.stream else ""
+    group_acronym = rfc.group.acronym if rfc.group else "none"
```
jennifer-richards: The use of `"none"` vs `""` here has a bit of a smell to it. It looks to me like it's differentiating the case where there's a `Group` with `acronym=""` (which winds up with `source = ""` below) and the case where there is no group. Is that a meaningful distinction to make?

jennifer-richards (commented):
A few comments inline. Generally looks good to me.

A high-level thing (touched on by the comment about `rfc_number not in pub_levels`) is that can fail semi-silently if there's a problem generating one of the json files. There are at least a couple probably transient situations that could cause this (the `pub_levels` issue, or a temporary failure accessing the red bucket are the ones I can think of). It might be worth retrying in those cases so we don't have to count on noticing the logged exception.

ietf/doc/utils_rfc_json.py
rjsparks: the guards here are almost superfluous as these cases are strongly meant never to occur, and much of the existing dt code would break if they did. Perhaps it would introduce less confusion to remove the guards here and look into making the the guarantees on stream an group elsewhere?

Or we could assert that the values aren't none (or otherwise falsey) instead.

jennifer-richards: I'd prefer the assertion - the code here leaves the reader wondering what they're not understanding.

(It's me, I'm the reader...)

ietf/doc/utils_rfc_json.py
```diff
@@ -66,18 +75,46 @@ def generate_rfc_json(rfc_number: int, *, pub_levels=None) -> None:
...
     # source: adapted from errata system's display_source() logic
-    stream_slug = rfc.stream.slug if rfc.stream else ""
-    group_acronym = rfc.group.acronym if rfc.group else "none"
-    area_acronym = rfc.group.parent.acronym if rfc.group and rfc.group.parent else ""
+    if rfc.stream is None or rfc.group is None:
+        # Basic expectations (should be constraints) on RFC Document objects
+        # have been violated.
+        assertion("rfc.stream is not None and rfc.group is not None")
+        log(
+            f"Malformed document object enoutered for rfc{rfc_number}. Aborting update of rfc{rfc_number}.json"
```
jennifer-richards: typo: should be `encountered`

ietf/doc/utils_rfc_json.py
```diff
@@ -66,18 +75,46 @@ def generate_rfc_json(rfc_number: int, *, pub_levels=None) -> None:
...
+        return
+    stream_slug = rfc.stream.slug
+    group_acronym = rfc.group.acronym
+
+    area_acronym = None
+    if stream_slug == "ietf":
+        if rfc.group.parent is None:
+            assertion("rfc.group.parent is not None")
+            log(
+                f"Malformed document object enoutered for rfc{rfc_number}. Aborting update of rfc{rfc_number}.json"
```
jennifer-richards: typo: again `encountered`

jennifer-richards (commented):
A couple typos - but this looks good

jennifer-richards (dismissed):

jennifer-richards (approved):

PR #10974, opened 2026-06-02
feat: Celery task to generate I-D BibXML

### LINKED ISSUE #10973: Add a Celery task to generate BibXML files for Internet-Drafts
### Description

Similar to #10830, Add a new celery task to generate BibXML files for I-Ds.

### Code of Conduct

---
Aggregated with agent assistance on 2026-08-13

kesara:
Fixes #10973

### DIFF
--- a/ietf/sync/bibxml.py
+++ b/ietf/sync/bibxml.py
@@ -191,7 +191,7 @@ def recreate_id_bibxml_by_draft_name(draft_name):
     for revision in reversed(doc.revisions_by_newrevisionevent()):
         doc_rev = doc.history_set.order_by("-time").filter(rev=revision).first()
         bibxml = get_id_bibxml(draft_name, doc_rev)
-        filename = f"bibxml-ids/reference.I-D.{draft_name}.xml"
+        filename = f"bibxml-ids/reference.I-D.{draft_name}-{revision}.xml"
         save_bibxml(bibxml, filename)
 
 
--- a/ietf/sync/bibxml.py
+++ b/ietf/sync/bibxml.py
@@ -81,6 +81,25 @@ def get_fyi_bibxml(fyi_number):
     return f"""<referencegroup anchor="FYI{fyi_number}" target="{fyi_link}">{rfc_bibxml}</referencegroup>"""
 
 
+def get_id_bibxml(draft_name, doc):
+    """Return BibXML entry for the given I-D doc"""
+    name = "-".join(draft_name.split("-", 2)[1:])
+    date = ""
+    if doc.is_dochistory():
+        latest_event = doc.latest_event(type="new_revision", rev=doc.rev)
+        if latest_event:
+            doc.pub_date = latest_event.time
+            date = doc.pub_date.strftime('<date day="%-d" month="%B" year="%Y"/>')
+    else:
+        date = doc.pub_date().strftime('<date day="%-d" month="%B" year="%Y"/>')
+    link = f"https://datatracker.ietf.org/doc/html/{draft_name}-{doc.rev}"
+    authors = ""
+    for author in doc.author_persons_or_names():
+        authors += f"""<author fullname={qa(author.person.name)} />"""
+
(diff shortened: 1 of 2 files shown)


ietf/sync/bibxml.py
```diff
@@ -81,6 +81,25 @@ def get_fyi_bibxml(fyi_number):
     return f"""<referencegroup anchor="FYI{fyi_number}" target="{fyi_link}">{rfc_bibxml}</referencegroup>"""
 
 
+def get_id_bibxml(draft_name, doc):
+    """Return BibXML entry for the given I-D doc"""
+    name = "-".join(draft_name.split("-", 2)[1:])
+    date = ""
+    if doc.is_dochistory():
```
jennifer-richards, line 88: I wonder if we should implement `DocHistory.pub_date()` so that you don't have to branch here

ietf/sync/bibxml.py
```diff
@@ -81,6 +81,25 @@ def get_fyi_bibxml(fyi_number):
...
+        latest_event = doc.latest_event(type="new_revision", rev=doc.rev)
+        if latest_event:
+            doc.pub_date = latest_event.time
+            date = doc.pub_date.strftime('<date day="%-d" month="%B" year="%Y"/>')
```
jennifer-richards, line 92: I recommend against pseudo-shadowing `pub_date` this way. It's not a true shadow since a `DocHistory` doesn't have a `pub_date()` method, but to a person reading the code it appears to be replacing the method.

I think it's just a temp variable and you don't actually need to attach it to the `DocHistory`.

ietf/sync/bibxml.py
```diff
@@ -156,3 +175,31 @@ def recreate_rfcsubseries_bibxml():
...
+    # draft BibXML for each revision
+    for revision in reversed(doc.revisions_by_newrevisionevent()):
+        doc_rev = doc.history_set.order_by("-time").filter(rev=revision).first()
+        bibxml = get_id_bibxml(draft_name, doc_rev)
+        filename = f"bibxml-ids/reference.I-D.{draft_name}-{revision}.xml"
+        save_bibxml(bibxml, filename)
+
+
+def recreate_id_bibxml():
+    """Creates BibXML for all Internet Drafts."""
```
jennifer-richards, line 199: I think we probably want to have a way to throttle this, run on subsets, etc. There are many drafts.

jennifer-richards (changes requested):
Major thing: I don't see a task, just helpers?

Couple comments inline. Some thought as to how we're going to run this is needed - I think we'll want more nuance than "all the drafts all at once."

ietf/sync/bibxml.py
kesara: Yeah, that'll make things easier. But this is the sort of same logic that has been used elsewhere in DT get revisions.

ietf/sync/bibxml.py
kesara: Yeah, happy to improve. I was using the same logic as in https://github.com/ietf-tools/datatracker/blob/1cccb088947cab50a5b8c837ed584f2396de2cdd/ietf/doc/views_doc.py#L1292-L1295

IMO there should be easier way to get the revisions.

ietf/sync/bibxml.py
kesara: I would imagine this to be a one time thing. So kind of hoping we run this manually as required because it's resource consuming.

jennifer-richards: Right, and that resource consumption is actually my worry. This is going to tie up a celery worker for the entire run, and if it fails part way through we have to start over.

Do you have a sense for how long it takes? If it's five minutes, then fine. If it's hours, then we want to be prepared to manage it in more detail

ietf/sync/bibxml.py
jennifer-richards: I suspect it's done there for use in a template (where `{{ doc.pub_date }}` will call the callable if necessary.

PR #10976, opened 2026-06-02
feat: self-serve queries for inputs to reporting and survey purposes

rjsparks:
This captures queries that have been run by hand for a few years to support further reporting and survey work.
The self-serve view and improvements to the tests in de600c6 were created using Claude.

cc: @ghwood, @A-Gohil


ietf/utils/tests_reports.py
```diff
@@ -0,0 +1,189 @@
+# Copyright The IETF Trust 2023, All Rights Reserved
```
jennifer-richards: 🎶 _Let's do the Time Warp again_ 🎶 
```suggestion
# Copyright The IETF Trust 2026, All Rights Reserved
```

jennifer-richards (changes requested):
Trivial nit

jennifer-richards (approved):

PR #11039, opened 2026-06-14
feat: add index page for past plenaries

### LINKED ISSUE #10919: Build the past-IETF-meetings index rendering number, dates, host(s), proceedings and survey-results links
### Description

This is different than 

https://datatracker.ietf.org/meeting/past

in that it:

* only shows IETF plenary meetings
* replicates closely the page currently at: https://www.ietf.org/meeting/past/ by, for each IETF plenary meeting, showing the meeting number as a heading e.g. "IETF NNN" and under each meeting number:
     * Dates of the meeting
     * Host(s) of the meeting (if there is one)
     * "Proceedings" linked to the proceedings for that meeting
     * "Survey results" linked to the survey results for that meeting (if they exist, usually hosted on www.ietf.org )

### Code of Conduct

---

noralmaani:
fixes [[#10919]](https://github.com/ietf-tools/datatracker/issues/10919)

Screenshot 2026-07-14 at 16 40 52

### DIFF
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -4487,6 +4487,16 @@ def past(request):
     return render(request, 'meeting/past.html', {
                   'meetings': meetings,
                   })
+    
+def past_ietf(request):
+    '''List of past IETF planetart meetings'''
+    today = timezone.now()
+    
+    meetings_ietf = data_for_meetings_overview(Meeting.objects.filter(type_id='ietf').filter(date__lte=today).order_by('-date'))
+    
+    return render(request, 'meeting/past_ietf.html', {
+                'meetings_ietf': meetings_ietf,
+                })
 
 def upcoming(request):
     '''List of upcoming meetings'''
--- a/ietf/meeting/views.py
+++ b/ietf/meeting/views.py
@@ -72,6 +72,7 @@ from ietf.meeting.models import Meeting, Session, Schedule, FloorPlan, \
 from ..blobdb.models import ResolvedMaterial
 from ietf.meeting.models import ImportantDate, SessionStatusName, SchedulingEvent, SchedTimeSessAssignment, Room, TimeSlotTypeName
 from ietf.meeting.models import Registration
+from ietf.meeting.models import MeetingSurvey
 from ietf.meeting.forms import ( CustomDurationField, SwapDaysForm, SwapTimeslotsForm, ImportMinutesForm,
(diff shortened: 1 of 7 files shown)


ietf/meeting/views.py
```diff
@@ -4487,6 +4487,16 @@ def past(request):
     return render(request, 'meeting/past.html', {
                   'meetings': meetings,
                   })
+    
+def past_ietf(request):
+    '''List of past IETF planetart meetings'''
```
jennifer-richards: typo: plenart -> plenary

jennifer-richards (commented):
This will need some additional work. Issue #10919 requests a list format closer to https://www.ietf.org/meeting/past/ rather than a table. Most of the data there are already available (see `ietf/templates/meeting/proceedings/title.html` to see how they can be extracted). ~~Note that the location we have will generally only be a city, not city + country.~~ (edit: I was wrong, meetings have a country field, we just don't currently show it on the proceedings page where I was looking)

The survey results link asked for in the issue will need some thought. It will probably require model adjustments to store the URL when it becomes available, because it's not available for all meetings, generally not available until some time after the end of a meeting, and it looks like the filename in the URL is not perfectly consistent.

A minor detail - the name for the page in the menu will probably need adjustment to clarify the distinction between "Past meetings" and "Past IETF meetings." My quick reaction is to change the existing one to "All past meetings" and the new to "Past plenary meetings", but I'll see what others think.

rjsparks:
> change the existing one to "All past meetings" and the new to "Past plenary meetings"

wfm

noralmaani:
> The survey results link asked for in the issue will need some thought. It will probably require model adjustments to store the URL when it becomes available, because it's not available for all meetings, generally not available until some time after the end of a meeting, and it looks like the filename in the URL is not perfectly consistent.

I tried to scrape the URL for meeting surveys (extracted all but one survey result hosted on qualtrics for ietf-120) . Let me know if there's a better way!

rjsparks:
> I tried to scrape the URL

Identifying the URL to use is an offline activity, not something that should be done by code in this PR.
The point of Jennifer's comment is that there will need to be some _design_ on the models and workflows expected for managing these things as they are created in the future.
We also need to have some conversations with Greg and Jay about how the survey results themselves should really be persisted.
So, until we can provide more inputs, I don't think you can really make more progress on this PR.

PR #11040, opened 2026-06-14
fix: scrub control characters from object values headed towards lxml

rjsparks:


ietf/api/__init__.py
```diff
@@ -177,27 +177,33 @@ def dehydrate(self, bundle, for_list=True):
...
         simple_data = super().to_simple(data, options)
-        if (
-            options.get(self.OPTION_ESCAPE_NULLS, False) 
-            and isinstance(simple_data, str)
-        ):
-            # replace nulls with unicode "symbol for null character", \u2400
-            simple_data = simple_data.replace("\x00", "\u2400")
+        if options.get(self.OPTION_ESCAPE_XML_INVALID, False) and isinstance(simple_data, str):
+            # Replace control chars invalid in XML 1.0 with their Unicode
+            # control pictures (U+2400\u2013U+241F) so lxml won't reject the string.
```
jennifer-richards: I think that `\u2013` should just be a hyphen? (Or an en-dash if you want to be true to the code point)

jennifer-richards (changes requested):
Trivial nit in a comment

jennifer-richards (approved):

PR #11094, opened 2026-06-25
fix(bibxml): Add subseries info

kesara:


ietf/sync/bibxml.py
```diff
@@ -27,13 +27,14 @@ def save_to_bucket(filename: str, content: str | bytes):
     log(f"Saved {bucket_path} in bibxml_bucket storage")
 
 
-def get_rfc_bibxml(rfc_number):
+def get_rfc_bibxml(rfc_number, subseries=None):
```
jennifer-richards: I'm a little wary of passing a `dict` like this - it's hard to understand what keys are needed. You could use a `datatclass` or add comments to document it (or type hints, though I've found that tricky to do effectively and succinctly with dicts).

Or, since you actually have a `Document` for the subseries in the caller code, you could just pass that document in as the parameter here. You can get the type from `doc.type` and parse the number here. That avoids needing to introduce a new type to pass the description around.

ietf/sync/bibxml.py
jennifer-richards: I should have noticed this before, but here (and for other subseries types), you should filter on `type_id="bcp"` in addition to the name filter. It should be redundant but would be better practice.

jennifer-richards (changes requested):
Minor refactor requested

ietf/sync/bibxml.py
kesara: a22a6234de5b4989a00e8b7287bcc21746d63e87 addresses this.

ietf/sync/bibxml.py
kesara: Refactored with 730a540c2658a626936574a89e0f396802b99f77.

jennifer-richards (approved):
nice

PR #11119, opened 2026-06-30
fix: Use primary address for ballot email from for non ietf-stream docs

### LINKED ISSUE #11114: Ballot "from" address needs to be more purposely chosen
Currently at https://github.com/ietf-tools/datatracker/blob/a89f27bcf5bb7d45119ae2cad5b6b553f91888c9/ietf/doc/views_ballot.py#L428, when someone takes a ballot position on a document from _ANY STREAM_ the email sent tries to find the email associated with the person's AD role (which they might not have ever had). This leads to a bad outcome for ex-ADs.

Instead, the from address should either stop looking at Roles and just use the primary address of the sender (my preference, but YMMV), or try to pay attention to role addresses for approval roles in the stream of the document being balloted.

rjsparks:
Fixes #11114

Note that this changes the default behavior of the Person factory to generate a primary Email object for the Person.

### DIFF
--- a/ietf/person/factories.py
+++ b/ietf/person/factories.py
@@ -93,7 +93,7 @@ class PersonFactory(factory.django.DjangoModelFactory):
     def default_emails(obj, create, extracted, **kwargs): # pylint: disable=no-self-argument
         if extracted is None:
             extracted = True
-        if create and extracted:           
+        if create and extracted:
             make_email = getattr(EmailFactory, 'create' if create else 'build')
             make_email(person=obj, address=obj.user.email, primary=True, **kwargs)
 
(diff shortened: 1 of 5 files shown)


ietf/person/factories.py
```diff
@@ -93,9 +93,9 @@ def default_aliases(obj, create, extracted, **kwargs): # pylint: disable=no-self
     def default_emails(obj, create, extracted, **kwargs): # pylint: disable=no-self-argument
         if extracted is None:
             extracted = True
-        if create and extracted:
+        if create and extracted:           
             make_email = getattr(EmailFactory, 'create' if create else 'build')
-            make_email(person=obj, address=obj.user.email)
+            make_email(person=obj, address=obj.user.email, primary=True, **kwargs)
```
jennifer-richards, line 98: nit: the `if create...` is always true since this is in a block where `create and extracted` is true, so this could just be `EmailFactory.create()` and the previous line could be removed

jennifer-richards (approved):

PR #11172, opened 2026-07-08
fix: prevent overlap in html page breaks

### LINKED ISSUE #11118: rendering of rfc page breaks ruled line looks lousy
### Describe the issue

try visiting https://datatracker.ietf.org/doc/html/rfc3597 (first w/o javascript enabled, but later with it).

The no site status message bit is odd:
Image
I can't tell what it's trying to do, and it _felt_ like it impacted the rest of the rendering, but...

Here's what I saw for page 1:
Image
Here's what someone else saw (probably w/ js enabled):
Image
When I enabled js:
Image

### Code of Conduct

jennifer-richards:
Fixes #11118

### DIFF
--- a/ietf/static/css/document_html.scss
+++ b/ietf/static/css/document_html.scss
@@ -205,10 +205,6 @@
         display: inline;
     }
 
-    .newpage {
-        margin-top: -1.25em;
-    }
-
 }
 
 tbody.meta tr {


jennifer-richards:
Changes the `<hr>` at page breaks to render like this instead of with the overlap shown in #11118 
image

rjsparks (approved):

jennifer-richards:
For the record: the screenshot above looks ~ the same on Firefox, Chrome, and Safari in a quick check.

jsoref:
This is what I see today in Chrome:
image

Am I doing something wrong?

jennifer-richards:
> Am I doing something wrong?

No, but you tested before the fix was released. The GH issue closes when the fix is merged to main, but it's usually a few days before it goes live. It was released yesterday afternoon North American time, so you should see something different now.

Issue #11167, opened 2026-07-07
Some dropdown SELECT inputs for ADs do not sort by I18N


evyncke:
### Describe the issue

Some dropdown SELECT inputs, e.g., https://datatracker.ietf.org/doc/draft-ietf-ipsecme-ikev2-pqc-auth/edit/ad/, list all ADs by without taking care of the I18N, e.g., Éric Vyncke is listed the last one while it should be between Deb and Gorry.

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

---
Aggregated with agent assistance on 2026-08-07

evyncke:
Please assign to me so that I will take care of it at the Code Sprint

evyncke:
To my surprise, it works fine when tried in the test environment... I.e., it is probably a Postgres setting on the production DB that is different than the settings_local.py or Postgres setting in the Docker...

While it can be fixed in the ORM queryset, it is probably better to fix it globally in the production DB.

PR #11234, opened 2026-07-18
fix: Use Django Collate() when listing all ADs

evyncke:
To address #11167 (on production DT, "Éric" does not appear between "Deb" and "Gorry" but is listed at the end), there are *TWO* ways:

1. change the production Postgres collation to support en_US.utf8 (like in the Docker Postgres DB)
2. apply this PR

Really unsure what is the best way forward.


jennifer-richards (dismissed):
Seems reasonable to me until we can fix this more cleanly. I'll create an issue for it.

jennifer-richards (approved):

PR #11241, opened 2026-07-18
fix: Point to pending interims in meetings list (#11069)

### LINKED ISSUE #10104: Pending interim meetings are hard to find
### Description

The place I would naturally look for pending interim meetings in order to approve them is on the WG's meetings page. However, nothing there shows unapproved meetings. It turns out they're at https://datatracker.ietf.org/meeting/interim/pending. That page is also not in the Datatracker Meeting tab for general navigation.

Could we consider either showing pending meetings on the group's page (to the AD or chairs, obviously), or at least putting a link to the pending meeting page from there?

### Code of Conduct

### LINKED ISSUE #11069: There is no way to see interim meetings in the pending or announce queue
### Description

When I request an interim meeting, it often goes in the pending or announce queue, sometimes for days.

There is no obvious way in the interface to view these queues without requesting another interim meetings. So now I have a duplicate request .

Relatedly, I have no idea why an automatically approved request has to sit in the queue for a few days. As I forgot to submit the request until the Friday before a Monday meeting, this is likely to result in a failed meeting for no good reason 

### Code of Conduct

russhousley:
Fix: Point to pending interim meeting requests in meetings list (#11069)

Closes #10104
Closes #11069

### DIFF
--- a/ietf/group/tests_info.py
+++ b/ietf/group/tests_info.py
@@ -2115,23 +2115,61 @@ class MeetingInfoTests(TestCase):
 
 
 class PendingInterimMeetingTests(TestCase):
+    """Tests for the pending-interim warning on a group's meetings list.
 
-    def test_pending_interim_meeting(self):
+    The meetings page shows a ``#pending_warning`` banner when the group has an
+    interim meeting that is either awaiting approval (session status ``apprw``)
+    or approved but not yet announced (session status ``scheda``). See
+    ietf.meeting.helpers.has_pending_interim.
+    """
+
+    def _meetings_page(self, group):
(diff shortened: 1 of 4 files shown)


jennifer-richards (approved):
Approving.

N.b., the helper you're using, `data_for_meetings_overview()`, is inefficient - each call is taking around 150 ms on my laptop, so this will be a bit slow as-is on production. That's an existing problem, though, so we'll take on refactoring to speed that up as a separate enhancement.

jennifer-richards:
#11279 is the issue to track the refactoring

PR #11245, opened 2026-07-18
fix: add email validation to secr announcement form

### LINKED ISSUE #11180: Secr announcement tool silently sends no message if addressee fields are badly formatted
### Describe the issue

The announcement tool in the secr app accepts messages whose addressee fields (To / Cc) contain badly formatted address lists. When `send_smtp()` is called to deliver the message, it passes these through `email.utils.getaddresses()` to validate and parse them. When an invalid address is present, that method returns `[("", "")]` which is treated as an empty addressee list. This causes datatracker to log a message that no message was sent, but no other feedback is made. No feedback at all is available to the user except that their message is not delivered.

This comes up, e.g., if the user mistakenly uses `;` rather than `,` to separate otherwise valid email addresses.

These form fields should be validated using the same (or equivalent) logic as `send_smtp()` so that messages will not be accepted if they will not be sent.

### Code of Conduct

noralmaani:
fixes [#11180](https://github.com/ietf-tools/datatracker/issues/11180#event-28158027013)

### DIFF
--- a/ietf/secr/announcement/tests.py
+++ b/ietf/secr/announcement/tests.py
@@ -100,7 +100,7 @@ class SubmitAnnouncementCase(TestCase):
             "nomcom": nomcom.pk,
             "to": "Other...",
             "to_custom": "phil@example.com",
-            "cc": "lizz@example.com, test@example.com",
+            "cc": "lizz@example.com, no-brackets@example.com",
             "frm": "IETF Secretariat &lt;ietf-secretariat@ietf.org&gt;",
             "reply_to": "secretariat@ietf.org",
             "subject": "Test Subject",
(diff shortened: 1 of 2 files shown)


ietf/secr/announcement/tests.py
```diff
@@ -112,7 +113,29 @@ def test_valid_submit(self):
         self.assertRedirects(response, url)
         self.assertEqual(len(outbox), 1)
         self.assertEqual(outbox[0]["subject"], "Test Subject")
-        self.assertEqual(outbox[0]["to"], "<phil@example.com>")
+        self.assertEqual(outbox[0]["to"], "<phil@example.com>, <lizz@example.com>, <test@example.com>")
```
jennifer-richards: Should make sure this accepts addresses like `no-brackets@example.com` and `Named Person <named@example.com>` as well

ietf/secr/announcement/tests.py
```diff
@@ -112,7 +113,29 @@ def test_valid_submit(self):
...
+        post_data = {
+                "nomcom": nomcom.pk,
+                "to": "Other...",
+                "to_custom": "phil@example.com",
+                "cc": "lizz@example.com, invalid_email@example",
+                "frm": "IETF Secretariat &lt;ietf-secretariat@ietf.org&gt;",
+                "reply_to": "secretariat@ietf.org",
+                "subject": "Test Subject",
+                "body": "This is a test.",
+            }
```
jennifer-richards, line 157: Also need test cases for
* both the `to_custom` and `cc` fields are empty (i.e., no addresses at all)
* `to_custom` field has a bad list of otherwise valid addresses (e.g., `lizz@example.com; valid_email@example.com`)
* same, but for `cc` field

jennifer-richards (changes requested):
A few more test cases are needed (and if these uncover problems, please fix)

ietf/secr/announcement/tests.py
noralmaani: > and `Named Person <named@example.com>` as well

Re this point: the [MultiEmail field](https://github.com/ietf-tools/datatracker/blob/main/ietf/utils/fields.py) passes django's email validator to the form which does not accept this format. One option may be to remove this customisation for the "to_custom" field (like is done for the "cc" field), and then format the emails so that they are passed to the validator safely then submitted. Otherwise, leaving it as is would send the form error.

jennifer-richards: Yuck. Updating/replacing that field will be a separate project. Thanks.

jennifer-richards (approved):

Issue #10582, opened 2026-03-19
Proposing slides for past sessions should not be possible


jennifer-richards:
### Description

On the meeting session page for a group (e.g., https://datatracker.ietf.org/meeting/125/session/moq), a non-chair user should not be able to propose slides for a session that has already passed.

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

---
Aggregated with agent assistance on 2026-08-10

jennifer-richards:
Was fixed by #11258

PR #11258, opened 2026-07-20
fix: prevent slide upload by participants after session (#10582)

jimfenton:
Restricts slide submission by other than those that can manage slides after the session has ended.

Does not include tests for this, and breaks two existing tests:

======================================================================
FAIL: test_propose_session_slides (ietf.meeting.tests_views.MaterialsTests.test_propose_session_slides)
----------------------------------------------------------------------
  File "/workspace/ietf/meeting/tests_views.py", line 7077, in test_propose_session_slides
    self.assertTrue(q('.proposeslides'))
AssertionError: [] is not true

======================================================================
FAIL: test_submit_and_approve_multiple_versions (ietf.meeting.tests_views.MaterialsTests.test_submit_and_approve_multiple_versions)
----------------------------------------------------------------------
  File "/home/dev/.local/lib/python3.12/site-packages/django/test/utils.py", line 461, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/unittest/mock.py", line 1396, in patched
    return func(*newargs, **newkeywargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/ietf/meeting/tests_views.py", line 7271, in test_submit_and_approve_multiple_versions
    self.assertEqual(r.status_code, 302)
AssertionError: 403 != 302

Apparently this was because the sessions created for testing did not specify a time. I'm surprised this didn't fail before, because the submission would probably have violated the grace time after which Secretariat action is needed. I tried but didn't manage to succeed in fixing these test cases by adding a session time to the SessionFactory calls.


jimfenton:
I found another "Upload slides" button that isn't covered by this fix yet (materials/views.py line 1937). Will add.

ietf/meeting/views.py
```diff
@@ -1931,7 +1931,9 @@ def api_get_session_materials(request, session_id=None):
 
     minutes = session.minutes()
     slides_actions = []
-    if can_manage_session_materials(request.user, session.group, session) or not session.is_material_submission_cutoff():
+    if has_role(request.user, "Secretariat") \
+            or (not session.is_material_submission_cutoff() and session.can_manage_materials(request.user)) \
+            or not session.is_past():
         slides_actions.append(
```
jennifer-richards:
```suggestion
    if (
        has_role(request.user, "Secretariat")
        or (not session.is_material_submission_cutoff() and session.can_manage_materials(request.user))
        or not session.is_past()
    ):
        slides_actions.append(
```

style nit

ietf/templates/meeting/session_details_panel.html
```diff
@@ -185,7 +185,7 @@ <h3 class="mt-4">Slides</h3>
                href="{% url 'ietf.meeting.views.upload_session_slides' session_id=session.pk num=session.meeting.number %}">
                 Upload new slides
             </a>
-        {% elif request.user.is_authenticated and not session.is_material_submission_cutoff %}
+        {% elif request.user.is_authenticated and not session.is_past %}
```
jennifer-richards, line 188: Should this be `...and not session.is_past and not session.is_material_submission_cutoff`? I.e., is this going to undesirably start allowing submissions between the material submission cutoff and the end of the session?

jennifer-richards (changes requested):
Style nit + a question

ietf/templates/meeting/session_details_panel.html
jimfenton: It was my understanding that the material submission cutoff was after the meeting, to lock down the content for proceedings and such. This was partly based on what I saw in the test cases. So is_past would be a tighter restriction. If my understanding was incorrect, this would need to change.

jennifer-richards: Ah, you're likely right - I was thinking of draft submission deadline.

jennifer-richards (approved):
Fixed the failing tests, added missing test coverage, and adjusted to handle a corner case.

PR #11288, opened 2026-07-23
fix: migrate missing registration records

rpcross:
For meetings before 100 the legacy stats.MeetingRegistration system allowed
    several registrations under one email with different names. The migration to
    meeting.Registration keyed on (meeting, email) and kept only one, so the
    extra legacy records - which are genuine registrations - were lost. This adds
    a Registration for any legacy (meeting, email, name) with no matching row.

    The migration also piled the extra legacy rows' tickets onto that single
    surviving Registration instead of splitting them across records. For each
    such duplicate email this trims the surviving record back to a single ticket
    (the recreated registrations get their own tickets above).

### DIFF
--- a/ietf/meeting/utils.py
+++ b/ietf/meeting/utils.py
@@ -1852,5 +1852,100 @@ def process_single_registration(reg_data, meeting):
     return registration, action_taken
 
 
+def fix_missing_registrations(meeting=None):
+    """Recreate legacy registrations dropped by the meeting.Registration migration.
+
+    For meetings before 100 the legacy stats.MeetingRegistration system allowed
+    several registrations under one email with different names. The migration to
+    meeting.Registration keyed on (meeting, email) and kept only one, so the
+    extra legacy records - which are genuine registrations - were lost. This adds
+    a Registration for any legacy (meeting, email, name) with no matching row.
+
+    The migration also piled the extra legacy rows' tickets onto that single
+    surviving Registration instead of splitting them across records. For each
+    such duplicate email this trims the surviving record back to a single ticket
+    (the recreated registrations get their own tickets above).
+
+    If meeting (a meeting number) is given, only that meeting is checked;
+    otherwise all of meetings 72 through 99 are checked.
+
+    Returns (created, removed): the number of Registration records created and
+    the number of surplus tickets removed.
+    """
+    # import here to avoid a circular import with ietf.stats.models
+    from ietf.stats.models import MeetingRegistration
+
+    def norm(value):
+        return (value or "").strip().lower()
+
+    numbers = [int(meeting)] if meeting is not None else list(range(72, 100))
+    meeting_ids = [
+        m.pk
+        for m in Meeting.objects.filter(type="ietf")
+        if m.number.isdigit() and int(m.number) in numbers
+    ]
+
+    # names already present in the new table, keyed by (meeting, email)
+    existing = defaultdict(set)
+    for meeting_id, email, first, last in Registration.objects.filter(
+        meeting_id__in=meeting_ids
+    ).values_list("meeting_id", "email", "first_name", "last_name"):
+        existing[(meeting_id, norm(email))].add((norm(first), norm(last)))
+
+    # Registrations created during this run, keyed by (meeting, email, name), so
+    # repeated legacy rows for one attendee add tickets instead of duplicates.
+    added = {}
+    created = 0
+    collision_keys = set()  # (meeting, email) that had duplicate-email records
+    for mr in MeetingRegistration.objects.filter(meeting_id__in=meeting_ids):
+        key = (mr.meeting_id, norm(mr.email))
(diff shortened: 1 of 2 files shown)


ietf/meeting/utils.py
```diff
@@ -1852,5 +1852,216 @@ def process_single_registration(reg_data, meeting):
...
+    from ietf.stats.models import MeetingRegistration
+
+    def norm(value):
+        return (value or "").strip().lower()
+
+    numbers = [int(meeting)] if meeting is not None else list(range(72, 100))
+    meeting_ids = [
+        m.pk
+        for m in Meeting.objects.filter(type="ietf")
+        if m.number.isdigit() and int(m.number) in numbers
```
jennifer-richards, line 1885: Mostly moot given the smallish number of meetings of type `"ietf"`, but it'd be better practice to turn this around:

```python
meetings = Meeting.objects.filter(type="ietf", number__in=[str(num) for num in numbers])
```

I won't insist, though. If you do this, you'd need to change your `meeting_id__in` filters later to `meeting__in` to work with the queryset.

ietf/meeting/utils.py
```diff
@@ -1852,5 +1852,216 @@ def process_single_registration(reg_data, meeting):
...
+    collision_keys = set()  # (meeting, email) that had duplicate-email records
+    for mr in MeetingRegistration.objects.filter(meeting_id__in=meeting_ids):
+        key = (mr.meeting_id, norm(mr.email))
+        name = (norm(mr.first_name), norm(mr.last_name))
+        if name in existing[key]:
+            continue
+        collision_keys.add(key)
+        reg = added.get(key + name)
+        if reg is None:
+            reg = added[key + name] = Registration.objects.create(
```
jennifer-richards, line 1908: Style nit: the `a = b = ...` syntax can be confusing and is relatively uncommon in Python. Consider splitting into two statements.

ietf/meeting/utils.py
```diff
@@ -1852,5 +1852,216 @@ def process_single_registration(reg_data, meeting):
...
+    surviving = defaultdict(list)
+    for reg in Registration.objects.filter(meeting_id__in=meeting_ids).exclude(
+        pk__in=created_ids
+    ):
+        surviving[(reg.meeting_id, norm(reg.email))].append(reg)
+
+    removed = 0
+    for key in collision_keys:
+        for reg in surviving.get(key, []):
+            extra = list(reg.tickets.order_by("id")[1:])
```
jennifer-richards, line 1940: Is the first ticket always the correct one (correct attendance type / ticket type)? Is there always only one to keep?

ietf/meeting/utils.py
```diff
@@ -1852,5 +1852,216 @@ def process_single_registration(reg_data, meeting):
...
+    legacy = {}
+    for meeting_id, email, affiliation, attended, checkedin in (
+        MeetingRegistration.objects.filter(meeting_id__in=meeting_ids).values_list(
+            "meeting_id", "email", "affiliation", "attended", "checkedin"
+        )
+    ):
+        key = (meeting_id, norm(email))
+        agg = legacy.get(key)
+        if agg is None:
+            agg = legacy[key] = {
```
jennifer-richards, line 2003: again - `a = b = ...` style nit

jennifer-richards (commented):
Couple style nits to take or leave, plus one question

ietf/meeting/utils.py
rpcross: Fixed

ietf/meeting/utils.py
rpcross: Fixed

ietf/meeting/utils.py
rpcross: Fixed

ietf/meeting/utils.py
rpcross: Yes, prior to 100 attendance type and ticket type are unknown. Another cleanup might be to change attendance type to onsite for all those meetings where it's all we had.

jennifer-richards (approved):

ietf/meeting/utils.py
```diff
@@ -1852,5 +1852,214 @@ def process_single_registration(reg_data, meeting):
     return registration, action_taken
 
 
+def fix_missing_registrations(meeting=None):
```
jennifer-richards: Can you add a comment `# pragma: no cover` to the fn definition line? (see elsewhere in the codebase for examples). That should let the code coverage test pass.

You could also test this method, but if you've tested this one-time piece of code adequately out of band I'm ok letting it in

jennifer-richards (changes requested):
Sorry, one more adjustment requested

ietf/meeting/utils.py
rpcross: I've run this multiple times locally. Added no cover

jennifer-richards (approved):

PR #11308, opened 2026-07-28
fix(bibxml): Derive orgs

kesara:
This assumes author is an organization when `titlepage_name` is a single word.
This works well for cases like [RFC 2077](https://bib.ietf.org/public/rfc/bibxml/reference.RFC.2077.xml) to recreate the existing BibXML.
But there are still data issues in RfcAuthor data set.
For example,
BibXML for [RFC 3177](https://bib.ietf.org/public/rfc/bibxml/reference.RFC.3177.xml) has
```
<author>
<organization abbrev="IAB">Internet Architecture Board</organization>
</author>
<author>
<organization abbrev="IESG">Internet Engineering Steering Group</organization>
</author>
```
While `RfcAuthor` model only provides abbreviation:
```
>>> for author in Document.objects.filter(rfc_number=3177)[0].rfcauthor_set.all():
...     print(author.__dict__)
...
{'_state': <django.db.models.base.ModelState object at 0xf005a14e8ce0>, 'id': 8045, 'document_id': 125142, 'titlepage_name': 'IAB', 'is_editor': False, 'person_id': 104167, 'affiliation': 'IAB', 'country': '', 'order': 1}
{'_state': <django.db.models.base.ModelState object at 0xf005a14e8b00>, 'id': 8046, 'document_id': 125142, 'titlepage_name': 'IESG', 'is_editor': False, 'person_id': None, 'affiliation': 'IESG', 'country': '', 'order': 2}
```

Relaton gem has achieved this by maintaining a list of orgs[^1].

[^1]: https://github.com/relaton/relaton-ietf/blob/main/lib/relaton_ietf/bibxml_parser.rb#L70-L109

### DIFF
--- a/ietf/sync/bibxml.py
+++ b/ietf/sync/bibxml.py
@@ -91,7 +91,6 @@ ORG_LOOKUP = {
     "The Internet Society": (None, "The Internet Society"),
     "IAB Advisory Committee": (None, "IAB Advisory Committee"),
     "ISOC Board of Trustees": (None, "ISOC Board of Trustees"),
-    "Mitra": (None, "Mitra"),
     "RFC Editor, et al.": (None, "RFC Editor, et al."),
 }
 
--- a/ietf/sync/bibxml.py
+++ b/ietf/sync/bibxml.py
@@ -37,10 +37,16 @@ def get_rfc_bibxml(rfc):
     subseries_info = ""
 
     for author in rfc.rfcauthor_set.all():
-        if author.is_editor:
-            author_entry = f"""<author fullname={qa(author.titlepage_name)} surname={qa(author.titlepage_name.split(".")[-1].strip())} role="editor"/>"""
-        else:
-            author_entry = f"""<author fullname={qa(author.titlepage_name)} surname={qa(author.titlepage_name.split(".")[-1].strip())}/>"""
+        try:
+            initials, surname = author.titlepage_name.split(maxsplit=1)
+            if author.is_editor:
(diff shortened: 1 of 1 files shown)


kesara:
0835bab48deb228efca75d075cea3ed023e28d19 commit introduces a lookup table based on relaton-ietf gem[^1].

[^1]: https://github.com/relaton/relaton-ietf/blob/main/lib/relaton_ietf/bibxml_parser.rb#L70-L109

ietf/sync/bibxml.py
```diff
@@ -1,18 +1,100 @@
...
+        None,
+        "Information Sciences Institute University of Southern California",
+    ),
+    "IAB and IESG": (None, "IAB and IESG"),
+    "RARE WG-MSG Task Force 88": (None, "RARE WG-MSG Task Force 88"),
+    "KOI8-U Working Group": (None, "KOI8-U Working Group"),
+    "The Internet Society": (None, "The Internet Society"),
+    "IAB Advisory Committee": (None, "IAB Advisory Committee"),
+    "ISOC Board of Trustees": (None, "ISOC Board of Trustees"),
+    "Mitra": (None, "Mitra"),
```
kesara: Should we remove this? See RFC 2077

rjsparks: yes

kesara: Removed with 29012e7f1ea763a958f8d4779b31a78c200b8e22

PR #11340, opened 2026-07-31
fix: handle exception for API filter values that fail below tastypie

rjsparks:
Filter values reach the ORM with very little validation, and some of them only fail after tastypie has stopped looking at them - either while the query is being built or once it runs. Those surfaced as unhandled exceptions.

Add guards at both layers, in the ModelResource that every datatracker resource subclasses:

dispatch() converts DataError into a bad request, covering values that are only rejected once the query reaches the database. Only DataError is treated this way, as the DBAPI error for a problem with the data in the query. OperationalError, ProgrammingError and InternalError indicate a broken database or a bug of ours, and still raise and report. The database's message is logged rather than returned, since it can quote the offending value into a response body that is
not escaped.


ietf/api/__init__.py
```diff
@@ -60,11 +93,11 @@ def generate_cache_key(self, *args, **kwargs):
...
-        #smooshed = ["%s=%s" % (key, value) for key, value in kwargs.items()]
+        # smooshed = ["%s=%s" % (key, value) for key, value in kwargs.items()]
         smooshed = urlencode(kwargs)
 
         # Use a list plus a ``.join()`` because it's faster than concatenation.
-        return "%s:%s:%s:%s" % (self._meta.api_name, self._meta.resource_name, ':'.join(args), smooshed)
+        return f"{self._meta.api_name}:{self._meta.resource_name}:{':'.join(args)}:{smooshed}"
 
     def _z_aware_fromisoformat(self, value: str) -> datetime.datetime:
         """datetime.datetime.fromisoformat replacement that works with python < 3.11"""
```
jennifer-richards, line 103: Guess we can refactor this helper since we're on py3.12... for another day, though.

jennifer-richards (dismissed):

ietf/api/tests.py
```diff
@@ -58,7 +58,8 @@ class CustomApiTests(TestCase):
     settings_temp_path_overrides = TestCase.settings_temp_path_overrides + ['AGENDA_PATH']
 
     def test_api_help_page(self):
-        url = urlreverse('ietf.api.views.api_help')
+        url = urlreverse('
+                         .api_help')
```
jennifer-richards:
```suggestion
        url = urlreverse('ietf.api.views.api_help')
```

jennifer-richards (approved):

PR #11361, opened 2026-08-04
feat: Debian "trixie" + Node 26

jennifer-richards:


NGPixel (approved):

Exhorder1 (commented):
Grat

w4rw2dssxw-netizen (commented):
Thanks for your help

w4rw2dssxw-netizen (commented):
The only way I could do this would probably have to be after work

PR #11387, opened 2026-08-06
fix(bibxml): Process abstract to match current BibXML

kesara:


jennifer-richards:
Seems like a strange convention?

rjsparks:
yes.
We're matching what bib spits out now though so that we minimally affect users when we switch bib to using this output. I plan on revisiting it with the RPC _after_ we've alleviated the pain on bib.

jennifer-richards:
> We're matching what bib spits out now though so that we minimally affect users when we switch bib to using this output. I plan on revisiting it with the RPC _after_ we've alleviated the pain on bib.

Matching is fine, but I'm not following context. Are you saying that making this change (replacing two spaces with a newline) will make it match?

rjsparks:
>making this change ... will make it match?

Hmm, yes, but no - I don't know where the newline is coming from - I thought this would be converting two spaces after a period to one. : @Kesara?

This is the bibxml served by bib.ietf.org at the moment:
```
% cat STD/reference.STD.0009.xml| grep -A2 "<abstract"
      <abstract>
        <t>This memo is the official specification of the File Transfer Protocol (FTP) for the DARPA Internet community. The primary intent is to clarify and correct the documentation of the FTP specification, not to change the protocol. The following new optional commands are included in this edition of the specification: Change to Parent Directory (CDUP), Structure Mount (SMNT), Store Unique (STOU), Remove Directory (RMD), Make Directory (MKD), Print Directory (PWD), and System (SYST). Note that this specification is compatible with the previous edition.</t>
      </abstract>
```

kesara:
@rjsparks I'm not 100% about logic behind new lines in the current BibXML output.

PR #11415, opened 2026-08-07
feat: Person UUID, related OIDC claims, and apis

rjsparks:


ietf/utils/converters.py
```diff
@@ -0,0 +1,24 @@
...
+    Django's built-in "uuid" converter matches lower case only, so an upper-cased
+    identifier would 404 rather than resolve.
+    """
+
+    regex = (
+        "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
+    )
+
+
+register_converter(AnyCaseUUIDConverter, "anycase_uuid")
```
jennifer-richards: Registering this in a module creates a nasty import side effect - Django does not allow re-registering converters, so importing the file in two places may lead to errors / undefined effects.

Closest thing to a canonical pattern would be to define the path converter in a module, then register it in the top level `urls.py`

https://docs.djangoproject.com/en/6.0/topics/http/urls/#registering-custom-path-converters

ietf/ietfauth/views.py
```diff
@@ -232,9 +233,9 @@ def confirm_account(request, auth):
             if not person:
                 name = form.cleaned_data["name"]
                 ascii = form.cleaned_data["ascii"]
-                person = Person.objects.create(user=user,
-                                               name=name,
-                                               ascii=ascii)
+
+                person = Person.objects.create(user=user, name=name, ascii=ascii)
+                assign_primary_uuid(person)
```
jennifer-richards: Should do this in a transaction that rolls back if the `assign_primary_uuid()` call fails

ietf/ietfauth/tests.py
```diff
@@ -759,6 +762,7 @@ def test_change_username(self):
         user.set_password(VALID_PASSWORD)
         user.save()
         p = Person.objects.create(name="Some One", ascii="Some One", user=user)
+        assign_primary_uuid(p)
```
jennifer-richards: Seems like this should use the factory instead of rolling its own uuid

ietf/ietfauth/utils.py
```diff
@@ -341,6 +341,30 @@ def scope_dots(self):
         dots = get_dots(self.user.person)
         return { 'dots': dots }
 
+    info_datatracker_uuid = (
+            "Datatracker person identifier",
+            "Access to the stable identifier the datatracker uses for you when telling "
+            "other systems who you are.",
```
jennifer-richards: Splitting these claims across separate scopes seems unnecessarily finely grained. I don't think we'd ever grant one but not the other, and there aren't going to be enough prior UUIDs to worry about shaving bytes off the response. Just use one scope with both claims.

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+    primary_uuid = serializers.UUIDField(read_only=True)
+    prior_uuids = serializers.ListField(child=serializers.UUIDField(), read_only=True)
+
+
+class PersonPkUnknownEntrySerializer(serializers.Serializer):
+    person_pk = serializers.IntegerField(read_only=True)
+    status = serializers.ChoiceField(choices=["unknown"], read_only=True)
+
+
+class PersonPkBatchResponseSerializer(serializers.Serializer):
```
jennifer-richards, line 108: Per the later comment, this is not a real serializer, it's usable for annotation only because it uses `PolymorphicProxySerializer`. This and other annotation helpers (if any) should be identified by naming convention ("...ProxySerializer") or at least a comment.

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+    results = serializers.ListField(
+        child=PolymorphicProxySerializer(
+            component_name="PersonPkBatchEntry",
+            serializers={
+                "resolved": PersonPkResolvedEntrySerializer,
+                "unknown": PersonPkUnknownEntrySerializer,
+            },
+            resource_type_field_name="status",
```
jennifer-richards: The docs "strongly recommend" using a list for the `serializers` parameter and letting spectacular sort out the mapping via the `status` field on each serializer. I don't have particular insight into why, but probably DRY.

https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.PolymorphicProxySerializer

ietf/nomcom/utils.py
```diff
@@ -421,6 +422,7 @@ def make_nomineeposition_for_newperson(nomcom, candidate_name, candidate_email,
     person = Person.objects.create(name=candidate_name,
                                    ascii=unidecode_name(candidate_name),
                                    )
+    assign_primary_uuid(person)
```
jennifer-richards: Again - transaction

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+        ),
+        read_only=True,
+    )
+
+
+@extend_schema(
+    tags=["person"],
+    parameters=[],
+)
```
jennifer-richards: I don't think `@extend_schema` works on a viewset - at least, it's unusual. These tags need to go on the individual endpoint annotations.

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+            "deleted, because deleting a Person deletes its UUIDs.\n\n"
+            "primary_uuid is the person's current primary. Re-read it; do not assume it "
+            "is unchanged from a previous response, and do not assume a UUID you hold "
+            "remains primary.\n\n"
+            "Responses contain identifiers only. No name, address or database key is "
+            "returned."
+        ),
+        responses=PersonUUIDResolutionSerializer,
+    )
+    def retrieve(self, request, *args, **kwargs):
```
jennifer-richards: We've normally used `@extend_schema_view(retrieve=extend_schema(...))` on the ViewSet class rather than needlessly overriding `retrieve()` just to make a superclass call. Here I'm on the fence because the other endpoint is actually a nontrivial implementation so it's a defensible choice.

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+class PersonPkResolvedEntrySerializer(serializers.Serializer):
+    person_pk = serializers.IntegerField(read_only=True)
+    status = serializers.ChoiceField(choices=["resolved"], read_only=True)
+    primary_uuid = serializers.UUIDField(read_only=True)
+    prior_uuids = serializers.ListField(child=serializers.UUIDField(), read_only=True)
+
+
+class PersonPkUnknownEntrySerializer(serializers.Serializer):
+    person_pk = serializers.IntegerField(read_only=True)
+    status = serializers.ChoiceField(choices=["unknown"], read_only=True)
```
jennifer-richards: Separate serializers here, plus the later instruction to the API consumer to use the `status` field instead of looking at which fields are present is silly. Just use the `status` field, make `primary_uuid` nullable / optional, and use one. That'll get rid of the `PolymorphicProxySerializer` complication below.

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+
+
+class ResolvedEntrySerializer(PersonUUIDResolutionSerializer):
+    status = serializers.ChoiceField(choices=["resolved"], read_only=True)
+
+
+class UnknownEntrySerializer(serializers.Serializer):
+    uuid = serializers.UUIDField(read_only=True)
+    status = serializers.ChoiceField(choices=["unknown"], read_only=True)
```
jennifer-richards: Described in more detail for the PK-based version below, but just use a single serializer with a nullable uuid and do away with the polymorphic serializer

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+class PersonUUIDBatchRequestSerializer(serializers.Serializer):
+    uuids = serializers.ListField(
+        child=serializers.UUIDField(), allow_empty=False, max_length=MAX_BATCH
+    )
+
+
+class PersonUUIDBatchResponseSerializer(serializers.Serializer):
```
jennifer-richards, line 85: As below, if this sticks around, name it `...ProxySerializer` or at least comment that it's not a real serializer.

ietf/person/api_uuid.py
```diff
@@ -0,0 +1,273 @@
...
+                    "primary_uuid": primary,
+                    "prior_uuids": priors,
+                }
+            )
+        # Rendered as plain dicts: PersonUUIDBatchResponseSerializer is an annotation
+        # helper (PolymorphicProxySerializer) and cannot serialize real responses.
+        return Response({"results": results})
+
+
+@extend_schema(tags=["person"])
```
jennifer-richards, line 199: Again, `@extend_schema` does not apply to `ViewSet` classes.

ietf/ietfauth/utils.py
```diff
@@ -341,6 +341,30 @@ def scope_dots(self):
...
+
+    def scope_datatracker_uuid(self):
+        # An empty string is dropped by ScopeClaims._clean_dic, so an inconsistent
+        # Person yields an absent claim rather than a bogus identifier.
+        return {"datatracker_uuid": str(self.user.person.primary_uuid or "")}
+
+    info_datatracker_prior_uuids = (
+            "Previous datatracker person identifiers",
+            "Access to identifiers the datatracker used for you in the past, as well as "
+            "the one it uses now.",
```
jennifer-richards: This comment is incorrect: as written, this is _only_ non-primary UUIDs.

ietf/person/tasks.py
```diff
@@ -57,3 +59,58 @@ def purge_personal_api_key_events_task(keep_days):
...
+    assign_primary_uuid() call at each site that creates a Person, so a site added
+    without one leaves Persons that cannot be named to any external system. This finds
+    them.
+    """
+    broken = (
+        Person.objects.annotate(
+            uuid_count=Count("uuids", distinct=True),
+            primary_count=Count("uuids", filter=Q(uuids__primary=True), distinct=True),
+        )
+        .filter(primary_count=0)
```
jennifer-richards: The db constraint should guarantee we never have multiple primaries, but it would be cheap to search for `.exclude(primary_count=1)` here instead and surface an error if we somehow wound up with grossly inconsistent data. Since we log "every Person has exactly one" below, we might as well actually check that condition.

ietf/person/tests.py
```diff
@@ -394,9 +452,12 @@ def test_merge_persons(self):
         request = HttpRequest()
         request.user = user
         source = PersonFactory()
+        source.uuids.create()  # give them an extra
```
jennifer-richards: use factory

ietf/person/tests.py
```diff
@@ -116,6 +128,52 @@ def test_person_profile_without_email(self):
         r = self.client.get(url)
         self.assertContains(r, person.name, status_code=200)
 
+    def test_person_profile_by_uuid(self):
+        person_a = PersonFactory(name="A Fine Person")
+        uuid_a = person_a.uuids.first()
```
jennifer-richards: While correct we should be using the `primary_uuid` / `prior_uuids` properties instead of directly accessing the fields except where we're testing those properties. Otherwise it's going to encourage bypassing those accessors by example.

This is kind of a nit, but I think it's worth fixing. If we're going to introduce those properties—and we should—we should use them everywhere to keep things maintainable.

ietf/person/tests.py
```diff
@@ -478,6 +547,399 @@ def test_send_merge_request(self):
...
+            confirm_url,
+            {
+                "name": "UUID Test",
+                "ascii": "UUID Test",
+                "password": "secret+password",
+                "password_confirmation": "secret+password",
+            },
+        )
+        created = Person.objects.get(name="UUID Test")
+        self.assertEqual(created.uuids.filter(primary=True).count(), 1)
```
jennifer-richards: `self.assertIsNotNone(created.primary_uuid)` + (maybe) `self.assertEqual(created.prior_uuids, [])` here and in the other branches

ietf/person/tests.py
```diff
@@ -478,6 +547,399 @@ def test_send_merge_request(self):
...
+            "Draft Author", "draftauthor@example.com", "draft-uuid-test"
+        )
+        author = Person.objects.get(name="Draft Author")
+        self.assertEqual(author.uuids.filter(primary=True).count(), 1)
+
+    def test_factory_assigns_a_primary(self):
+        person = PersonFactory()
+        self.assertEqual(person.uuids.filter(primary=True).count(), 1)
+        self.assertIsNotNone(person.primary_uuid)
+        self.assertEqual(person.prior_uuids, [])
```
jennifer-richards, line 610: An equivalent simple test of `person.prior_uuids` would be good - to prove that it'd doing what we think wrt the actual models.

ietf/utils/test_data.py
```diff
@@ -40,6 +41,7 @@ def create_person(group, role_name, name=None, username=None, email_address=None
     user.set_password(password)
     user.save()
     person = Person.objects.create(name=name, ascii=unidecode_name(smart_str(name)), user=user)
+    assign_primary_uuid(person)
```
jennifer-richards, line 44: In these random other touches, I wonder if we should switch to the factory?

ietf/person/utils.py
```diff
@@ -33,28 +34,44 @@ def get_person_uuid_object(uuid_value):
...
     """
     from ietf.person.tasks import push_person_uuids_task  # avoid a circular import
 
     person_pk = person.pk
-    transaction.on_commit(lambda: push_person_uuids_task.delay(person_pk=person_pk))
+
+    def enqueue():
+        try:
+            push_person_uuids_task.apply_async(
+                kwargs={"person_pk": person_pk}, retry=False
```
jennifer-richards: The default retries are pretty minimal - three tries over a total of 0.4 seconds. I think we'd be better off allowing these quick retries because, e.g., I suspect it would let this work through a MQ failover if we ever get that working.

If this is in a critical path, then completely disabling it would be ok, but it doesn't seem like it to me?

jennifer-richards (changes requested):
A few high level things I think we should fix (details are inline):
* fix / simplify the DRF serializers to get rid of polymorphic complications and @extend_schema` misuse
* the oidc claims should just be in a single scope
* guard person creation with transactions
* avoid registering the custom path converter on import
* consider allowing retries on the uuid push task

We should manually test that the schema that comes out of this by running spectacular is correct (if you haven't done that already) and can be used to generate a client.

There are a bunch of other nit-level comments. The ones related to using the Person uuid `@property` accessors in tests could be ignored if it's a lot of manual work, but I'm assuming it's simple to generate code incorporating that comment (if you agree).

ietf/person/api_uuid.py
jennifer-richards: til: it does (for some purposes)

ietf/person/tests.py
```diff
@@ -1016,6 +1016,21 @@ def test_schema(self):
         self.assertEqual(by_pk["operationId"], "person_uuid_by_person_pk")
         self.assertTrue(by_pk["deprecated"])
         self.assertIn("PersonUUIDResolution", schema["components"]["schemas"])
+        # The declared success codes have to be the ones the views actually return -
+        # nothing here creates anything, so nothing may advertise a 201.
```
jennifer-richards: Vague nit (not demanding action): the obsession with _not_ returning a 201 and overexplaining why is a bit odd, but 🤷‍♀️

ietf/person/tasks.py
jennifer-richards, line 110: Nit: will report, e.g., "1 Person(s) repaired" if it's called with `fix=True` and the only broken record has two uuids. 

Ok not to fix since it's an outrageous corner case that's separately logged

ietf/person/models.py
```diff
@@ -242,10 +242,14 @@ def primary_uuid(self):
 
     @property
     def prior_uuids(self):
-        """UUIDs that still resolve to this Person but are no longer primary"""
+        """UUIDs that still resolve to this Person but are no longer primary
+
+        Oldest first, breaking ties on the UUID itself so the order is deterministic -
+        a merge gives every UUID it moves the same timestamp.
```
jennifer-richards: Does it? It only sets the `person` field on each `PersonUUID`. The `time` field is not `auto_now`, it's set on creation. I don't see where the timestamp is changed.

The deterministic ordering is a good change either way, this is just a comment about the rationale.

jennifer-richards (commented):
Thanks for the updates. One question/suggestion to clarify a comment, plus a nit just to point at (in case it bothers you more than it bothers me and you want to adjust the logging, I think it's ok as-is)

ietf/person/tests.py
rjsparks: Getting Claude to produce a clean result rather than a narrative (even in code) about how the result was reached is a challenge.

ietf/person/tasks.py
rjsparks: I'll leave it be.

jennifer-richards (approved):

PR #10139, opened 2025-12-19
fix: use America/Los_Angeles instead of PST8PDT for time zone settings

rjsparks:

PR #11435, opened 2026-08-10
fix: feat/rpc-api timezone

holloway:
## fix

* `feat/rpc-api` timezone PST8PDT to America/Los_Angeles, per #10139 fix as main


rjsparks:
This is a dead branch I think. Needs to be cleaned away instead of maintained

PR #11478, opened 2026-08-12
fix(bibxml): Update orgs

kesara:
* Adds missing organization entries.
 * Fixed IANA organization entry to match existing BibXML output


kesara:
I'm merging this to feature branch since it's a trivial change.

PR #11507, opened 2026-08-13
perf: improve /doc/search performance

rjsparks:
# perf: improve /doc/search performance

## The incident

```
/doc/search/?author=Barry+Leiba&by=author&name=rfc&rfcs=on
```

**118 seconds**, 745M shared buffer hits, ~1GB of temp spill — to return 32 documents. The
same search with the draft checkboxes ticked did not complete at all: Postgres failed it
with `could not resize shared memory segment ... No space left on device` in a parallel
worker, reproducibly.

Profiling turned up two independent failure modes. One pathological query, and a per-row
query storm that affects **every** search, not just this one.

### 1. Multi-valued relations ORed into a single `filter()`

The author filter ORed five lookups spanning `documentauthor → person → {alias,email}` and
`rfcauthor → person → {alias,email}` into one `.filter()` call. Django cannot split those
across separate joins, so the two author paths cross-multiply. The name filter added a
third multi-valued relation, `targets_related`, to the same OR group.

Measured fan-out for the URL above:

| stage | rows |
|---|---|
| `doc_document` where type=rfc | 9,819 |
| × documentauthor | 23,552 |
| × rfcauthor | 81,460 |
| × person/email/alias (draft side) | 467,328 |
| × doc_document_states | 474,948 |
| × person/email/alias (rfcauthor side) | **2,175,516** |
| × targets_related (the `name` filter) | **~126,000,000** |
| surviving the five ILIKEs | 930 |

The planner estimated 7 rows for the node that produced 41.9M, so it committed to nested
loops. A six-million-fold misestimate is not fixable by tuning — the fan-out has to go.

`.distinct()` then ran over the full column list, which `prepare_document_table` widens
further with `select_related()`, forcing a sort over **72 columns** including `abstract`,
`group.description` and `person.biography`.

### 2. A per-row query storm

Rendering the table cost 1,900–3,800 queries for 200 rows: `related_ipr` walked the
obs/replaces graph with a query per node **per row**; `review_assignments_to_list_for_docs`
was called once per document from inside a loop despite being built to answer for many;
`friendly_state` asked for replacements and became-RFC per row; `Person.email()` and
`person_link`'s alias check ran per row because `select_related("ad")` gives every row its
own `Person` instance; and `fill_in_telechat_date` ran a query and threw the result away
because the two lines that stored it were commented out.

## Results

| search | queries before | after | main query before | after |
|---|---|---|---|---|
| the reported URL | 298 | **63** | 118 s | **0.14 s** |
| same author, drafts included | — | 65 | *did not complete* | 0.16 s |
| `by=group&group=dnsop&…` | 1,902 | **88** | | |
| `by=group&group=tls&…` | 1,594 | **164** | | |
| `name=security&…` | 1,708 | **107** | | |
| `activedrafts=on&olddrafts=on&rfcs=on` | 3,765 | **305** | | |

Other views that render the same table, all with byte-identical output:

| view | before | after |
|---|---|---|
| `iesg.agenda_documents` | 616 | **274** |
| `iesg.agenda` | 382 | **196** |
| `doc.docs_for_iesg` | 4,353 | **837** |
| `doc.recent_drafts` | 527 | **126** |
| `doc.drafts_in_last_call` | 304 | **94** |

## Correctness bugs fixed along the way

These were found while profiling and are not performance issues:

**One user's personal bookmark state was served to other users.** `prepare_document_table`
attaches per-user flags (community-list membership, review wishes), and the whole prepared
list was cached under a key with no user in it. Whoever requested a search first had their
state baked in and served to everyone for the next 300 seconds. Demonstrated on `main`:
with user A tracking `draft-ietf-sfc-nsh-ecn-support`, user B's page showed it bookmarked;
with B first, A's own tracked draft showed as untracked. Fixed by caching document ids and
re-preparing per request.

**Two different searches shared a cache entry.** The key came from `QueryDict.items()`,
which returns only the last value of a multi-valued key, so
`?doctypes=charter&doctypes=statchg` and `?doctypes=statchg` collided and could serve each
other's results.

**The heaviest searches were never cached at all.** The cached payload was the prepared
`Document` objects — 1.28MB for 200 rows, over memcached's 1MB item limit, which
`LenientMemcacheCache` exists to swallow silently. The searches most worth caching were
recomputed every time.

**A cached result carried the sort order it was first computed with**, since `sort` is not
part of the key, so clicking a column header could return the previous ordering.

**`fill_in_telechat_date` was a no-op**, so the IESG agenda views it exists to speed up
paid a `latest_event()` query per document anyway.

## Reading the commits

Nine commits, each green on its own, tests alongside the change they guard. The first
three carry the behaviour risk and fix the incident; 4–9 are mechanical batching with
output verified byte-identical.

1. **stop the author and name filters cross-multiplying** — the 118s → 0.14s fix
2. **give each document search its own cache entry** — the `doctypes` collision
3. **cache result ids rather than prepared documents** — payload, personalization, sort order
4. **make `fill_in_telechat_date` actually fill in the telechat date**
5. **look up review assignments once for the whole table** — 342 queries → 3
6. **batch the document relations the table renders**
7. **batch the person lookups behind the AD and shepherd columns**
8. **stop the table refetching rows it already has** — select_related/prefetch corrections
9. **guard the table against per-row queries** — the invariance test

## Verification

- **26 query shapes** compared old vs new at the queryset level against a production-scale
  database: identical result sets, no duplicate rows anywhere. This is what licenses
  dropping `.distinct()`.
- **13 representative searches** rendered through the view before and after and compared
  byte for byte; differences reduce to elapsed-day counters and deliberate tie-ordering.
- **Every view** rendering `search_result_row.html` or `status_columns.html` exercised
  against the live database; output identical, query counts down or unchanged.
- **Full test suite**: 1786 tests. Plus 695 across `doc`, `group`, `community`, `person`,
  `review` and `iesg` on the branch tip.

## Notable Behaviour changes 

**Ordering is now deterministic.** `retrieve_search_results` ordered by `-time`, which is
not a total order — documents sharing a timestamp to the second are common. Both the set
of rows kept by the 200-row truncation and the order equal sort keys fell back to varied
between executions. It now orders by `('-time', 'pk')`. This reorders only what was
previously arbitrary, but it *is* a visible change for tie-heavy sorts (ipr, status, ad).

**`related_that` / `related_that_doc` consult a per-instance cache** when one has been
seeded (commit 6). The cache is read-only and opt-in: nothing else in the tree sets those
attributes and an unseeded document falls through to the original query, so other callers
are untouched. This is the one model-level change.

**`Person.has_alias_for_name()`** is new (commit 7) and `person_link` — used by 62
templates — now calls it. It is the same expression that was inline in the tag, memoised
per instance, with `save()` dropping the memo when the aliases change.

----
Prepared with the assistance of Claude

### DIFF
--- a/ietf/doc/models.py
+++ b/ietf/doc/models.py
@@ -608,12 +608,21 @@ class DocumentInfo(models.Model):
         return related
 
     def related_that(self, relationship):
+        # _cached_related_that is populated in bulk by callers that render many
+        # documents at once (see ietf.doc.utils_search.fill_in_document_relations);
+        # without it each document costs a query per relationship it displays.
+        cached = getattr(self, "_cached_related_that", None)
+        if cached is not None and relationship in cached:
+            return cached[relationship]
         return list(set([x.source for x in self.relations_that(relationship)]))
 
     def all_related_that(self, relationship, related=None):
         return list(set([x.source for x in self.all_relations_that(relationship)]))
 
     def related_that_doc(self, relationship):
+        cached = getattr(self, "_cached_related_that_doc", None)
+        if cached is not None and relationship in cached:
+            return cached[relationship]
         return list(set([x.target for x in self.relations_that_doc(relationship)]))
 
     def all_related_that_doc(self, relationship, related=None):
--- a/ietf/doc/tests.py
+++ b/ietf/doc/tests.py
@@ -5,6 +5,7 @@
 import os
 import datetime
 import io
+import re
 from hashlib import sha384
 
 from django.http import HttpRequest
@@ -24,6 +25,7 @@ from zoneinfo import ZoneInfo
 
 from django.urls import reverse as urlreverse
 from django.conf import settings
+from django.core.cache import cache
 from django.forms import Form
 from django.http import QueryDict
 from django.utils.html import escape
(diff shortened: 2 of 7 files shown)


jennifer-richards (dismissed):
Looks good. Not thrilled with splitting the caching of doc relationships / person aliases and emails between methods on the models that read a cache populated by special purpose utility code, but we can sort out how to rearchitect that later if it becomes a problem.

jennifer-richards (approved):

PR #11549, opened 2026-08-17
feat: show rfc editor queue status on search result rows

rjsparks:


ietf/doc/utils_search.py
```diff
@@ -159,6 +160,34 @@ def reachable_from(start):
...
+    """Attach each document's RFC Editor publication queue status.
+
+    The status column shows it for every document sitting in the RFC Editor queue, and
+    Document.rfc_editor_queue_status() costs a query per such row to find the latest
+    RpcAssignmentDocEvent. Here they take one query between them.
+    """
+    queued_ids = [
+        d.pk
+        for d in docs
+        if d.get_state_slug("draft-rfceditor") in ("in_progress", "blocked")
```
jennifer-richards, line 178: The call to `fill_in_rfc_editor_queue_status()` in `utils_search.py` has prefetched `states` for its documents, but there's also [a call in `iesg/views.py` that does not](https://github.com/ietf-tools/datatracker/blob/b11e7041c057c2b475729c74241f425d2813f777/ietf/iesg/views.py#L408). This will add a query per doc to fetch the states. Would it be worth adding a prefetch or is the number of docs involved small enough that we don't care?

(There's an existing comment in that view that prefetch-related turns out not to be worth it, but maybe it is with this?)

jennifer-richards (commented):
Question inline

ietf/doc/utils_search.py
jennifer-richards: Capturing offline discussion: existing state checks on the `iesg/views.py` code path were already paying that per-query cost, so this was not adding new queries.

jennifer-richards (approved):

PR #11596, opened 2026-08-19
feat: surface RPC requests for AD decisions at datatracker

### LINKED ISSUE #6409: In AD dashboard: Highlight documents that are in AUTH48 and need AD approval
(restoring the original description - the agent consolidation latched onto some OBE text)
### Description

(You can guess why I'd really appreciate this enhancement.)

Yes, this would require some data flow from the RFC editor to datatracker; I don't expect this feature next week.

---
**Agent-assisted triage**

- Suggested size: M
- Suggested priority: medium
- Suggested lane: standard
- Analysis confidence: high
- Updated with agent assistance on 2026-08-07
---

Note this will use the new RPC workflow management system through existing or new datatracker APIs.

<details><summary>Original report</summary>

### Description

(You can guess why I'd really appreciate this enhancement.)

Yes, this would require some data flow from the RFC editor to datatracker; I don't expect this feature next week.

### Code of Conduct

</details>

rjsparks:
Fixes #6409
---
Add a section to /doc/ad/{ad.name} for the actions the RPC is waiting
on that AD for. Each row links to the queue site final review page for
its document.

Add a table to /doc/ad/ listing each AD the RPC is currently waiting on, 
above the state count tables and linking to that AD's document list. 
Only ADs with something pending get a row, and the table is absent 
when the RPC is waiting on nobody.

    This is kept out of the state count tables because it cannot honestly join
    them: those show a 120-day trend, while the open entries the RPC pushes
    are a snapshot with no history to bucket.

Add the people the RPC is waiting on to the RFC Editor block of the
document's main page, so the document itself says who holds the 
decision and since when.

### DIFF
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -457,11 +457,13 @@ def add_state_change_event(doc, by, prev_state, new_state, prev_tags=None, new_t
 
 
 def can_see_rpc_action_holder_comments(user, entries):
-    """Can this user see what the RPC asked an action holder for?
+    """Can this user be shown what the RPC asked an action holder for?
 
-    Who holds an action and since when is public, but the request itself is
-    not: it is shown to the IESG, the Secretariat and the RPC, and to whoever
-    is being asked for the decision.
+    True for the IESG, the Secretariat and the RPC, and for whoever is being
+    asked for the decision. This keeps internal RPC prose off the document
+    page, which is not where most readers are looking for it - it is not
+    protecting the text, which the queue site already publishes on its
+    final-review pages. The AD document list shows it to everyone.
     """
     if has_role(user, ["Area Director", "Secretariat", "RFC Editor"]):
         return True
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -460,10 +460,10 @@ def can_see_rpc_action_holder_comments(user, entries):
     """Can this user be shown what the RPC asked an action holder for?
 
     True for the IESG, the Secretariat and the RPC, and for whoever is being
-    asked for the decision. This keeps internal RPC prose off the document
-    page, which is not where most readers are looking for it - it is not
-    protecting the text, which the queue site already publishes on its
-    final-review pages. The AD document list shows it to everyone.
+    asked for the decision. This gates the document page only - the AD document
+    list shows the request to everyone. It is a display choice rather than a
+    confidentiality boundary: the queue site publishes the same text on its
+    public final-review pages.
     """
     if has_role(user, ["Area Director", "Secretariat", "RFC Editor"]):
         return True
--- a/ietf/doc/utils.py
+++ b/ietf/doc/utils.py
@@ -456,14 +456,16 @@ def add_state_change_event(doc, by, prev_state, new_state, prev_tags=None, new_t
     return e
 
 
-def can_see_rpc_action_holder_comments(user, entries):
-    """Can this user be shown what the RPC asked an action holder for?
+def show_rpc_action_holder_comments(user, entries):
+    """Should this user be shown the RPC's requests to these action holders?
 
-    True for the IESG, the Secretariat and the RPC, and for whoever is being
-    asked for the decision. This gates the document page only - the AD document
-    list shows the request to everyone. It is a display choice rather than a
-    confidentiality boundary: the queue site publishes the same text on its
-    public final-review pages.
+    Controls whether a view renders the comment the RPC attached to an action
+    holder. Only the document main page views ask: they show it to the IESG, the
+    Secretariat and the RPC, and to whoever is being asked for the decision.
+    Views that do not call this show the comments to everyone, as the AD
+    document list does. The text is not confidential either way - the queue site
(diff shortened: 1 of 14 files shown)


ietf/sync/tasks.py
```diff
@@ -382,6 +391,73 @@ def format_rpc_queue_status(obj):
...
+    if holder.get("body"):
+        return None  # a body holds the action, not a person
+    person = Person.objects.filter(pk=person_id).first()
+    if person is None:
+        log.log(
+            f"process_rpc_queue_task: unknown action holder person {person_id}"
+        )
+        return None
+    if person == system_person or person.pk == SYSTEM_PERSON_ID:
+        return None
```
jennifer-richards: Redundant with l412-413

ietf/templates/doc/ad_list.html
```diff
@@ -37,6 +37,29 @@ <h1>IESG Dashboard</h1>
         <a class="btn btn-primary" href="{% url 'ietf.doc.views_search.docs_for_iesg' %}">Documents in IESG Processing</a>
         <a class="btn btn-primary" href="{% url 'ietf.iesg.views.working_groups' %}">IESG view of Working Groups</a>
     </p>
+    {% if rpc_pending %}
+        <h2 class="mt-5" id="rpc-pending">RPC decisions pending</h2>
+        <p class="text-body-secondary">
+            Documents in the RFC Editor queue for which the RPC is currently waiting on a
+            decision from an Area Director. Current state, not a {{ delta }}-day trend.
```
jennifer-richards: The "Current state, not a ..." comment seems unnecessary

ietf/doc/utils.py
```diff
@@ -456,6 +456,22 @@ def add_state_change_event(doc, by, prev_state, new_state, prev_tags=None, new_t
...
 
 
+def can_see_rpc_action_holder_comments(user, entries):
+    """Can this user be shown what the RPC asked an action holder for?
+
+    True for the IESG, the Secretariat and the RPC, and for whoever is being
+    asked for the decision. This gates the document page only - the AD document
+    list shows the request to everyone. It is a display choice rather than a
+    confidentiality boundary: the queue site publishes the same text on its
+    public final-review pages.
```
jennifer-richards: Also available via api/v1 and from a dev database snapshot (don't need to add this comment, just pointing it out)

Framing it as a `can_` method, which we've mostly used as a permission control for actions, is a possible source of confusion. The docstring is probably adequate to clarify though.

ietf/templates/doc/drafts_for_ad.html
```diff
@@ -80,6 +80,43 @@ <h2 class="mt-4">Missing ballot positions for {{ ad.name }}</h2>
...
+            </thead>
+            <tbody>
+                {% for entry in rpc_action_holders %}
+                    <tr>
+                        <td>{{ entry.document.displayname_with_link }}</td>
+                        <td>
+                            {% if entry.final_review_url %}
+                                <a href="{{ entry.final_review_url }}">Final review</a>
+                            {% else %}
+                                <span class="text-body-secondary">(No RFC number yet)</span>
```
jennifer-richards: Switching the label between "Final review" and "(No RFC number yet)" is a peculiar toggle

ietf/sync/tasks.py
```diff
@@ -382,6 +391,73 @@ def format_rpc_queue_status(obj):
     return ", ".join(parts)
 
 
+# The datatracker's "(System)" person. The RPC tool substitutes a placeholder
+# person of its own whenever no real person was named, and sends that person's
+# datatracker id - so an action holder must never be associated with this pk.
+# Which pk the RPC tool uses is configurable at its end, hence the separate
+# check against the "(System)" person the datatracker actually has.
+SYSTEM_PERSON_ID = 1
```
rjsparks, line 399: yuck. I think this is already captured somewhere more global

jennifer-richards (changes requested):
Comment and a few nits

ietf/sync/tasks.py
rjsparks: No it isn't - the constant is scattered around the code. Cleanup for a separate task (see #11607)

ietf/templates/doc/ad_list.html
rjsparks: It's there to answer "Why aren't you graphing this"

rjsparks: I'll remove it anyhow.

jennifer-richards: It seemed evident to me from the description / existence as just a number, but I'm not the audience so fine either way.

ietf/templates/doc/drafts_for_ad.html
rjsparks: Claude doesn't understand yet (and neither the dt nor purple code enforce) that there won't be action holders until the doc is in final review, and it's built a few gate places around that happening. Still, it's a confusing artifact, and I'll remove it.

jennifer-richards (approved):

PR #11597, opened 2026-08-19
feat: Person UUID, related OIDC claims, and apis (#11415)

### LINKED ISSUE #11221: Give Person a set of stable UUIDs (one primary) as the external identifier, durable across merges
### Background

Sometimes a workflow causes a new `Person` object to be created when an existing `Person` should have been used. We have tooling to merge these duplicate `Person` objects (and the objects associated with them). We're only acknowledging here that duplicates happen and get merged — preventing them is out of scope for this issue.

### Problem

Systems outside the datatracker — e.g. **MeetEcho** and **purple** — pass `Person` references around using the datatracker's database **pk**. When we merge duplicate `Person` records, a pk that an external system is holding **disappears**. Without synchronous, coordinated changes across every relying system, there is no simple way to reattach a reference to a `Person` pk that has gone away (because of a merge) to the right surviving pk. Merging people in those external systems is difficult, and the pk was the wrong thing to export in the first place.

### Proposal

Add a **set of UUIDs** to `Person`, with exactly one marked **primary**.

- Relying applications are handed the **primary UUID** as the person identifier — never the pk.
- When two `Person` records are merged, the UUIDs from **both** records are added to the surviving record's set (union). A single primary is retained.
- When a relying application later asks about a UUID it holds and a merge has since happened, the lookup consults the whole set, so the old UUID **still finds the right (surviving) `Person`**.
- On such a lookup, the application can then be told the **new primary UUID** so it can update its own stored reference at its own pace — no synchronous cross-system migration required.

### Relationship to #6087

This **replaces #6087** ("Capture Person/User merges going forward"). #6087 proposed `MergedPersons`/`MergedUsers` tables so APIs could follow `old_pk → new_pk`; a UUID set decouples external identity from the pk entirely and makes old identifiers self-resolving, so relying systems no longer need to chase a merge chain keyed on an internal pk. Adopting this obviates #6087, which should be closed in favor of this issue.

### Sketch of work

1. Add UUID storage to `Person` (a related `PersonUUID` table or equivalent) with a uniqueness constraint and a single "primary" flag; migration to backfill one UUID per existing `Person`.
2. Update the merge flow to union the UUID sets of the merged records and keep exactly one primary.
3. Add a resolver (API + internal helper): UUID → `Person`, returning the current primary UUID so callers can self-heal.
4. Expose the primary UUID through the API surfaces that MeetEcho and purple consume; document the "we may hand you back a new primary UUID" contract.

### Code of Conduct

jennifer-richards:
Resolves #11221 

## Commit history pre-squash

* feat: UUIDs as person identifiers

* feat: person uuid oidc claims and apis

* chore: ruff ruff

* fix: adjust how push is triggered

* fix: keep the mypy ignore on the person model import

Reformatting the import into a parenthesized block moved the ignore comment to the closing paren. mypy reports the simple_history HistoricalPerson and HistoricalEmail attribute errors against the 'from ... import (' line, so the comment has to sit there to suppress them.

* refactor: register the anycase_uuid converter in the root URLconf

Registering it in ietf/utils/converters.py made importing that module a side effect, and Django refuses to register a converter twice, so naming the converter from a second URLconf was a latent error. Define it there, register it once in ietf/urls.py before urlpatterns names it.

* fix: create a Person and its primary UUID atomically

A Person with no primary UUID cannot be named to any external system, so the create and the assign_primary_uuid() that follows it have to succeed or fail together. Covers all three production creation sites, including the draft submission one, and wraps the surrounding aliases and nominee email so a failure part way leaves nothing half-built.

* refactor: give each UUID batch endpoint a single response serializer

The resolved/unknown split needed a PolymorphicProxySerializer, which is an annotation helper rather than a real serializer, so the endpoints hand-built dicts and told consumers not to infer the outcome from which fields were present. Use one entry serializer per endpoint instead, discriminated on status, with the identifier fields nullable and always present, and actually serialize responses through it so the schema cannot drift from what is returned.

Drops the ResolvedStatusEnum/UnknownStatusEnum overrides that existed only to keep the two single-valued status enums apart - there is now one StatusEnum. The entry fields are not read_only because read_only implies required=False, which left a generated client treating even status as optional.

Also annotates retrieve with @extend_schema_view rather than overriding it just to call super().

* refactor: serve the pk-to-UUID conversion from a plain APIView

Routing this lookup through a GenericViewSet forced the handler to be named create, because that is what SimpleRouter maps POST to on a collection route. Nothing is created: the view returned 200 while drf-spectacular inferred 201 from the action name, so the schema advertised a status code the endpoint never sends and a generated client would treat the real response as unexpected.

An APIView.post() returns 200 with no annotation gymnastics. The viewset was buying nothing else - no retrieve, no mixins, and an empty queryset. api_key auth is unaffected, since HasApiKey just reads api_key_endpoint off the view.

The URL is unchanged. Its name loses the router's -list suffix, and the schema test now checks the declared success codes so this cannot drift again.

* feat: carry both UUID claims in one OIDC scope

Splitting the current identifier and the superseded ones across two scopes was finer-grained than any consumer needs - there is no case for granting one and not the other, and the prior list is far too short for response size to matter.

Also corrects the scope description, which claimed the prior list included the identifier in use now. It does not, and datatracker_uuid is where that lives.

* fix: check for exactly one primary UUID, not just one or more

The job logged that every Person has exactly one primary while only looking for Persons with none. The partial unique constraint should make more than one impossible, so finding one means the data is grossly inconsistent and worth reporting - and ensure_primary_uuid() cannot repair that case, since it would be picking a survivor arbitrarily, so it is reported and skipped rather than silently 'fixed'. Same change in the base-test-data check.

* fix: let the UUID push enqueue use the default retry policy

Celery's default is three attempts over well under a second, which is cheap enough on the request path that changed the UUID set and is the difference between riding out a broker blip or failover and dropping the push on the floor. The broker-error catch still keeps an outright outage from failing the datatracker operation.

* chore: add dev API tokens for the person UUID endpoints

Neither endpoint had an APP_API_TOKENS entry in the container config, so every call to them from a dev environment got a 403.

* test: build Person UUIDs with the factories and read them through the accessors

PersonFactory now makes its Person's primary UUID with PersonUUIDFactory instead of calling assign_primary_uuid() itself, so all UUID handling in tests goes through the factories. PersonFactory(primary_uuid=False) covers the no-UUIDs-at-all case, which no production path can reach, replacing the tests that created a Person and then deleted its UUID rows.

Tests now assert through Person.primary_uuid and Person.prior_uuids rather than querying uuids directly, so the accessors are the example to copy. Direct queries remain only where they are the point: the test proving the accessors agree with the rows, and setup that deliberately builds inconsistent state.

Also drops the retry kwarg assertion that went with the old retry=False.

* fix: order prior_uuids deterministically

A merge stamps every UUID it moves with the same time, so ordering the prior list on time alone left the order undefined in exactly the case where there is more than one prior. Break ties on the UUID, which also makes the claim that uuid_sets_for() matches this accessor true - it was already ordering on both.

* docs: correct why prior_uuids breaks ties on the uuid

The previous comment justified the tie-break by claiming a merge gives every UUID it moves the same timestamp. It does not: merge_persons() moves them with a queryset update that names only person and primary, and PersonUUID.time is a per-row default with no auto_now, so each keeps its original timestamp.

The tie-break stands on narrower ground - it makes the order total instead of leaving equal timestamps to the database, and matches the ordering uuid_sets_for() already used - so only the comment changes.

* test: clear over-zealous concerns about API return values

### DIFF
--- a/docker/configs/settings_local.py
+++ b/docker/configs/settings_local.py
@@ -115,6 +115,8 @@ for storagename in ARTIFACT_STORAGE_NAMES:
 APP_API_TOKENS = {
     "ietf.api.red_api" : ["devtoken", "redtoken"],  # Not a real secret
     "ietf.api.views_rpc" : ["devtoken"],  # Not a real secret
+    "ietf.person.api_uuid" : ["devtoken"],  # Not a real secret
+    "ietf.person.api_uuid_by_pk" : ["devtoken"],  # Not a real secret
 }
 
 # Errata system api configuration
--- a/ietf/api/urls.py
+++ b/ietf/api/urls.py
@@ -9,6 +9,7 @@ from django.views.generic import TemplateView
 from ietf import api
 from ietf.doc import views_ballot, api as doc_api
 from ietf.meeting import views as meeting_views
+from ietf.person import api_uuid as person_uuid_api
 from ietf.submit import views as submit_views
 from ietf.utils.urls import url
 
@@ -20,6 +21,14 @@ from .routers import PrefixedSimpleRouter
 # core_router = PrefixedSimpleRouter(name_prefix="ietf.api.core_api")  # core api router
 # core_router.register("email", person_api.EmailViewSet)
 # core_router.register("person", person_api.PersonViewSet)
+
+# Person identity API router
+person_router = PrefixedSimpleRouter(
+    use_regex_path=False, name_prefix="ietf.api.person_api"
+)
+person_router.register(
+    "uuid", person_uuid_api.PersonUUIDViewSet, basename="person-uuid"
+)
 
 # todo more general name for this API?
 red_router = PrefixedSimpleRouter(name_prefix="ietf.api.red_api")  # red api router
@@ -88,6 +97,16 @@ urlpatterns = [
     url(r'^person/email/$', api_views.active_email_list),
     # Related Email listing
     url(r'^person/email/(?P<email>[^/\x00]+)/related/$', api_views.related_email_list),
+    # Transitional pk-to-UUID conversion. Before the router include below so it wins
+    # over the router's uuid/ routes.
+    path(
+        "person/uuid/by-person-pk/",
+        person_uuid_api.PersonUUIDByPersonPkView.as_view(),
+        name="ietf.api.person_api.person-uuid-by-pk",
+    ),
+    # Person UUID resolution API. After the ^person/email/ routes above so those keep
+    # matching first.
+    path("person/", include(person_router.urls)),
     # Draft submission API
     url(r'^submit/?$', submit_views.api_submit_tombstone),
     # Draft upload API
--- a/ietf/ietfauth/tests.py
+++ b/ietf/ietfauth/tests.py
@@ -34,7 +34,13 @@ from ietf.group.models import Group, Role, RoleName
 from ietf.ietfauth.utils import has_role
 from ietf.meeting.factories import MeetingFactory, RegistrationFactory, RegistrationTicketFactory
 from ietf.nomcom.factories import NomComFactory
-from ietf.person.factories import PersonFactory, EmailFactory, UserFactory, PersonalApiKeyFactory
+from ietf.person.factories import (
+    PersonFactory,
+    EmailFactory,
+    UserFactory,
+    PersonalApiKeyFactory,
+    PersonUUIDFactory,
+)
 from ietf.person.models import Person, Email
 from ietf.person.tasks import send_apikey_usage_emails_task
 from ietf.review.factories import ReviewRequestFactory, ReviewAssignmentFactory
@@ -657,7 +663,9 @@ class IetfAuthTests(TestCase):
         )
         user.set_password(VALID_PASSWORD)
         user.save()
-        p = Person.objects.create(name="Some One", ascii="Some One", user=user)
+        p = PersonFactory(
+            user=user, name="Some One", ascii="Some One", default_emails=False
+        )
         Email.objects.create(address=user.username, person=p, origin=user.username)
 
         # log in
@@ -758,7 +766,9 @@ class IetfAuthTests(TestCase):
         )
         user.set_password(VALID_PASSWORD)
         user.save()
-        p = Person.objects.create(name="Some One", ascii="Some One", user=user)
+        p = PersonFactory(
+            user=user, name="Some One", ascii="Some One", default_emails=False
+        )
         Email.objects.create(address=user.username, person=p, origin=user.username)
         Email.objects.create(
             address="othername@example.org", person=p, origin=user.username
@@ -1162,7 +1172,16 @@ class OpenIDConnectTests(TestCase):
             session["nonce"] = rndstr()
             args = {
                 "response_type": "code",
-                "scope": ['openid', 'profile', 'email', 'roles', 'registration', 'dots', 'pronouns' ],
+                "scope": [
+                    "openid",
+                    "profile",
+                    "email",
+                    "roles",
+                    "registration",
+                    "dots",
+                    "pronouns",
+                    "datatracker_uuid",
+                ],
                 "nonce": session["nonce"],
                 "redirect_uri": redirect_uris[0],
                 "state": session["state"]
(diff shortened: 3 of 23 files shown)


rjsparks (approved):

mortalis43:
> Resolves #11221 
> 
> ## Commit history pre-squash
> 
> * feat: UUIDs as person identifiers
> 
> * feat: person uuid oidc claims and apis
> 
> * chore: ruff ruff
> 
> * fix: adjust how push is triggered
> 
> * fix: keep the mypy ignore on the person model import
> 
> Reformatting the import into a parenthesized block moved the ignore comment to the closing paren. mypy reports the simple_history HistoricalPerson and HistoricalEmail attribute errors against the 'from ... import (' line, so the comment has to sit there to suppress them.
> 
> * refactor: register the anycase_uuid converter in the root URLconf
> 
> Registering it in ietf/utils/converters.py made importing that module a side effect, and Django refuses to register a converter twice, so naming the converter from a second URLconf was a latent error. Define it there, register it once in ietf/urls.py before urlpatterns names it.
> 
> * fix: create a Person and its primary UUID atomically
> 
> A Person with no primary UUID cannot be named to any external system, so the create and the assign_primary_uuid() that follows it have to succeed or fail together. Covers all three production creation sites, including the draft submission one, and wraps the surrounding aliases and nominee email so a failure part way leaves nothing half-built.
> 
> * refactor: give each UUID batch endpoint a single response serializer
> 
> The resolved/unknown split needed a PolymorphicProxySerializer, which is an annotation helper rather than a real serializer, so the endpoints hand-built dicts and told consumers not to infer the outcome from which fields were present. Use one entry serializer per endpoint instead, discriminated on status, with the identifier fields nullable and always present, and actually serialize responses through it so the schema cannot drift from what is returned.
> 
> Drops the ResolvedStatusEnum/UnknownStatusEnum overrides that existed only to keep the two single-valued status enums apart - there is now one StatusEnum. The entry fields are not read_only because read_only implies required=False, which left a generated client treating even status as optional.
> 
> Also annotates retrieve with @extend_schema_view rather than overriding it just to call super().
> 
> * refactor: serve the pk-to-UUID conversion from a plain APIView
> 
> Routing this lookup through a GenericViewSet forced the handler to be named create, because that is what SimpleRouter maps POST to on a collection route. Nothing is created: the view returned 200 while drf-spectacular inferred 201 from the action name, so the schema advertised a status code the endpoint never sends and a generated client would treat the real response as unexpected.
> 
> An APIView.post() returns 200 with no annotation gymnastics. The viewset was buying nothing else - no retrieve, no mixins, and an empty queryset. api_key auth is unaffected, since HasApiKey just reads api_key_endpoint off the view.
> 
> The URL is unchanged. Its name loses the router's -list suffix, and the schema test now checks the declared success codes so this cannot drift again.
> 
> * feat: carry both UUID claims in one OIDC scope
> 
> Splitting the current identifier and the superseded ones across two scopes was finer-grained than any consumer needs - there is no case for granting one and not the other, and the prior list is far too short for response size to matter.
> 
> Also corrects the scope description, which claimed the prior list included the identifier in use now. It does not, and datatracker_uuid is where that lives.
> 
> * fix: check for exactly one primary UUID, not just one or more
> 
> The job logged that every Person has exactly one primary while only looking for Persons with none. The partial unique constraint should make more than one impossible, so finding one means the data is grossly inconsistent and worth reporting - and ensure_primary_uuid() cannot repair that case, since it would be picking a survivor arbitrarily, so it is reported and skipped rather than silently 'fixed'. Same change in the base-test-data check.
> 
> * fix: let the UUID push enqueue use the default retry policy
> 
> Celery's default is three attempts over well under a second, which is cheap enough on the request path that changed the UUID set and is the difference between riding out a broker blip or failover and dropping the push on the floor. The broker-error catch still keeps an outright outage from failing the datatracker operation.
> 
> * chore: add dev API tokens for the person UUID endpoints
> 
> Neither endpoint had an APP_API_TOKENS entry in the container config, so every call to them from a dev environment got a 403.
> 
> * test: build Person UUIDs with the factories and read them through the accessors
> 
> PersonFactory now makes its Person's primary UUID with PersonUUIDFactory instead of calling assign_primary_uuid() itself, so all UUID handling in tests goes through the factories. PersonFactory(primary_uuid=False) covers the no-UUIDs-at-all case, which no production path can reach, replacing the tests that created a Person and then deleted its UUID rows.
> 
> Tests now assert through Person.primary_uuid and Person.prior_uuids rather than querying uuids directly, so the accessors are the example to copy. Direct queries remain only where they are the point: the test proving the accessors agree with the rows, and setup that deliberately builds inconsistent state.
> 
> Also drops the retry kwarg assertion that went with the old retry=False.
> 
> * fix: order prior_uuids deterministically
> 
> A merge stamps every UUID it moves with the same time, so ordering the prior list on time alone left the order undefined in exactly the case where there is more than one prior. Break ties on the UUID, which also makes the claim that uuid_sets_for() matches this accessor true - it was already ordering on both.
> 
> * docs: correct why prior_uuids breaks ties on the uuid
> 
> The previous comment justified the tie-break by claiming a merge gives every UUID it moves the same timestamp. It does not: merge_persons() moves them with a queryset update that names only person and primary, and PersonUUID.time is a per-row default with no auto_now, so each keeps its original timestamp.
> 
> The tie-break stands on narrower ground - it makes the order total instead of leaving equal timestamps to the database, and matches the ordering uuid_sets_for() already used - so only the comment changes.
> 
> * test: clear over-zealous concerns about API return values
>

PR #11635, opened 2026-08-24
perf: optimize the person endpoint

rjsparks:


ietf/person/utils.py
```diff
@@ -337,11 +337,14 @@ def get_dots(person):
...
-    aliases = Alias.objects.filter(name__iexact=email_or_name)
+    aliases = Alias.objects.filter(name__iexact=email_or_name).select_related("person")
     persons = set(a.person for a in aliases)
 
     if '@' in email_or_name:
-        emails = Email.objects.filter(address__iexact=email_or_name)
+        # Email.address is a citext column, so an exact match is already
+        # case-insensitive and can use the index on it. Asking for iexact wraps the
+        # column in UPPER() and costs a scan of the table.
+        emails = Email.objects.filter(address=email_or_name).select_related("person")
```
jennifer-richards, line 347: This obligates us to use a case-insensitive collation for this field when we move to Django 5.2. I think we want to do that anyway, but we need to remember to adjust this if we take a different approach with collation.

ietf/person/views.py
```diff
@@ -26,6 +28,11 @@
     merge_persons,
 )
 from ietf.utils.mail import send_mail_text
+from ietf.utils.timezone import RPC_TZINFO
+
+REFERENCE_RELATIONSHIPS = ("refnorm", "refinfo", "refunk", "refold")
+
+PROFILE_CACHE_SECONDS = 900
```
jennifer-richards: perhaps expose via settings?

ietf/person/views.py
```diff
@@ -77,9 +84,144 @@ def ajax_select2_search(request, model_name):
...
+    """Render each person's part of the profile page
+
+    The rendered sections are cached, so a repeat view of a profile - including the
+    revalidation a conditional request makes - costs neither the queries nor the
+    render. They are keyed on the date because they report what a person's roles and
+    drafts are "as of" today.
+    """
+    slowpages = caches["slowpages"]
+    keys = {
+        person.pk: f"person:profile:{person.pk}:{timezone.localdate(today)}"
```
jennifer-richards: I don't think keying on the date makes sense - it's redundant with the cache timeout knob, except in not meaningful corners. In particular, it doesn't keep all the sections of the profile page in sync with each other unless you assume particular timing of role changes relative to midnight.

I guess it's here to keep the sections all agreeing on "has no roles as of <date>" in the display. IMO either let that vary, or (even better) change it to "does not currently have any roles" and be vague about exactly what instant that refers to. Strictly speaking that's more accurate than the existing behavior.

ietf/templates/person/profile_body.html
```diff
@@ -0,0 +1,152 @@
...
+                        </td>
+                        <td>
+                            <a href="mailto:{{ role.email.address }}">{{ role.email.address }}</a>
+                        </td>
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+        {% else %}
+            <p>{{ person.first_name }} has no active roles as of {{ today|date:"Y-m-d" }}.</p>
```
jennifer-richards: See comment above about perhaps changing the "as of <date>" text + interpolation

jennifer-richards (changes requested):
Seems generally fine. Minor comment, a nit, and a more substantive change suggested for the cache keying / statement about "today" - see inline.

jennifer-richards (approved):

PR #11670, opened 2026-08-28
feat: dynamic App API tokens

jennifer-richards:
Adds Django model-backed API tokens alongside the existing settings-backed ones. Uses Django admin for management. Tokens are stored as salted hashes (SHA-384) only, not in the clear.

Caches the token table so most requests only hit the cache to avoid adding any db overhead for token checks. Changes made in the admin immediately refresh the cache to pick up changes. The cost of refreshing the cache is only two queries, so I haven't bothered with keeping the cache continuously warmed via a task - once a day we'll have one API call that hits a cold cache, and I think we can live with that.

The admin generates tokens by default and displays them once to be copied and installed / distributed. A token can be manually specified or replaced if needed, with a minimum length to ensure we don't make bad choices.

Any tokens installed in the settings.py are respected, so we can add this and migrate tokens from settings to the database post-deployment.

We still need to scrub these from our dev database snapshots before beginning to add tokens here. It's somewhat less risky because it no longer stores cleartext tokens, but let's be responsible with our secrets.

One thing to note: this stores the token table in the default cache. We might consider adding a separate "secrets" or "tokens" cache just to segregate these credentials from the general cache contents. I think it's ok to keep it simple to start, but it wouldn't be difficult to add a separate cache.


ietf/api/models.py
```diff
@@ -0,0 +1,67 @@
...
+        if not AppApiToken.objects.filter(token=new_token).exists():
+            return new_token
+        tries = tries - 1
+    # never expected to reach this
+    raise RuntimeError("Unable to generate a unique API token")
+
+
+class AppApiToken(models.Model):
+    endpoints = models.ManyToManyField("api.KnownApiEndpoint", related_name="tokens")
+    token = models.CharField(
```
kesara, line 15: Do we want to keep hashed token here? Similar to `PersonalApiKey`[^1]?

[^1]: https://github.com/ietf-tools/datatracker/blob/main/ietf/person/models.py#L523-L533

jennifer-richards: Hashing it is a good idea in principle. Need to think about the tradeoffs here (the concern about operators seeing the value is not as significant as it is with user-owned secrets)

jennifer-richards: I think you're right, hashing is the right thing to do. 

I updated to store the hash instead. That requires a more sophisticated admin interface to support, so I added that, too.

ietf/api/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @staticmethod
+    def hash(token: str):
+        """Hash a token for storage / comparison
+
+        Salts the value as a precaution. Tokens will generally be long, high-entropy
+        byte strings that are not vulnerable to rainbow table attacks, but this will
+        provide some insurance if someone ill-advisedly adds a simple token.
+        """
+        salt = getattr(settings, "APP_API_TOKEN_SALT_BYTES", b"5a1+Y&+45t`/")
```
kesara, line 75: Might not be a good idea to have a default value for the salt bytes. Ideally this should be in settings (populated from the vault).
If the salt is missing, DT should error out?

kesara:
+1 for scrubbing this data from dev snapshots. We are already doing that for other hashed data.[^1] I think we should start doing from zero-hour, so that change has to happen before we create any new App API tokens.

[^1]: https://github.com/ietf-tools/infra-k8s/blob/main/.github/workflows/datatracker-dev-db.yml#L60-L100

ietf/api/ietf_utils.py
```diff
@@ -1,22 +1,43 @@
...
+from .models import AppApiToken
+
+
+def cached_hashed_token_store(force_update=False):
+    cache = caches["default"]
+    cache_key = "ietf.api.ietf_utils.cached_hashed_token_store"
+    cached_value = None if force_update else cache.get(cache_key)
+    if cached_value is None:
+        cached_value = AppApiToken.objects.as_hashed_token_dict()
+        cache.set(cache_key, cached_value, 86400)
```
kesara, line 21: With this approach, cache provider becomes concern. If cache service was exploited, attacker can set any string for the hashed token.
I'm pointing this because caching services usually have lax security compared to other services.

ietf/api/admin.py
```diff
@@ -0,0 +1,112 @@
...
+        except IntegrityError:
+            self.message_user(
+                request,
+                "This token value collided with one saved by another request "
+                "just now and was not saved. Edit the object again and try a "
+                "different value.",
+                level=messages.ERROR,
+            )
+            raise
+        cached_hashed_token_store(force_update=True)  # update the cache
```
kesara, line 79: Force update to cache here and in `KnownApiEndpointAdmin` are an important steps, specially if we are removing or disabling tokens.

Should there be a verification step to see if cache is updated?

I'm worried about any cache update failures or silently failing to update the cache and we a stale cache until the the cache expiry.

ietf/api/ietf_utils.py
kesara, line 21: Can this silently fail?
Specially in the case of force updates from admin actions, it's not verifying the cached values.

ietf/api/models.py
jennifer-richards: We can set it in production, but I'd rather provide a default here than add another setting to the dev/test/etc settings. It's barely worth salting this to begin with - the tokens themselves are long random values.

ietf/api/admin.py
jennifer-richards: An update should result in an error if the write to memcached fails, it won't be silent. The exception is that we're using the default cache, which ignores errors due to oversized (> 1 megabyte) payloads. That is an arguement for separating the tokens from the default cache so it can use a non-lenient cache backend, but it'll be a while before we have anything approaching a megabyte of tokens to store.

ietf/api/ietf_utils.py
jennifer-richards: Other than due to an oversize payload (see other comment), I am pretty sure it will raise an exception if it's not able to set the value.

ietf/api/models.py
jennifer-richards: Ehhh ok fine you're right. I've done this and also added better salting (and also, I guess it's pepper, not salt, since it's not per-token)

ietf/api/ietf_utils.py
jennifer-richards: That's true. I've added signing using Django's signing utility, so it's as protected as our sessions are.

kesara: Cool, Might be an good idea to include a test for this.

jennifer-richards: Adding a test would get into testing Django / memcache, so I think we need to trust that. I've adjusted the admin so that the cache refresh is in a transaction with the model saves so there is a loud message if something goes wrong. That's probably about as well as we can do. Anything preventing the cache from updating for more than a glitch is going to render the entire datatracker pretty well dead.

rjsparks (approved):

PR #11678, opened 2026-08-31
feat: add idnits3 to draft submission checks

rjsparks:


ietf/submit/models.py
```diff
@@ -145,6 +145,16 @@ def has_warnings(self):
         return self.warnings != '[]'
     def has_errors(self):
         return self.errors != '[]'
+    @property
```
jennifer-richards, line 148: nit: style shifting to using `@property` (vs existing `has_warnings()` and `has_errors()` that are methods)

(though tbf I have no idea what those methods are even doing, they don't make sense at a skim)

jennifer-richards (approved):

phisant46:
[]()@phisant.com

SEZAR03 (commented):
What u mean

PR #11719, opened 2026-09-04
feat: API Key and person UUID based meetecho APIs

rjsparks:
Adds a parallel set of DRF endpoints for the session data Meetecho pushes, keyed by
Person UUID and authenticated by service token. The six existing personal-API-key
endpoints keep working, so Meetecho migrates on its own schedule with no coordinated
cutover.

## New endpoints

| Endpoint | Body |
| --- | --- |
| `POST /api/meeting/session/{id}/video-url/` | `{"url": ...}` |
| `POST /api/meeting/session/{id}/recording-name/` | `{"name": ...}` |
| `POST /api/meeting/session/{id}/bluesheet/` | `{"bluesheet": [{"name", "affiliation"}]}` |
| `POST /api/meeting/session/{id}/attendees/` | `{"attendees": [{"person_uuid", "join_time"}]}` |
| `POST /api/meeting/session/{id}/chatlog/` | `{"chatlog": [...]}` |
| `POST /api/meeting/session/{id}/polls/` | `{"polls": [...]}` |
| `GET /api/meeting/registration/attended/by-uuid/{uuid}/` | — |

JSON in, JSON out, `X-Api-Key` via `api_key_endpoint`. Errors use the project's
standardized error envelope.

## Decisions

- **Identifiers.** Only the attendees endpoint carried a person identifier
  (`user_id`, a `User` pk — the OIDC `sub`); it becomes `person_uuid`. Any UUID the
  datatracker issued resolves, including one superseded by a merge. Meetecho already
  gets these from the `datatracker_uuid` OIDC claim. Sessions stay addressed by pk.
- **Free text stays free text.** Bluesheet names and chatlog authors are what the
  uploader observed, not person references — Meetecho can't always resolve a room
  attendee or chat participant to a Person.
- **Authorship.** A token identifies an application, so `request.user` is
  `AnonymousUser` and document events are authored by the `(System)` Person. This also
  drops the Recording Manager role check: authorization is possession of the token.
- **All-or-nothing attendees.** Matching the existing api, one unresolvable UUID rejects the 
  request; the rows and an interim meeting's bluesheet are written in one transaction. A version 
  that allows partial success is left as a future exercise.

## Deploying

Needs seven `APP_API_TOKENS` keys: `ietf.api.meeting.session.{video_url,
recording_name,bluesheet,attendees,chatlog,polls}` and
`ietf.api.meeting.registration.attended_by_uuid`. Separate keys so any one capability
can be withdrawn independently; the same token *value* can serve all seven.

## Changes to existing behavior

Commits 1–3 touch code the existing endpoints share. Five of the six are
behavior-identical. The exceptions, both on a session with no official timeslot:

- `api_set_session_video_url` and `api_upload_bluesheet` already answered 400 for that
  input, because `require_api_key` catches `AttributeError`. Only the message changes.
- `upload_session_bluesheets` and `finalize` have no such catch, so they change from a
  500 to a form error and a user message.

## Notes

Reviewed adversarially before merge; the findings are folded into the commits they
belong to. Known and deliberately not fixed here: the legacy recording-name endpoint
still does no length validation against its 64-character column.

### DIFF
--- a/ietf/meeting/api.py
+++ b/ietf/meeting/api.py
@@ -3,8 +3,9 @@ from django.core.files.base import ContentFile
 from django.db import transaction
 from django.db.models import IntegerField
 from django.db.models.functions import Cast
+from django.http import Http404
 from django.template.loader import render_to_string
-from drf_spectacular.utils import extend_schema
+from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework import generics, serializers, status
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
@@ -28,6 +29,37 @@ from ietf.meeting.utils import (
     save_session_video_url,
 )
 from ietf.person.models import Person, PersonUUID
+from ietf.person.utils import get_person_uuid_object
+
+
+def attended_registrations(person: Person):
+    meetings_with_data = (
+        Meeting.objects.filter(type="ietf")
+        .annotate(number_as_int=Cast("number", output_field=IntegerField()))
+        .exclude(number_as_int__lt=110)
+        .values_list("pk")
+    )
+    has_attended_record = (
+        person.attended_set.filter(
+            session__meeting_id__in=meetings_with_data,
+            session__meeting__type="ietf",
+        )
+        .values_list("session__meeting__id", flat=True)
+        .distinct()
+    )
+    return sorted(
+        [
+            reg
+            for reg in (
+                person.registration_set.onsite_or_remote()
+                .with_plenary_ticket_details()
+                .filter(meeting_id__in=meetings_with_data)
+                .select_related("meeting")
+            )
+            if (reg.attended or reg.checkedin or reg.meeting_id in has_attended_record)
+        ],
+        key=lambda reg: reg.meeting.date,
+    )
 
 
 class MeetingsAttendedByEmail(generics.RetrieveAPIView):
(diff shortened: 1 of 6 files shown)


ietf/meeting/utils.py
```diff
@@ -202,7 +202,7 @@ def bluesheet_data(session):
     ]
 
 
-def save_bluesheet(request, session, file, encoding='utf-8'):
```
onesezar, line 205: Hey

ietf/meeting/utils.py
```diff
@@ -854,6 +855,37 @@ def new_doc_for_session(type_id, session):
     session.presentations.create(document=doc,rev='00')
     return doc
 
+
+def save_session_json_doc(session, type_id, data, by):
+    """Store a chatlog or polls upload as a new revision of the session's document
+
+    Returns an error message, or None on success.
```
jennifer-richards, line 864: Yuck - I don't like the "return an error or None" pattern for python (would prefer exceptions), but I see that it is similar to other things in this space. I might be inclined to start breaking with that pattern and move toward using the exceptions though. It's more idiomatic unless I'm missing something

ietf/meeting/utils.py
```diff
@@ -1200,6 +1232,39 @@ def store_blobs_for_one_meeting(meeting: Meeting):
         store_blobs_for_one_material_doc(doc)
 
 
+def save_session_video_url(session, url, by):
+    """Point the session's video recording at url
+
+    Updates the newest existing video recording, or creates one. Returns an error
+    message, or None on success.
```
jennifer-richards, line 1241: Same comment re: exception vs returning an error msg

ietf/meeting/api.py
```diff
@@ -59,3 +78,163 @@ def _attended_registrations(person: Person):
             ],
             key=lambda reg: reg.meeting.date,
         )
+
+
+@extend_schema(tags=["meeting"])
+class SessionDataView(APIView):
```
jennifer-richards: This is looking like a reinvention of a DRF `ViewSet` for the `Session` model with custom actions. It's worth looking to see if that would give a similar or cleaner implementation instead of rolling a new pattern here. 

I'm not sure how it would shake out. There may be some friction involved (minor: POST might try to return 201 vs 200 and need to be overridden; perhaps more significant: it might be more difficult to give each endpoint its own API key endpoint to use)

jennifer-richards (commented):
Not completely done reading, but wanted to get some comments up

ietf/meeting/serializers.py
```diff
@@ -57,6 +58,34 @@ class SessionBluesheetSerializer(serializers.Serializer):
     bluesheet = BluesheetEntrySerializer(many=True, allow_empty=True)
 
 
+class AwareDateTimeField(serializers.DateTimeField):
+    """DateTimeField that requires an explicit UTC offset
+
+    DateTimeField makes a naive value aware in the process timezone, which silently
+    shifts a timestamp the caller meant as UTC.
```
jennifer-richards: In other existing DRF code (the RPC APIs), we instead used `DateTimeField(default_timezone=datetime.UTC)`. That treats ambiguous timestamps as UTC instead of system time. It's probably better to force API users to be specific rather than guess UTC, but we should avoid mixing behaviors in the API.

I expect we can safely switch the existing uses of `serializers.DateTimeField` to this one without breaking anything.

If we go that route, this shouldn't be a meeting-specific field class. We probably should start an `ietf.utils.rest_framework.fields` module (or `...drf.fields` if we want to be terse rather than consistent with the library) and put this there.

jennifer-richards (changes requested):
One further comment in addition to my previous set. See inline - we should avoid treating `DateTimeField`s on serializers differently. I'm happy with either the minimal effort approach (modifying this PR to default to UTC instead of rejecting naive timestamps) or the refactor-existing-code approach, but the latter is probably slightly better practice.

jennifer-richards (dismissed):

jennifer-richards (approved):

Issue #4670, opened 2022-10-28
testReorderSlides does nothing


jennifer-richards:
### Describe the issue

The test case `ietf.meeting.tests_js.SlideReorderTests.testReorderSlides()` has a body preceded by `return`, so is entirely short-circuited. It has been this way since 2016. Should cut the test out, remove the `return` and make the test do something, or document why we're keeping it around in its no-op form.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

---
Aggregated with agent assistance on 2026-08-18

rjsparks:
It should be replaced with a PlayWright test. It's fine to just remove the unexercised-since-2016 test rather than try to bring it back into a working state.

PR #11735, opened 2026-09-09
fix: use data-name when moving slides between sessions (#7144)

### LINKED ISSUE #7144: Drag and drop of slides between sessions does not work
### Describe the issue

The materials page for a group with multiple sessions at a full IETF meeting allows drag/drop reordering of slide decks and allows slides to be dragged from one session to another. This seems to be broken. Upon dropping, a calls are made to `ajax_add_slides_to_session()` and `ajax_remove_slides_from_session()` appropriately, but both calls return `{"success": false, "error": "Supplied name is not valid"}` with a 200 status code.

### Code of Conduct

---
Aggregated with agent assistance on 2026-08-10

rjsparks:
Fixes #7144.

Dragging a slide deck from one session's slides table to another on a group's
materials page (`/meeting/<num>/session/<acronym>`) always failed with "Supplied
name is not valid". There were two independent defects; fixing only the first
leaves the case chairs actually care about still broken, so both are handled here.

### 1. The dragged row's document name was never sent

`session_details.js` read `event.item.getAttribute('name')` in `onAdd()` and
`onRemove()`, but the row template sets `data-name`, not `name`. `getAttribute()`
returned null, jQuery serialised it to an empty string, and both
`ajax_add_slides_to_session` and `ajax_remove_slides_from_session` rejected the
post at their `Document.objects.filter(name=name).first()` guard.

The two spellings diverged in ed30521e14 ("Many more HTML fixes.", Feb 2022),
which renamed the `<tr>`'s `name=` to `data-name=` without updating the reader.
`onEnd()` sends no name, which is why reordering *within* a session kept working
and this went unnoticed.

### 2. A session with no slides had no drop target

The `<tbody>` carrying `data-add-to-session` / `data-remove-from-session` /
`data-reorder-in-session` was inside `{% if session.filtered_slides %}`, so a
session with no slides rendered a `<table>` with no `<tbody>` at all.
`document.querySelectorAll('.slides tbody')` never found it and no Sortable was
created for it. Moving a deck into a session that doesn't already have one — the
main thing chairs want to do — was impossible even with defect 1 fixed.

This is beyond #7144 as filed, but shipping 1 without it would have left the
feature effectively still broken, so it is included rather than raised separately.

The `{% for %}` loop handles the empty case on its own, so the conditional is
simply removed. An empty `<tbody>` has no height, though, and cannot be dropped
onto, so the JS adds a placeholder row to any empty slides table and keeps it in
sync as decks move in and out. Sortable is now configured with
`draggable: '.draggable'` (the class the template already puts on slide rows) so
the placeholder is not itself draggable and `oldDraggableIndex`/`newDraggableIndex`
count only real slide rows — the indices the endpoints expect.

The empty table's placeholder reads "No slides. Drag a slide deck here to add it
to this session.", and the hint under each table now says decks can be dragged
between sessions as well as reordered — previously it said only "Drag-and-drop to
reorder slides", which undersold what the feature is for.

### Verified

- Confirmed the attribute mismatch and its origin in ed30521e14, and that both
  endpoints reject on it.
- Confirmed `list.js` only initialises `table.tablesorter`; the slides table is
  not one, so the always-rendered `<tbody>` does not affect it. Nothing else on
  the page touches `.slides`, and an empty `<tbody>` passes the suite's HTML
  validation.
- Walked the SortableJS 1.15.7 source to confirm `add` fires before `remove`
  before `end`, that `index(el, selector)` and `lastChild(el, selector)` respect
  the `draggable` selector, and that a hover over the placeholder lands in the
  "insert to end of list" branch and appends the row.

### Tests

`SessionDetailsTests.test_session_details_slides_drag_and_drop_markup` (new,
`ietf/meeting/tests_views.py`) asserts every session's slides table renders a
`<tbody>` with the three endpoint URLs — including a session with no slides — and
that slide rows carry `data-name` and `draggable`. It fails against the
pre-change template ("Session N has no slides tbody to drop onto"), so it is a
real guard for defect 2 and for the attribute contract behind defect 1.

**The JS behaviour itself ships without automated regression cover.** The bug is
entirely client-side and only a browser can exercise it. The existing server-side
`ReorderSlidesTests` posts valid document names directly and passes both before
and after this change — it could never have caught this. The Selenium harness is
being retired, so no browser test was added there.

I did verify the client-side behaviour manually, outside the repo: a throwaway
static page mocking two sessions' slides tables, with `$.post` stubbed to record
calls, driven by a real browser. Against `main` a cross-session drag posts

    POST .../222/add_to_session       {"order":1,"name":null}
    POST .../111/remove_from_session  {"oldIndex":2,"name":null}

which is exactly the "Supplied name is not valid" path in #7144. With this change
the same drag posts the correct `name`, and across 15 checks the `order` /
`oldIndex` / `newIndex` sent always matched the row's final position among the
draggable rows — including drops onto an empty session's placeholder, drops onto
a session that already has decks, and reordering within a session. None of that
is committed; it was verification, not cover.

**#4670 should stay open.** `tests_js.py::testReorderSlides` is still stranded
below a bare `return`, untouched by this PR. Worth recording there: its selector
is `#slides`, but the table's id is `slides_<session pk>`, so even with the
`return` removed the test could not have found the rows. Given Selenium is going
away, that test wants deleting along with the harness rather than repairing.

### Noticed, not fixed

- `onAdd()` rewrites the moved row's links with a bare
  `href.replace(old_session, new_session)` applied to *every* `<a>` in the row's
  second cell — including the document's `get_href` and its `document_main` link,
  neither of which contains a session id. If a slide document's name or the
  meeting number happens to contain the source session's pk as a substring, those
  URLs get corrupted. Replacing the `/session/<pk>/` path segment, on the two
  management links only, would be safer.
- `onAdd()` and `onRemove()` fire two independent `$.post()`s with no success or
  error handling, so a failed call leaves the page silently disagreeing with the
  database until reload.

Both are pre-existing and neither blocks the drag from working.


rjsparks:
Sample of the added placeholder for dropping slides:
image

jennifer-richards (approved):

PR #11738, opened 2026-09-10
refactor: require an explicit UTC offset on API datetime fields

rjsparks:
Adds `AwareDateTimeField` in a new `ietf.utils.rest_framework.fields` module and uses
it for every DRF datetime field in the API.

Addresses the review comment on #11719 about not having two conventions for datetime
fields, taking the refactor-existing-code option rather than the minimal one.

## Why

`DateTimeField(default_timezone=...)` assigns a zone to a value that carries none, so
an offsetless timestamp is accepted and stored as a different instant than the caller
meant. The field requires the offset instead of guessing at it. It overrides
`enforce_timezone`, which is the hook that sees the parsed value while it is still
naive -- checking after `to_internal_value` is too late, the zone has already been
applied.

## Fields changed

- `SessionAttendeeSerializer.join_time` (`ietf/meeting/serializers.py`)
- `RfcPubSerializer.published` (`ietf/api/serializers_rpc.py`)
- `EditableRfcSerializer.published` (`ietf/api/serializers_rpc.py`)
- `RfcPubFilesView`'s `mtime` (`ietf/api/serializers_rpc.py`)

The last three previously used `default_timezone=UTC`.

## Breaking change, needs confirmation before merge

This rejects offsetless timestamps that `POST /api/purple/rfc/publish/` and
`POST /api/purple/rfc/publish/files/` used to accept. Every existing test sends an
offset, so a green suite is **not** evidence that Purple does. Someone should confirm
what Purple actually sends before this merges.

## Follows

#11719, where this was raised in review. One commit, rebased onto main.

### DIFF
--- a/ietf/api/serializers_rpc.py
+++ b/ietf/api/serializers_rpc.py
@@ -31,6 +31,7 @@ from ietf.group.models import Group, Role
 from ietf.group.serializers import AreaSerializer
 from ietf.name.models import StreamName, StdLevelName
 from ietf.person.models import Person
+from ietf.utils.rest_framework.fields import AwareDateTimeField
 from ietf.utils import log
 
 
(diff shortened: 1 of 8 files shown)


rjsparks:
Rudi has already confirmed Purple does not send naive timestamps.

jennifer-richards (approved):

muhammadtahir8900msb-netizen:
``

PR #11760, opened 2026-09-14
feat: account migration apis

rjsparks:


ietf/ietfauth/api_migration.py
```diff
@@ -0,0 +1,438 @@
...
+    # An Email with no Person names nobody, so it cannot disagree with the username.
+    addressee = email.person if email else None
+
+    if user is None:
+        user = addressee.user if addressee else None
+    person = Person.objects.filter(user=user).first() if user else None
+    if addressee is not None and person is not None and addressee != person:
+        return None
+    if user is None:
+        User().set_password(password)
```
jennifer-richards: If we're going to resist this timing attack, it'd be better to call `check_password("blahblah")` to avoid questions about whether `set_password()` and `check_password()` are similar. They _are_, because `check_password()` does the same thing, but better to let Django auth be responsible for keeping the fake-timing path consistent.

While at it, the path above that fails fast when multiple usernames match is also leaking information about accounts existing.

OTOH, if someone wants to enumerate accounts by a timing analysis then maybe it's good to give them a signal to work with so they stay busy and don't realize the account lists aren't actually secret. (Seriously though, we should do it right if at all, just on principle.)

ietf/ietfauth/api_migration.py
```diff
@@ -0,0 +1,438 @@
...
+            "password material, and no database keys apart from legacy_sub, which is the "
+            "subject the datatracker used to issue for this account and is needed only "
+            "to keep existing relying parties working."
+        ),
+        request=VerifyRequestSerializer,
+        responses={
+            200: VerifyResponseSerializer,
+            400: None,
+            401: None,
+        },
```
jennifer-richards: I don't think this is the right way to declare the custom errors. Since these are explicitly raised in the view code, I believe they should be annotated via the `@extend_validation_errors` decorator (https://drf-standardized-errors.readthedocs.io/en/latest/openapi.html#customize-error-codes-on-an-operation-basis).

(That's not something we've used before, but we should sort out how to generate the schema that we want)

ietf/ietfauth/api_migration.py
```diff
@@ -0,0 +1,438 @@
...
+
+
+class AddressHasNoOwner(exceptions.APIException):
+    """The address exists but no Person owns it
+
+    Refused rather than adopted. These rows come from places that record an address
+    without establishing who is behind it - draft submissions, roles - and history points
+    through them. Handing one to whoever proved a password would attribute that history on
+    no evidence, and the endpoint cannot tell the true owner from a namesake. Support
+    establishes ownership; enrollment does not.
```
jennifer-richards: Comment is just not obviously AI enough. Can we amp that up a little bit?

ietf/ietfauth/api_migration.py
```diff
@@ -0,0 +1,438 @@
...
+            "established who is behind it - support establishes that, not enrollment.\n\n"
+            "The response describes the address as it now stands. It does not make the "
+            "address primary - which address is primary is the person's own profile "
+            "choice and is not changed here."
+        ),
+        request=ClaimEmailRequestSerializer,
+        responses={
+            200: MigrationEmailSerializer,
+            404: None,
+            409: None,
```
jennifer-richards: Again - need to be sure spectacular will generate the expected schema for the error cases

jennifer-richards (commented):
A few comments. Need to check that we're happy with the OpenAPI schema that we get out of this - I'm not sure that the existing annotations get the errors right. The password timing attack resistance is also a little wonky but that's not an essential feature, so this is really just a nit.

ietf/ietfauth/api_migration.py
rjsparks: For anyone reading along, this is Claude assisted work and I stopped it digging too deep on the timing resistance attack. As Jennifer notes, it's not essential for the interfaces this will be exposed to.

jennifer-richards (approved):
**Approved!** 🎉

I want to start by saying I really appreciate the work that went into this PR — and I mean that genuinely. Before I get into the substance, let me be direct about something: my overall assessment here is positive.

That said, I think it's worth taking a step back and being nuanced about the response codes. You've clearly put thought into them, and I'm noticeably happier with where they've landed compared to the previous iteration. However, I'd be doing you a disservice if I didn't acknowledge that I'm not completely convinced this represents the ideal approach. To be clear, that's not a criticism — it's more of an observation, and reasonable people could disagree.

A few things I want to be transparent about:

* **What I'm confident in:** The direction is sound, and the changes are a meaningful improvement.
* **What I'm less certain about:** Whether this particular set of codes is the optimal long-term choice. I could be wrong here!
* **What I'm not expecting:** That we resolve this open question in the scope of this PR.

Ultimately, I don't think this needs to block anything, and I'm comfortable moving forward. I'd gently suggest we keep it in mind as a potential follow-up, but that's entirely your call.

I hope this feedback is helpful! Please don't hesitate to reach out if you'd like to discuss any of this further — I'm always happy to dig deeper. 🙂

NGPixel (approved):

PR #11770, opened 2026-09-16
fix: apply materials only to a group's scheduled sessions

### LINKED ISSUE #10579: Upload / approve session slides incorrectly numbers sessions
### Describe the issue

Sessions are numbered in chronological order on the session page for a group (e.g., https://datatracker.ietf.org/meeting/125/session/moq). When uploading or approving slides, the UI shows the correct session but sometimes has the number wrong.

In the identified cases, it is because of a canceled and rescheduled session. On the moq session listed above, a session on Monday was canceled and replaced by the Tuesday session. The session page lists the Tuesday session as "Session 1" but when uploading / approving it is "Session 2". This is very likely because that view is including the canceled session in its counting.

Need to ensure that the numbering is consistent.

### Code of Conduct

---
Aggregated with agent assistance on 2026-08-10

rjsparks:
When a group has several sessions at a meeting and some are not in state
`sched`, uploading materials to one session could change another.

Every material upload view built its idea of "all of the group's sessions"
from the full session list. Cancelled sessions and the tombstones left by
rescheduling keep their timeslot assignment, so they are in that list. With
"apply to all" checked, which is the default, a deck uploaded to one session
was attached to those too, and Meetecho was told about slides for sessions
that will not happen. Minutes were worse: `save_session_minutes_revision()`
deleted and replaced the minutes on every session of the group in any state.

The same list drove the "Session N" heading on the upload pages, while the
materials page numbered each panel separately with `forloop.counter`. So a
group with sessions A (scheduled), B (cancelled) and C (scheduled) called C
"Session 2" on the materials page and "Session 3" on its upload page. A
reschedule tombstone also rendered under "Unscheduled Sessions" with a time,
no status, and a full set of upload buttons, indistinguishable from a live
session.

The fix is one definition of the covered set. `apply_to_all_sessions()`
returns the group's scheduled sessions in schedule order together with the
acted-on session's position in that list. A session that is not itself
scheduled is covered alone: it gets no number, no "apply to all" checkbox,
and changes to it stay with it. The materials page numbers from the same
list and shows a status name instead of a number for unscheduled sessions.

Out of scope, noted for later: revision uploads default "apply to all" to
checked, so revising a deck on one session adds it to every scheduled
session. `docname_token()` is unchanged; its pk ordering is deliberate and
now says so in a comment.

The commits are ordered for review: two pure refactors that introduce the
helpers, three behavior changes each with tests that fail without them,
and the comment.

Fixes #10579

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ietf/meeting/utils.py
```diff
@@ -690,6 +690,37 @@ def get_meeting_sessions(num, acronym):
     return sessions
 
 
+def get_sessions(num, acronym):
```
jennifer-richards: I think this can be safely squashed into `get_meeting_sessions()`. I don't see any other use of that function and it improves the naming a bit

ietf/meeting/models.py
```diff
@@ -1307,6 +1307,9 @@ def _alpha_str(n: int):
         return "".join(chars[::-1])
 
     def docname_token(self):
+        # Position in pk order, so the token never changes for the life of the session. It is an
+        # identifier, not a sequence number: sessions get reordered, cancelled and moved after
+        # documents carrying the token exist, so "sessb" says nothing about when the session meets.
```
rjsparks, line 1312: I had this comment added so that future analysis of this code doesn't assume that sessa, sessb, etc. implies an ordering (without the comment, Claude made that assumption).

ietf/meeting/utils.py
```diff
@@ -690,6 +690,37 @@ def get_meeting_sessions(num, acronym):
...
+    """The sessions from get_sessions() that are currently in the 'sched' state, in the same order
+
+    Cancelled sessions and the tombstones left by rescheduling keep their timeslot assignment, so
+    get_sessions() still returns them. Anything that treats "all of the group's sessions" as a unit,
+    such as material uploads that apply to every session, must use this subset instead.
+    """
+    return [s for s in sessions if s.current_status == "sched"]
+
+
+def apply_to_all_sessions(session):
```
jennifer-richards: nit: naming reads like a verb, implying it will apply

jennifer-richards (approved):

Issue #4671, opened 2022-10-28
Problematic unscheduled session labels on interim session details view


jennifer-richards:
### Describe the issue

The `ietf.meeting.views.session_details` view for an interim shows both scheduled and unscheduled sessions. These are labeled by their session time or meeting time, depending on what is available. For scheduled interims, this is the local time if a country is specified or UTC time For unscheduled sessions, this is the meeting date. This selection is made in the view.

The tzaware branch displays the chosen time in meeting time zone using the `timezone` template filter. This creates at least a couple problems
* For unscheduled interims, where the "time" is actually a date object, the timezone template filter fails and the result is no label at all
* For scheduled interims, time is now always shown in meeting time because the tzinfo attached by the view is ignored

To resolve, need to decide what the desired behavior is and update the implementation to match that.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

rjsparks:
While working out what the right thing to do is, remember to take edge cases like https://datatracker.ietf.org/meeting/interim-2020-teep-03/session/teep into account.

rjsparks:
Generally:
* Don't show materials in unscheduled sessions when there are scheduled sessions and materials have the "apply to all session in meeting" configured. (This may require back-computation to see if the sessionpresentation sets are the same).
* If there are no materials to show, group the remaining unscheduled sessions by name and just list names with counts. This will be only be really useful for groups like `ietf` that have a bunch of unrelated sessions connected to them.

rjsparks:
@jennifer-richards Is this still in need of change?

jennifer-richards:
Yes, I think so. At least, it still blindly calls `time|timezone:session.meeting.time_zone`, even if `time` is a `datetime.date`. 

I'm not sure why I was concerned that "the tzinfo attached by the view is ignored." The timestamp for a scheduled session is shown in meeting time with the time zone labeled. Since there's no UI to change the display time zone, I think that's fine.

So probably just keep it from showing nothing when unscheduled?

rjsparks:
Updating my comment based on preparation to address this and several other issues:

We should _show_ materials in unscheduled sessions if they exist. What we should not do is attach more to them if apply to all is selected - we should only apply to actually scheduled sessions. We shouldn't allow drag-drop of slides into the unscheduled sessions, but _should_ allow drag out of them.

PR #11783, opened 2026-09-17
fix: cancelled and rescheduled sessions are read-only on the materials page

rjsparks:
A cancelled session, or the tombstone a rescheduling leaves behind, will
not happen. The materials page still listed it with every upload button
and drop target of a live session, and the views behind those buttons
accepted changes.

Such a session now appears only if it still holds material, under a
folded "Cancelled Sessions (n)" or "Rescheduled Sessions (n)" heading,
with no management controls. Every management view refuses it. Slides
can be dragged out of it to a session that will happen, but not into it.

Part of #4671.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ietf/meeting/utils.py
```diff
@@ -705,6 +705,22 @@ def scheduled_only(sessions):
     return [s for s in sessions if s.current_status == "sched"]
 
 
+INACTIVE_SESSION_STATUSES = Session.CANCELED_STATUSES + ["resched"]
+
+
+def session_is_inactive(session):
```
jennifer-richards, line 711: This should perhaps be a method (or computed property) on `Session`. At the least, `INACTIVE_SESSION_STATUSES` should be `Session.INACTIVE_STATUSES` I think.

ietf/meeting/views.py
```diff
@@ -3019,10 +3020,10 @@ def session_details(request, num, acronym):
         raise Http404
 
     scheduled_sessions = scheduled_only(sessions)
-    unscheduled_sessions = [s for s in sessions if s not in scheduled_sessions]
 
     status_names = {n.slug: n.name for n in SessionStatusName.objects.all()}
     for session in sessions:
+        session.inactive = session_is_inactive(session)
```
jennifer-richards, line 3026: Here and `session.cancelled =...` below  seem like `Session` properties rather than monkey-patch candidates

jennifer-richards (approved):
Looks ok. There's some minor refactoring that we should consider, but it's not used widely enough to be a priority so going to accept this and we can think more about that.

PR #11808, opened 2026-09-22
fix: cache handling + constraints for API tokens

jennifer-richards:
API token cache handling in the admin 12.77.0 is incorrect - it updates the cache before the admin has finished updating related models. This fixes that and adds a missing cache update after deleting an API token entirely.

Also adds a unique constraint on `KnownApiEndpoint.name`. This unfortunately requires two migrations because postgresql transactions don't permit both data and schema migrations in a single migration without turning off atomicity. Oh well.


kesara (approved):
LGTM

